PR #448, opened 2022-08-03
curl|sh install docs

vincent99:


docs/docs/30-installation/01-installing.md
```diff
@@ -46,16 +54,16 @@ cp ~/Downloads/acorn /usr/local/bin/acorn
 
 #### Linux
 
-Download the tar.gz binary for your architecture. Uncompress and move the binary to your PATH.
+Download the tar.gz archive for your architecture. Uncompress and move the binary to your PATH.
```
saiyam1814, line 57: I raised a PR for this https://github.com/acorn-io/acorn/pull/404 but @ibuildthecloud mentioned we wont be needing this. 
But since I see it in this PR the we can add the exact steps which are easier

```shell 
# Linux amd64
version=$(curl -s "https://api.github.com/repos/acorn-io/acorn/releases" | grep tag_name | grep controller | sort -r | awk -F'"' 'NR==1 {print $4}')
curl -LO acorn-$version-linux-amd64.tar.gz
tar -xvf acorn-$version-linux-amd64.tar.gz
mv acorn /usr/local/bin/
# verify binary (assume local directory)
acorn -v
```

saiyam1814: Also I wanted to contribute to the install script but I am down with covid :(

vincent99: Searching and sorting github's releases is problematic various reasons, so we setup a channelserver (like in k3s) to do the mapping of latest/stable to a vx.y.z for both this and the setup action.  It's based on the k3s install script, you can tweak it in acorn-io/get-acorn if you want.

saiyam1814: ok got it

cloudnautique (approved):

PR #449, opened 2022-08-03
goreleaser improvements

ekristen:
changelog to use github as source, pre-release suppose and discussions creation


ibuildthecloud (approved):

cjellick:
does this mean it will use PR titles/links rather than commits?

Issue #452, opened 2022-08-03
Can't create Acorn DNS entries if the ingress has CNAMES and IPs


saiyam1814:
The controller logs says 
Cluster - Civo Kuebrnetes
Ingress controller - nginx 
```
kubectl get ing -A
NAMESPACE                 NAME   CLASS   HOSTS                                     ADDRESS       PORTS   AGE
floral-sun-c9c3c365-01e   web    nginx   web.floral-sun.k18kpd.alpha.on-acorn.io   212.2.241.2   80      29m
```
ingress svc 

```
ingress-nginx             ingress-nginx-controller             LoadBalancer   10.43.203.193   212.2.241.2     80:30772/TCP,443:30265/TCP   32m

```
```
time="2022-08-03T16:15:03Z" level=warning msg="Cannot create DNS for ingress web because it has both IPs and hostnames"
```

yaml

```
apiVersion: v1
items:
- apiVersion: networking.k8s.io/v1
  kind: Ingress
  metadata:
    annotations:
      acorn.io/targets: '{"web.floral-sun.k18kpd.alpha.on-acorn.io":{"port":8080,"service":"web"}}'
      apply.acorn.io/applied: H4sIAAAAAAAA/4xSsW7bMBD9leJmSZHrxpYJdGi3okuHbFWGI3myBFMkQZ7sBAL/vaBc13HTFNnIu3eP7x3fDCMxamQEMQNa6xh5cDYuV+WCrQZ3xxj2xBEEzC2cSFadcQFNGSdbHVbNwesKje+xcra8DLUg5ha8C9yCaOqmLlqIFI6DIigAvTfP1Z8H3MlSKPfHAwgYLFOwaK7d46r48H2w+vMX77/ZyGjf5rA4Egi4CvwfMHpUGb203gLGSZbKWaYnBgGQClCBliU9DCNFxtGDsJMxBRiUZG5Xh97/W9PfgFdaLoARLe5JgwAOE73s/F7nhf9E8rWHHmMPAra6a7abptad1rtPH7d6s9VrKTfynqiTtdab9W5V1yrbu6F7Ke1qoFQ7tVbrzX1ZryjPRE8q+w6ToQji5wy9i3xmeU9aoICe2WcKj9yfKSSqA1mdi5fkiPlWXs7XUpxGSeEctJRSsbCAgDs4Hx+efR76EagbniA9pvSYVTPytPyXcai/osnJCiDmlNKvAAAA//+8Ps6rGgMAAA
[...]

Issue #451, opened 2022-08-03
Deleting acorn dns secret doesnt cause DNS to update


cjellick:

cjellick:
fixed!

PR #460, opened 2022-08-04
Fix DNS creation bugs

cjellick:
This fixes the following bugs:
- IPv6 addresses just weren't supported at all. The calls to the DNS
  service would blow up
- If the ingress had both hostnames and IPs in the status, the handler
  threw an error, but this is a valid combination
- If the acorn-dns secret was deleted and then recreated, ingresses
  would get updated, but new DNS records would not get created. The hash
algorithm wasn't taking domain name into account

Addresses: #452 and #451

### DIFF
--- a/pkg/dns/helpers.go
+++ b/pkg/dns/helpers.go
@@ -16,33 +16,19 @@ import (
 // ToRecordRequestsAndHash creates DNS records based on the ingress and domain supplied. It also returns a hash of those
 // records, suitable for using overtime to determine if the ingress's records need to change.
 func ToRecordRequestsAndHash(domain string, ingress *v1.Ingress) ([]RecordRequest, string) {
-	var ips, lbHosts, recordValues []string
+	var ipv4s, ipv6s, lbHosts, recordValues []string
 
 	for _, i := range ingress.Status.LoadBalancer.Ingress {
 		if i.IP != "" {
-			ips = append(ips, i.IP)
+			if strings.Contains(i.IP, ":") {
+				ipv6s = append(ipv6s, i.IP)
+			} else {
+				ipv4s = append(ipv4s, i.IP)
+			}
 		}
 		if i.Hostname != "" {
 			lbHosts = append(lbHosts, i.Hostname)
 		}
-	}
-	var recordType string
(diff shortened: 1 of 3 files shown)


cjellick:
marking as WIP until i do a little end-to-end testing.

cjellick:
Tested CNAME integration on civo and ipv6 integration on DO + k3s/w dualstack support enabled

pkg/dns/helpers.go
```diff
@@ -16,34 +16,20 @@ import (
 // ToRecordRequestsAndHash creates DNS records based on the ingress and domain supplied. It also returns a hash of those
 // records, suitable for using overtime to determine if the ingress's records need to change.
 func ToRecordRequestsAndHash(domain string, ingress *v1.Ingress) ([]RecordRequest, string) {
-	var ips, lbHosts, recordValues []string
+	var ipv4s, ipv6s, lbHosts, recordValues []string
 
 	for _, i := range ingress.Status.LoadBalancer.Ingress {
 		if i.IP != "" {
-			ips = append(ips, i.IP)
+			if strings.Contains(i.IP, ":") {
```
ibuildthecloud, line 23: I'd probable do net.ParseIP and check the result if it's v4 or v6. Maybe looking for : is sufficient, I don't know, but everything about IPv6 is terrible.

ibuildthecloud (changes requested):

pkg/dns/helpers.go
cjellick: I went around on this several times. Googled around a good bit. Surprising that there is not 1 clean and standard way.

My logic here was that this was the simplest solution give the constraints. The biggest constraint/assumption is that this field is already a validated IP address, so it will always be a basic ipv4 of v6 addr.

Are you strongly opposed to keeping it this way until proven wrong?

cjellick: I really think they only way it wouldnt be sufficient is if the addr contained a port and I don't think they will in this case. Interestingly, ParseIP decides which it is (and how to parse it) by which occurs first, a dot or colon

ibuildthecloud (approved):

PR #487, opened 2022-08-08
adding "class" to volume docs

clemenko:
adding a doc section for storage classes in the volume reference.

Signed-off-by: Andy Clemenko <clemenko@users.noreply.github.com>

### DIFF
--- a/docs/docs/100-reference/03-acornfile.md
+++ b/docs/docs/100-reference/03-acornfile.md
@@ -446,6 +446,7 @@ container: db: {
 volumes: data: {
 	size: "100G"
 	accessModes: "readWriteOnce"
+	class: "default"
 }
 ```
 ### size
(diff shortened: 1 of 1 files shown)


ibuildthecloud:
This currently isn't documented because we were going to remove it from the Acornfile.  Let me know if you disagree.  The reason is that "class" is a deployment specific parameter and by putting it in the acorn it will tie your application to a specific infrastructure implementation.

Instead of putting the class in the Acornfile you can specify it during the run by doing `acorn run -v "data,class=longhorn" my-acorn`.  More info here: https://docs.acorn.io/running/volumes

clemenko:
This makes it problematic for sharing and documenting. The less in the command line the better. I know it is similar to how docker does things. But I would rather document in the Acornfile rather than adding a "how to run" section for the app. 

I guess you could add notes to the Acornfile on how to run. OH, Will labels be added to the Acornfile? aka dockerfile labels?

cjellick:
Adding this to our project planning to make sure we come to a conclusion on it and it doesnt just sit

ibuildthecloud:
@clemenko I'm might revert my position on this.  I'm thinking a lot :)  Definitely don't want you to have do document how to run your app with big long CLI args.

clemenko:
Cool. To me, having as much in the acornfile is better due to the infra as code approach now a days. aka keep `acorn run -f Acornfile` as clean as possible.

ibuildthecloud (approved):

ibuildthecloud:
Okay, changed my mind :)  Thanks for the docs and making me think about this more.  I'm merge once all the CI is green.

PR #524, opened 2022-08-10
add t.Cleanup func to clear out the local cert directories

luthermonson:


ibuildthecloud:
I guess this is better than adding it to the .gitignore :)

ibuildthecloud (approved):

Issue #574, opened 2022-08-17
Support ingress to app running in k3d on macOS


tekumara:
An acorn app deployed to k3d:
```
$ acorn app
NAME            IMAGE          HEALTHY   UP-TO-DATE   CREATED   ENDPOINTS                                                       MESSAGE
awesome-acorn   832d8f10698f   3         3            46s ago   http://app.awesome-acorn.192iu1.alpha.on-acorn.io => app:5000   OK

$ kubectl get service -n awesome-acorn-6fe36cda-a4c app
NAME   TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)    AGE
app    ClusterIP   10.43.106.45   <none>        5000/TCP   12m

$ kubectl get ingress -n awesome-acorn-6fe36cda-a4c
NAMESPACE                    NAME   CLASS    HOSTS                                        ADDRESS      PORTS   AGE
awesome-acorn-6fe36cda-a4c   app    <none>   app.awesome-acorn.192iu1.alpha.on-acorn.io   172.24.0.3   80      8m33s
```

ie: app.awesome-acorn.192iu1.alpha.on-acorn.io => 172.24.0.3

However this is not routable from the macOS host.

cjellick:
@iwilltry42 do you know some trick here? I believe there's some port-forwarding that needs setup or something?

cjellick:
actually, i got this working awhile ago. Required a little extra setup around port proxying. I'l try to spin it up again and let you know what you need

cjellick:
@tekumara can you share the command you used to spin up your k3d cluster?

cjellick:
So, there was a convoluted way to get it to work as-is, but it kind of sucked. So, I have a PR to make this better:
https://github.com/acorn-io/acorn/pull/576

tekumara:
I used `k3d cluster create acorn -p 8000:80@loadbalancer`.

When I need access from the host to a service running in the k3d cluster, I've had to either:
- create an [IngressRoute](https://github.com/tekumara/orion-demo/blob/327d306b17e4d664122445e35aead5b3b9f95bdd/infra/ingress-orion.yaml) for the service inside the cluster, or
- create a [Service with type LoadBalancer](https://github.com/tekumara/orion-demo/blob/327d306b17e4d664122445e35aead5b3b9f95bdd/infra/lb-minio.yaml)

cjellick:
[...]

PR #576, opened 2022-08-17
Fix k3d ingress integration

cjellick:
- If the cluster is k3d backed, then we should use local.on-acorn.io,
  rather than the real DNS service
- Support adding a port to the `--cluster-domain` flag so that if you
  are forwarding traffic to your k3d cluster on an arbitrary port, you
can still get urls that resolve properly
- Update documentation on how to setup a k3d cluster for Acorn


cjellick:
should address https://github.com/acorn-io/acorn/issues/574

ibuildthecloud (approved):

Issue #553, opened 2022-08-15
Ensure exec works


ibuildthecloud:
As part of install we should validate that exec works as that will break a lot of things in obscure ways.

cjellick:
@iwilltry42, putting this in your bucket for now

cjellick:
verified

PR #585, opened 2022-08-18
Introduce preflight checks and `acorn check` command

### LINKED ISSUE #464: Add a check option to check for readiness of Kubernetes cluster before install
We should add either a separate command or an option on `acorn install` to check that the Kubernetes cluster is up and running and the user has proper admin privilege. Otherwise `acorn install` fails and the user is confused.

### LINKED ISSUE #476: Ingress: Detect and deal with missing ingress controller
We need to inform the user that their cluster is not "Ingress" capable.  The unfortunate this is that a lack of a IngressClass resource does not mean that there are no ingress controllers.  In k3s we have ingress, but there is no IngressClass.

### LINKED ISSUE #545: check for service loadbalancer functionality on install
Services of type load balancer are necessary to fulfill non-http (ie TCP) acorn endpoints.

Not all clusters have this capability. Moreover, if you are using a cloud provider, a service of type loadbalancer will cost you money because an actual LB in the cloud will be created.

We don't want to cost people money without asking, so a few thoughts:
- We could make this an "opt-in" feature where we ask the user (or they tell us via a flag) if they want us to check if service loadbalancers are functional
- We could do a best effort that at least checks for the out-of-tree cloud providers and reports back on that.

At this point, we are just brainstorming. Welcome better ideas on this.

### LINKED ISSUE #546: Check for default storage class during install
During installation, we should check to see if the cluster has a default storage class. If it doesn't we should warn the user that the cluster won't be capable of handling persistent volumes correctly.

### LINKED ISSUE #547: Check for cluster-admin perms during install
When you're installing acorn, you need cluster-admin privileges. @ibuildthecloud believes we already check for this and abort installation with an error message that clearly indicates what the problem is, but he isn't sure.

Can we double check this? Try to install as non-cluster-admin and see how the install fails and if it is clear to the user.

iwilltry42:
References:

## Pre-Flight (critical) Checks

:exclamation: Those (could) make `acorn install` fail :exclamation: 

- [x] #464 
	- cluster up and running: Preflight checks if all nodes are up (FIXME!)
		-	Would it be enough to just test API connectivity?
	- admin privileges: see below
- [x] #547
	- Errors are pretty clear about Kubernetes RBAC permissions already. The preflight check tests if the user can create namespaces as that's highest privilege in Kubernetes and also required for the installation process.

## In-Flight (non-critical) Checks

:warning: Those won't make Acorn fail but limit its functionality

- [x] #476 
	- Test: create Endpoint + Service + Ingress: watch ingress for 10s max. to see if it gets updated by an ingress controller 
- [ ] #545 
- [x] #546 
- [x] #553

### DIFF
--- a/pkg/cli/install.go
+++ b/pkg/cli/install.go
@@ -20,6 +20,8 @@ acorn install`,
 }
 
 type Install struct {
+	DisablePreflightChecks bool `usage:"Disable preflight checks"`
+
 	Image  string `usage:"Override the default image used for the deployment"`
 	Output string `usage:"Output manifests instead of applying them (json, yaml)" short:"o"`
 
@@ -38,9 +40,10 @@ func (i *Install) Run(cmd *cobra.Command, args []string) error {
 	}
 
 	return install.Install(cmd.Context(), image, &install.Options{
-		OutputFormat:       i.Output,
-		Config:             i.Config,
-		APIServerReplicas:  i.APIServerReplicas,
-		ControllerReplicas: i.ControllerReplicas,
+		DisablePreflightChecks: i.DisablePreflightChecks,
+		OutputFormat:           i.Output,
+		Config:                 i.Config,
+		APIServerReplicas:      i.APIServerReplicas,
+		ControllerReplicas:     i.ControllerReplicas,
 	})
 }
--- a/pkg/cli/install.go
+++ b/pkg/cli/install.go
@@ -20,7 +20,7 @@ acorn install`,
 }
 
 type Install struct {
-	DisablePreflightChecks bool `usage:"Disable preflight checks"`
+	Checks *bool `usage:"Disable preflight checks with --checks=false"`
 
 	Image  string `usage:"Override the default image used for the deployment"`
 	Output string `usage:"Output manifests instead of applying them (json, yaml)" short:"o"`
@@ -40,10 +40,10 @@ func (i *Install) Run(cmd *cobra.Command, args []string) error {
 	}
 
 	return install.Install(cmd.Context(), image, &install.Options{
-		DisablePreflightChecks: i.DisablePreflightChecks,
-		OutputFormat:           i.Output,
-		Config:                 i.Config,
-		APIServerReplicas:      i.APIServerReplicas,
-		ControllerReplicas:     i.ControllerReplicas,
+		Checks:             i.Checks,
+		OutputFormat:       i.Output,
+		Config:             i.Config,
+		APIServerReplicas:  i.APIServerReplicas,
+		ControllerReplicas: i.ControllerReplicas,
 	})
 }
--- a/pkg/install/check/check.go
+++ b/pkg/install/check/check.go
@@ -7,16 +7,22 @@ import (
 	"os"
 	"time"
 
+	"github.com/acorn-io/acorn/pkg/client/term"
+	"github.com/acorn-io/acorn/pkg/k8schannel"
 	"github.com/acorn-io/acorn/pkg/k8sclient"
+	"github.com/acorn-io/acorn/pkg/scheme"
+	"github.com/acorn-io/acorn/pkg/streams"
 	"github.com/acorn-io/acorn/pkg/system"
 	"github.com/acorn-io/baaah/pkg/restconfig"
 	authorizationv1 "k8s.io/api/authorization/v1"
 	corev1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
 	storagev1 "k8s.io/api/storage/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/rest"
 	"k8s.io/klog"
 	klogv2 "k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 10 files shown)


cjellick:
Oh @iwilltry42 I think we want this to be baked into the install command. Maybe we can have a quick chat with @ibuildthecloud sometime about it today or tomorrow.

iwilltry42:
> Oh @iwilltry42 I think we want this to be baked into the install command. Maybe we can have a quick chat with @ibuildthecloud sometime about it today or tomorrow.

It's both right now.
If preflight checks fail during install, it will suggest to debug using the check command.
That way the install output will be consistent and not spammed (i.e. only using the progress log output).
But feel free to setup a meeting to discuss this 👍

pkg/cli/install.go
```diff
@@ -20,6 +20,8 @@ acorn install`,
 }
 
 type Install struct {
+	DisablePreflightChecks bool `usage:"Disable preflight checks"`
```
ibuildthecloud, line 23: Can we just change this to "checks' not have disable in there.  So `--checks=false` to disable it.  You need to change the field type to `*bool` which should work and then treat nil as true.

pkg/install/check/check.go
```diff
@@ -0,0 +1,392 @@
...
+// silenceKlog is a helper function to silence klog output which could disturb
+// the clean installation logs.
+func silenceKlog() {
+	klog.SetOutput(io.Discard)
+	klogv2.SetOutput(io.Discard)
+	utilruntime.ErrorHandlers = nil
+}
+
+// newClient is a helper function to quickly create a new k8s client instance
+func newClient() (client.WithWatch, error) {
```
ibuildthecloud, line 85: this function seems to be basically the same as k8sclient.Default()

pkg/install/check/check.go
```diff
@@ -0,0 +1,392 @@
...
+	cfg, err := restconfig.Default()
+	if err != nil {
+		return nil, err
+	}
+
+	return k8sclient.New(cfg)
+}
+
+func CheckIngressCapability() CheckResult {
```
ibuildthecloud, line 94: Can you pass in ctx and use it in the subsequent calls

pkg/install/check/check.go
```diff
@@ -0,0 +1,392 @@
...
+
+}
+
+/*
+ * CheckNodesReady checks if all nodes are ready.
+ * -> This is a critical check, which "could" affect the installation.
+ * TODO: We only need to check if the cluster is operational.
+ * -> A single malfunctiuning node should not prevent the installation.
+ */
+func CheckNodesReady() CheckResult {
```
ibuildthecloud, line 252: pass in ctx here also

ibuildthecloud: Basically all preflight check methods

iwilltry42: Done :heavy_check_mark:

pkg/install/check/check.go
iwilltry42: replaced :heavy_check_mark:

pkg/cli/install.go
iwilltry42: Done :heavy_check_mark:

iwilltry42:
Had to do some force-pushes to add the required signoffs

cjellick:
@iwilltry42 let's get this PR green. You need to update the generated docs via:
```
make gen-docs
```
and update generated code via
```
make generate
```

iwilltry42:
@cjellick done :heavy_check_mark:

pkg/install/check/check.go
```diff
@@ -0,0 +1,525 @@
...
+	pod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "acorn-install-check-exec",
+			Namespace: system.Namespace,
+		},
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
+				{
+					Name:            "exec",
+					Image:           "busybox",
```
cjellick, line 114: could present an issue for airgap installs.

pkg/install/check/check.go
```diff
@@ -0,0 +1,525 @@
...
+// They are not critical and should not affect the installation process.
+func InFlightChecks(ctx context.Context) []CheckResult {
+	checks := []func(ctx context.Context) CheckResult{
+		CheckDefaultStorageClass,
+		CheckIngressCapability,
+		CheckExec,
+	}
+
+	// Some debugging test
+	if os.Getenv("ACORN_INSTALL_FAIL_CHECKS") == "true" {
```
cjellick, line 54: wonder if we should keep this

pkg/install/check/check.go
cjellick: would using the acorn image be better? ...of course then you need to know the version of the image thats available

cjellick: This is how it is retrieved/got during install:
https://github.com/acorn-io/acorn/blob/main/pkg/cli/install.go#L35-L38

You don't want to just use the defaultImage, you want to use the user-supplied one. So somehow you have to hook into that logic

pkg/install/check/check.go
iwilltry42: It was actually only for you two to test this. For some of this I found it cumbersome to come up with scenarios where the checks failed. I will delete it :)

pkg/install/check/check.go
iwilltry42: > @ibuildthecloud said that you should be able to find the image we are going to install. Should be an api for it.

I guess this won't work for the `acorn check` command, right?
Should I introduce a `--image` flag there? :thinking:

iwilltry42: I found it was the easiest to just pass in that information via an options struct :man_shrugging: 
If you have any better idea, let me know :)

cjellick (approved):

cjellick:
Merging, but FYI im giong to put up a PR to tweak the styling a bit.

cjellick (approved):

PR #601, opened 2022-08-22
Fix TLS docs example indentation

vknabel:
The indentation of the example YAML definition of the the Certificate on the [Acorn Docs TLS Certificates](https://docs.acorn.io/running/certificates#add-with-cert-manager) is wrong.

This leads to YAML parsing errors when trying to apply.


cjellick (approved):
Thanks!

Issue #486, opened 2022-08-08
TLS: Support for user provided domains


cjellick:
This is NOT for on-acorn.io domains

This is for the case where user supplies their own cluster domain, or even more likely creates an app with the following publish flag:
```
-p myapp.mydomain.com:<app deatils>
```

Supporting this means installing and managing cert-manager and auto-creating the `Certificate` CRs if the app wants TLS

cjellick:
We _arent_ going to install cert-manager.
User expresses they want TLS. If we find the Certificate CR, then we program it

andyjeffries:
What's the proposal then for how cert-manager should be installed? Just manually in to the cluster?

I don't know your design goals, but this was feeling to me like "bring me a cluster, give me access, deploy your apps with easy commands". Darren just said in the Webinar:

![Screenshot 2022-08-10 at 18 14 34](https://user-images.githubusercontent.com/22904/183975320-5e772cb8-0a91-47d5-a499-1102ab41bafc.png)

This feels like "No knowledge of Kubernetes YAML required, FOOTNOTE: unless you want to use your own domain name and use TLS with it". Hopefully it's reconsidered, having Acorn install cert-manager for you unless already installed/available.

andyjeffries:
Ahhh, I took the wording of the issue to be almost like a "we won't be doing this", but I guess it's just pending, based on:

> In the future, Acorn will provide built-in mechanisms to automatically provide certificates for each endpoint. Today, adding a certificate must follow the manual approach.

-- https://docs.acorn.io/running/certificates

vincent99:
@andyjeffries There's a balance we're trying to find the right line on between batteries-included (you just want a cert, I don't know or care about CertManager) and tolerable-for-cluster-operator-to-install.  The more stuff we add to someone's cluster the more they worry about trying it, compatibility between versions issues, should it be cleaned up on uninstall, etc.

CertManager does a lot of stuff and we don't even actually need so one option we're considering is not requiring it at all by doing ACME dances in the controller instead.  That's straightforward enough for single name http-01 challenges.

The part CertManager does do that's really useful is taking creds and talking to all the provider APIs.  But we can maybe do a one-time setup `_acme_challenge.yourdomain.com CNAME [your-cluster].on-acorn.io` yourself step, and then with that we can manage a *.yourdomain.com wildcard without having any creds or provider-specific implementation.

andyjeffries:
[...]

PR #614, opened 2022-08-24
Don't create generated dns name when user provides one

cloudnautique:


cjellick (approved):

cloudnautique:
See [#486 ](https://github.com/acorn-io/acorn/issues/486#issuecomment-1225833916) for additional context.

PR #623, opened 2022-08-25
Improve container build time in Getting Started document.

eengelking:
Updated the Dockerfile section to improve load time from ~7s to ~2s each time a change is made to code.

### DIFF
--- a/docs/docs/37-getting-started.md
+++ b/docs/docs/37-getting-started.md
@@ -120,8 +120,9 @@ WORKDIR /app
 ENV FLASK_APP=app.py
 ENV FLASK_RUN_HOST=0.0.0.0
 RUN apk add --no-cache gcc musl-dev linux-headers
-COPY . .
+ADD requirements.txt .
 RUN pip install -r requirements.txt
+COPY . .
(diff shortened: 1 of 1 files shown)


cjellick:
@iwilltry42 can you test this and approve if it works?

iwilltry42:
Hey @eengelking , thanks for this optimization!
It makes total sense the way the layering works, so that requirements don't have to be installed everytime something in the code is changed :+1: 
In the getting-started guide we also use the dev mode, which works around this for development by live-reloading code with Flask :)

iwilltry42 (approved):
LGTM

eengelking:
> In the getting-started guide we also use the dev mode, which works around this for development by live-reloading code with Flask :)

@iwilltry42 In this usecase, I was using the `--dev` or `-i` flags, but I was updating the Acornfile `welcome` argument instead of the Python code directly. As a result, it doesn't allow the Flask live reload to be taken into consideration.  

Thank you for the approval. 👍

cjellick:
Thanks @eengelking!

PR #625, opened 2022-08-26
Do not release PVs until a new PVC is bound

ibuildthecloud:
If we release the PV it could be claimed by another PVC and we don't want
that.  So we leave it in the released state until a PVC binds it by name
and then we move it to available


cjellick (approved):
I see you didn't add a test. No judgement, but is that because there is something particularly tricky about testing it or just not worth the effort?

Otherwise, LGTM. So merge away

PR #627, opened 2022-08-28
fix: respect render output type for json

azazeal:
This PR reworks the `render` command so that it outputs JSON when required.

PR #630, opened 2022-08-29
Fix: make render -o equivalent to render --output

cjellick:


cjellick:
noticed while testing https://github.com/acorn-io/acorn/pull/627

ibuildthecloud (approved):

PR #616, opened 2022-08-24
change: replace nginx ingress controller with traefik

iwilltry42:
Signed-off-by: Thorsten Klein <tk@thklein.io>

ref: #544

Issue #544, opened 2022-08-12
Switch ingress from nginx to traefik


cjellick:
When installing acorn on Docker Desktop, if no Ingress Controller is found, we auto-install the nginx ingress controller.

We've noticed that it is very slow to populate Ingresses with an IP. This leads to a bad user experience, so we want install traefik instead. It seems pretty fast with K3s and Rancher Desktop installs of acorn.

Here's the PR where @ibuildthecloud introduced this change:
https://github.com/acorn-io/acorn/pull/421
Should be fairly straightforward to swap traefik in.

cjellick:
@iwilltry42 - @ibuildthecloud had made nginx the defaultIngressClass, so acorn worked out of the box.

You did not do that, so out of the box, apps get stuck in `Pending Ingress`.

Rather than setting traefik at the k8s-level default ingress class, can you make it so that traefik gets set in acorn's `ingress-class-name` field (when we install traefik, that is)?

cjellick:
Hm. i set `--ingress-class-name traefik` and the arguement is making it through to the ingresses that are created, but traefik still isnt picking them up.

Judging from the logs of the traefik pod, looks like a permissions issue:
```
E0831 20:58:12.822040       1 reflector.go:138] k8s.io/client-go@v0.22.1/tools/cache/reflector.go:167: Failed to watch *v1.Endpoints: failed to list *v1.Endpoints: endpoints is forbidden: User "system:serviceaccount:default:traefik" cannot list resource "endpoints" in API group "" at the cluster scope
E0831 20:58:13.106374       1 reflector.go:138] k8s.io/client-go@v0.22.1/tools/cache/reflector.go:167: Failed to watch *v1.Secret: failed to list *v1.Secret: secrets is forbidden: User "system:serviceaccount:default:traefik" cannot list resource "secrets" in API group "" at the cluster scope
```

cjellick:
Let's also change the namespace we install traefik to from `default` to `acorn-system`.

cjellick:
Did a fresh install of your PR and this is what I'm getting:
```
time="2022-09-01T15:10:01Z" level=error msg="Cannot create service: service not found" servicePort="&ServiceBackendPort{Name:,Number:80,}" ingress=default namespace=red-resonance-1c78caa4-2ba serviceName=default providerName=kubernetes
[...]

PR #640, opened 2022-09-01
change: install traefik into acorn-system namespace and fix ingressclass issues

iwilltry42:
Ref #544 
Ref #616

### DIFF
--- a/pkg/install/traefik.yaml
+++ b/pkg/install/traefik.yaml
@@ -1,5 +1,4 @@
 ---
-# Source: traefik/templates/rbac/serviceaccount.yaml
 kind: ServiceAccount
 apiVersion: v1
 metadata:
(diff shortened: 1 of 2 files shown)


pkg/install/traefik.yaml
```diff
@@ -198,25 +199,30 @@ spec:
...
+items:
+  - apiVersion: v1
+    kind: Service
+    metadata:
+      name: traefik
+      labels:
+        app.kubernetes.io/name: traefik
+        app.kubernetes.io/managed-by: Acorn
+        app.kubernetes.io/instance: traefik
+      annotations:
```
cjellick, line 214: empty? still need to have this line?

iwilltry42: Nope, missed during cleanup. Deleted.

cjellick (approved):

Issue #644, opened 2022-09-01
`check` fails because the exec pod already exists


cjellick:
If you run `acorn check` right after install or quickly twice in a row, you get:
<img width="1254" alt="Screen Shot 2022-09-01 at 2 02 51 PM" src="https://user-images.githubusercontent.com/2473240/188012657-6b171ecc-8338-4db3-9955-7c9142af9efe.png">

Can we delete the pod before attmepting to create it?

iwilltry42:
True.. unfortunately it's super slow to terminate that test pod. One way would be to wait for deletion before creating it.
Another one would be to just use a unique ID on every run.

I actually favor the latter since 
1. we don't leave dangling resources anyway since we defer deletion
2. we cannot make use of the existing pod
3. waiting for termination/deletion would need another chunk of logic and especially time

WDYT?

cjellick:
Yeah, I live using unique ID every run. let's do that

meldafrawi:
Validation: **PASS**

`acorn check` will create a pod with unique suffix with every run

Running `acorn check`  (x4)

```
kubectl get pod -n acorn-system 
NAME                                READY   STATUS        RESTARTS       AGE
buildkitd-68cdbf6979-rpdfk          2/2     Running       4 (100m ago)   21h
acorn-api-59fb7c754d-mbj9k          1/1     Running       3 (100m ago)   21h
acorn-controller-556fdd9bf9-d568k   1/1     Running       3 (100m ago)   21h
acorn-check-exec-78ztp4k4           1/1     Terminating   0              9s
acorn-check-exec-zc5tkrlg           1/1     Terminating   0              7s
acorn-check-exec-22vl6wlr           1/1     Terminating   0              5s
acorn-check-exec-jbbnm6d8           1/1     Terminating   0              3s

```

PR #671, opened 2022-09-08
change: generate random suffices for check objects

iwilltry42:
Ref: #644

### DIFF
--- a/pkg/install/check.go
+++ b/pkg/install/check.go
@@ -13,7 +13,9 @@ import (
 	"github.com/acorn-io/acorn/pkg/scheme"
 	"github.com/acorn-io/acorn/pkg/streams"
 	"github.com/acorn-io/acorn/pkg/system"
+	"github.com/acorn-io/baaah/pkg/randomtoken"
 	"github.com/acorn-io/baaah/pkg/restconfig"
+	"github.com/rancher/wrangler/pkg/name"
 	authorizationv1 "k8s.io/api/authorization/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


cjellick (approved):

cjellick (changes requested):
Could you just use the GenerateName field instead of the Name field and let k8s make the random name for you?

iwilltry42:
> Could you just use the GenerateName field instead of the Name field and let k8s make the random name for you?

For the exec check yes, but in the ingress-check, I have to use name references, so I have to have it available before creating the other objects. I could do that, but that would require the logic to first fetch the first object from the API and retrieve its name.
Doing it based on labels would yield the same "challenge".

cjellick (approved):

meldafrawi:
Validation: **PASS**

`acorn check` will create a pod with unique suffix with every run

Running `acorn check`  (x4)

```
kubectl get pod -n acorn-system 
NAME                                READY   STATUS        RESTARTS       AGE
buildkitd-68cdbf6979-rpdfk          2/2     Running       4 (100m ago)   21h
acorn-api-59fb7c754d-mbj9k          1/1     Running       3 (100m ago)   21h
acorn-controller-556fdd9bf9-d568k   1/1     Running       3 (100m ago)   21h
acorn-check-exec-78ztp4k4           1/1     Terminating   0              9s
acorn-check-exec-zc5tkrlg           1/1     Terminating   0              7s
acorn-check-exec-22vl6wlr           1/1     Terminating   0              5s
acorn-check-exec-jbbnm6d8           1/1     Terminating   0              3s

```

PR #675, opened 2022-09-09
Enable the thelper linter

azazeal:
This PR enables the `thelper` linter in `.golangci.yml`. It additionally decorates existing tests with a call to `(*testing.T).Helper` for better stack traces.

Signed-off-by: Panagiotis Siatras <azazeal@users.noreply.github.com>

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -16,5 +16,6 @@ linters:
   - ineffassign
   - staticcheck
   - typecheck
+  - thelper
   - unused
   fast: false
--- a/integration/client/client.go
+++ b/integration/client/client.go
@@ -8,6 +8,8 @@ import (
 )
 
 func NewImage2(t *testing.T, namespace string) string {
+	t.Helper()
+
 	image, err := build.Build(helper.GetCTX(t), "../testdata/nginx2/Acornfile", &build.Options{
 		Client: helper.BuilderClient(t, namespace),
 		Cwd:    "../testdata/nginx2",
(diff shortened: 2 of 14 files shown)


cjellick:
can you explain in a little more detail? 
What does the linter do and what do the method calls do?
I guess it has to do with printing more useful line numbers on test failures, but I dont get the linter bit.

azazeal:
The `thelper` linter makes sure that test helper functions call `(*testing.T).Helper`.

Functions that call this method are skipped by the test suite when printing file and line 
information ([ref](https://pkg.go.dev/testing#T.Helper)) so failure traces are more relevant.

Code readers also get the added benefit of understanding whether a function is a test-helper or a sub-test as sub-tests won't typically call `(*testing.T).Helper`.

cjellick:
Oh, so I guess it work like this: If the function accepts testing.T as a parameter, but doesn't start with the word Test, it must be a helper and must call t.Helper()?

@ibuildthecloud any issues with this?

azazeal:
Before @ibuildthecloud pitches in, I'd like to point out that I plan on making an effort to submit PRs that enable additional linters.

If you don't see a point in this, I'd be happy to not put any effort in this but I think a few linters (like context inheritance) would allow better code hygiene for a small cost.

Lemme know :)

ibuildthecloud:
wow, took me a while to grok what t.Helper is, but now I got it :).

ibuildthecloud (approved):

PR #678, opened 2022-09-12
Initial libsodium encryption of secrets

cloudnautique:
This adds the basics for encryption of secret data on a per-cluster acorn user namespace basis. There is a convenience call to bulk encrypt a message for multiple targets given a list of public keys. The implementation leverages libsodium sealed boxes where the nonce is a computed blake2b hash value of the sender and recipient's public keys.

On startup of Acorn, the cluster will generate a public/private key pair for the default namespace and store it in a K8s secret in the `acorn-system` namespace.
The key pair is available as a base64 encoded string in the `acorn info` command under the "namespace" top level key.

Users can encrypt a message for the current namespace by running:

`acorn secret encrypt MSG` 

or

`acorn secret encrypt --public-key <ns A pubkey> --public-key <ns B pubkey> MSG`

The output can be placed in an Acornfile, passed as an argument to an Acorn image or provided by an existing secret at Runtime.

The format of an encrypted secret is:

`ACORNENC:base64encode({"publicKey":"base64encode(ciphertext)", "publicKey":"base64encode(ciphertext)"})`

This allows for the same message to be encrypted and passed to multiple clusters. Each cluster will look up the ciphertext in the map by it's own public keys. An error will be returned if there isn't a key that can decrypt the value.


pkg/apis/api.acorn.io/v1/types.go
```diff
@@ -325,6 +326,11 @@ type Config struct {
 	AcornDNSEndpoint             *string        `json:"acornDNSEndpoint" name:"acorn-dns-endpoint" usage:"The URL to access the Acorn DNS service"`
 }
 
+type EncryptionKeySpec struct {
```
cjellick, line 329: Why does this end in "Spec"? Generally, we use the Spec suffix to denote a struct that is literally the Spec field for a particular time. That's not the case with this type. So, unless I'm misunderstanding its point, i'd drop Spec from the name.

cloudnautique: It was the non-k8s native object I was using when going down the path of adding Acorn APIs. I can remove it.

pkg/cli/secret_encrypt.go
```diff
@@ -0,0 +1,63 @@
...
+	out := table.NewWriter([][]string{
+		{"Name", "{{.}}"},
+	}, "", true, "")
+
+	c, err := client.Default()
+	if err != nil {
+		return err
+	}
+
+	if len(args[0]) > 4096 {
```
cjellick, line 36: is this our arbitrary limit or does something break if its bigger than that?

cloudnautique: It's the limit of libsodium. GH and golang mention it.

cjellick (changes requested):
Just that one change to Spec. Otherwise seems good to go.

pkg/apis/api.acorn.io/v1/types.go
cloudnautique: Removed the spec suffix.

cjellick (approved):
LGTM, but looks like your tests are failing

Issue #205, opened 2022-06-20
TLS: Support for acorn-dns provided domains


cjellick:
## Problem Statement:
We want to give users an easy way to get TLS endpoints for the acorn apps.
There are three different use cases where we need to secure endpoints with TLS:
1. Acorn generated [on-acorn.io](http://on-acorn.io/) domains
2. Domains specified by the user at install time using the `--cluster-domain .mydomain.com` syntax (which replaces and overrides [on-acorn.io](http://on-acorn.io/))
3. Domains specified by the user at run Time using the `--publish hostname:service:port` syntax

This issue represents **JUST** the first use case.

## Constraints and assumptions
- We don’t want to install, manage, and upgrade cert-manger in the user's cluster. So, we want to use libraries to do our own direct ACME based integration.
- There is some "prior art" for doing similar things in rio. You can check that code out for ideas/implementation details, here: https://github.com/rancher/rio/tree/e9d490246e7252a04eb872a99f542c5da2f135cf/modules/letsencrypt/pkg. I (Craig) didn't write that code and don't know how useful it will be, but it should at least be reviewed for ideas.

## Background on acorn-dns
To better understand this feature, here's some background on how acorn-dns works.

Right now, when you install acorn, an `on-acorn.io` subdomain is provisioned for your cluster. It will be like: `<random slug>.on-acorn.io`. This is done as a handler, watching the acorn configMap that contains the installer's config choices, here: https://github.com/acorn-io/acorn/blob/main/pkg/controller/dns/config.go#L36

When you then create an application, acorn makes a request to our [acorn-dns](https://github.com/acorn-io/acorn-dns) service to create a DNS entry for the containers in the application. The domain will look like: `container-name.app-name.<random slug>.on-acorn.io`. All this DNS logic is driven by the ingress that acorn creates for the app. That logic is here: https://github.com/acorn-io/acorn/blob/main/pkg/controller/dns/ingress.go

Note: if your cluster type is a "local" one like rancher desktop, docker desktop, minikube, while we will reserve a domain for you (which doesnt actually create any real dns entries), when you launch an application, we won't actually create a dns entry for you. We have a wildcard `*.local.on-acorn.io` entry that always resolves to `127.0.0.1` and we'll use that for you local app.

## Proposed solution
[...]

PR #693, opened 2022-09-14
Automatic TLS Certificates for on-acorn domains

iwilltry42:
Ref #205 

## To-Do

- [x] Renew if expired
- [x] Set as default for Acorn Managed Ingress
- [x] Get Let's Encrypt details from user input
- [x] Re-Generate if Let's Encrypt settings changed (environment or email)


iwilltry42:
Still hitting some issue that needs fixing:
```
acorn-controller-5d889f4d64-k8xc5 acorn-controller time="2022-09-14T18:50:00Z" level=error msg="error syncing 'acorn-system/acorn-config': handler acorn /v1, Kind=ConfigMap: problem generating wildcard certificate: error: one or more domains had a problem:\n[*.o84oxs.staging.on-acorn.io] [*.o84oxs.staging.on-acorn.io] acme: error presenting token: unexpected response status code: 500\n, requeuing"
```

pkg/controller/dns/letsencrypt.go
```diff
@@ -0,0 +1,295 @@
...
+			}
+		} else {
+			return leUser, nil
+		}
+	}
+
+	// Create and Register Let's Encrypt User
+	leUser := &LEUser{
+		Name:  targetUsername,
+		Email: fmt.Sprintf("%s@on-acorn.io", targetUsername),
```
ibuildthecloud, line 205: We are going to need to prompt the user during install for an email address. We can't use a generic one, that got us in trouble with Rio when we did that.

pkg/publish/ingress.go
```diff
@@ -19,12 +19,12 @@ import (
...
 
 func toPrefix(serviceName string, appInstance *v1.AppInstance) string {
-	hostPrefix := serviceName + "." + appInstance.Name
+	hostPrefix := serviceName + "---" + appInstance.Name
 	if serviceName == "default" {
 		hostPrefix = appInstance.Name
 	}
 	if appInstance.Namespace != system.DefaultUserNamespace {
-		hostPrefix += "." + appInstance.Namespace
+		hostPrefix += "---" + appInstance.Namespace
```
ibuildthecloud, line 27: What is this for?

pkg/controller/dns/letsencrypt.go
```diff
@@ -0,0 +1,295 @@
...
+	"github.com/go-acme/lego/v4/registration"
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	LEEndpoint = "https://acme-staging-v02.api.letsencrypt.org/directory"
```
ibuildthecloud, line 31: This is going to have to come from config.  The only reason the existing DNS was a compile time option is we wanted to change the default value for dev builds. But for LE the default should always be staging and the user should opt into production. We should prompt during install on which one they want (along with email address for production, staging can use a dummy email).

pkg/controller/dns/letsencrypt.go
```diff
@@ -0,0 +1,295 @@
...
+			if err != nil {
+				logrus.Errorf("problem parsing existing TLS secret: %v", err)
+				// continues to generate new cert
+			} else {
+				if x509crt.NotAfter.After(time.Now().UTC()) {
+					// cert is still valid, return it
+					logrus.Infof("existing TLS secret %s is still valid until %s", x509crt.Subject.CommonName, x509crt.NotAfter)
+					return sec, nil
+				} else {
+					logrus.Infof("existing TLS secret %s is expired since %s", x509crt.Subject.CommonName, x509crt.NotAfter)
```
ibuildthecloud, line 251: Is this how renew works? If so you need to renew the cert before it expires. So something like a week before. Also you can't delete the secret until a new cert is obtained to avoid downtime.

iwilltry42:
Screenshot-20220919102050-848x645

pkg/publish/ingress.go
iwilltry42: We can't use dots here since the wildcard cert is only valid one level deep and the triple dash is mostly to avoid ambiguities as mentioned by Vincent here: https://github.com/acorn-io/acorn/issues/205#issuecomment-1240070940 (does that make sense?)

pkg/controller/dns/letsencrypt.go
iwilltry42: Alright, two things here
1. LE staging certs are not trusted by default in browsers.. that's not a problem? Could as well use self-signed certs then, right?
2. Is there some standard way defined already to ask for user input? Also, can one specify a config file upon installation?

iwilltry42: Oh wow... sorry for my still newborn-son-influenced brain... You probably just mean some new flags, right? :man_facepalming:

iwilltry42:
**Default: Let's Encrypt Staging**
Screenshot-20220922175004-918x774

**Let's Encrypt Production**

Screenshot-20220922175333-899x661


Screenshot-20220922175311-904x383

pkg/install/install.go
```diff
@@ -100,12 +101,36 @@ func DefaultImage() string {
...
+
 func Install(ctx context.Context, image string, opts *Options) error {
 	// I don't want these errors on the screen. Probably a better way to do this.
 	klog.SetOutput(io.Discard)
 	klogv2.SetOutput(io.Discard)
 	utilruntime.ErrorHandlers = nil
 
+	if opts.Config.LetsEncrypt != nil && *opts.Config.LetsEncrypt == "production" {
+		if opts.Config.LetsEncryptEmail == "" {
+			result, err := pterm.DefaultInteractiveTextInput.WithMultiLine(false).Show("Enter your email address for Let's Encrypt (required for production)")
```
iwilltry42, line 120: Should this be replaced by something similar to https://github.com/acorn-io/acorn/blob/main/pkg/cli/credential_login.go#L54-L70 ?

.github/workflows/main-release.yaml
```diff
@@ -57,6 +57,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           COSIGN_PASSWORD: ${{ secrets.COSIGN_PASSWORD }}
           ACORN_DNS_ENDPOINT: https://staging-dns.acrn.io/v1
+          LE_ENDPOINT: https://acme-staging-v02.api.letsencrypt.org/directory
```
cjellick, line 60: this and the next two files should be unnecessary now, right?

pkg/apis/api.acorn.io/v1/types.go
```diff
@@ -317,6 +317,8 @@ type Config struct {
 	IngressClassName             *string        `json:"ingressClassName" usage:"The ingress class name to assign to all created ingress resources (default '')"`
 	ClusterDomains               []string       `json:"clusterDomains" name:"cluster-domain" usage:"The externally addressable cluster domain (default .on-acorn.io)"`
 	TLSEnabled                   *bool          `json:"tlsEnabled" name:"tls-enabled" usage:"If true HTTPS URLs will be rendered for HTTP endpoint URLs (default false)"`
+	LetsEncrypt                  *string        `json:"letsEncrypt" name:"lets-encrypt" usage:"staging|production|disabled. If set, the Let's Encrypt environment to use for TLS certificates (default staging)"`
+	LetsEncryptEmail             string         `json:"letsEncryptEmail" name:"lets-encrypt-email" usage:"Required if --lets-encrypt=production. The email address to use for Let's Encrypt (default '')"`
```
cjellick, line 321:
```suggestion
	LetsEncryptEmail             string         `json:"letsEncryptEmail" name:"lets-encrypt-email" usage:"Required if --lets-encrypt=production. The email address to use for Let's Encrypt registration(default '')"`
```

cjellick:
@iwilltry42 - need to do a `make generated` and a `make gen-docs`.

pkg/controller/dns/letsencrypt.go
```diff
@@ -0,0 +1,349 @@
+package dns
```
cjellick, line 1: Can you go through this file and ensure that only the things that are used outside of package have public visibility?

All the fields on the LEUser struct should be private and many of the functions and methods should also be private.

It'll help know exactly what we're exposing for consumption outside this package.

pkg/controller/dns/letsencrypt.go
```diff
@@ -0,0 +1,349 @@
...
+		Registration: &reg,
+		Key:          privateKey,
+		URL:          string(secret.Data["url"]),
+	}, nil
+}
+
+func (u *LEUser) GetEmail() string {
+	return u.Email
+}
+func (u LEUser) GetRegistration() *registration.Resource {
```
cjellick, line 71: This function isn't using a pointer like the others. It should

pkg/controller/dns/letsencrypt.go
```diff
@@ -0,0 +1,349 @@
...
+	return hex.EncodeToString(dig.Sum(nil))
+}
+
+func EnsureLEUser(ctx context.Context, cfg *apiv1.Config, client kclient.Client, domain string) (*LEUser, error) {
+
+	/*
+	 * Construct new LE User
+	 */
+	targetUsername := strings.TrimPrefix(domain, ".") // leading dot can be an issue
+	email := fmt.Sprintf("%s@on-acorn.io", targetUsername)
```
cjellick, line 201: lets just hard code the staging email to `staging-certs@acorn.io`. I'm going to create that as a real email address so they dont bounce.

pkg/controller/dns/letsencrypt.go
```diff
@@ -0,0 +1,349 @@
...
+	conf := lego.NewConfig(u)
+
+	conf.CADirURL = u.URL
+
+	client, err := lego.NewClient(conf)
+	if err != nil {
+		return err
+	}
+
+	reg, err := client.Registration.Register(registration.RegisterOptions{TermsOfServiceAgreed: true})
```
cjellick, line 104: We probably need to prompt the user to agree to LE's TOS.

I am thinking that if you say you are doing production LE, you can either pass the flag `--lets-encrypt-tos-agree=true` or if you don't do that, the CLI blocks and prints out a message like:

```
You are choosing to use Let's Encrypt production environment for TLS certificates. To do so, you must agree to their Terms of Serivice: https://community.letsencrypt.org/tos

Do you agree to Let's Encrypt TOS? y/n
```

Thoughts?

pkg/system/constants.go
```diff
@@ -5,6 +5,8 @@ import "os"
 const (
 	Namespace            = "acorn-system"
 	ConfigName           = "acorn-config"
+	TLSSecretName        = "acorn-tls"
+	LESecretName         = "acorn-le"
```
cjellick, line 9: For clarity, can you add "account" to this?
```suggestion
	LEAccountSecretName         = "acorn-le-account"
```

pkg/dns/daemon.go
```diff
@@ -82,6 +82,8 @@ func (d *Daemon) internal(ctx context.Context) (bool, error) {
 		return false, nil
 	}
 
+	// TODO: renew certificate
```
cjellick, line 85: Yeah, I think renewal should happen here, not in the handler. Any reason you can't move that logic here? Would be fine if its in both places

pkg/controller/dns/config.go
```diff
@@ -63,6 +63,20 @@ func (h *configHandler) Handle(req router.Request, resp router.Response) error {
...
 
+	if !strings.EqualFold(*cfg.LetsEncrypt, "disabled") {
+		// Ensure that we have a Let's Encrypt account ready
+		leUser, err := EnsureLEUser(req.Ctx, cfg, req.Client, domain)
+		if err != nil {
+			return err
+		}
+
+		// Generate wildcard certificate for domain
+		_, err = leUser.EnsureWildcardCertificateSecret(req.Ctx, req.Client, *cfg.AcornDNSEndpoint, domain, token)
```
cjellick, line 74: said elsewhere, but i think we need to also handle renew in the dns daemon, not just the handler

pkg/controller/dns/provider.go
```diff
@@ -0,0 +1,86 @@
...
+
+/*
+ * DNS01 Challenge Solver (Lego Interface)
+ */
+
+type DNSProvider struct {
+	client DNSClient
+}
+
+func NewDNSProvider(endpoint, domain, token string) *DNSProvider {
```
cjellick, line 27: For clarity, can you rename this `NewACMEDNS01ChallengeProvider`. It wasn't clear to me what it was for at first

pkg/dns/helpers.go
```diff
@@ -9,7 +9,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/api/networking/v1"
+	v1 "k8s.io/api/networking/v1"
```
cjellick, line 12: weird change. can you eliminate it?

pkg/install/install.go
cjellick: What you did here looks fine when I run it locally. I don't see a need to to switch to survey

pkg/controller/dns/letsencrypt.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	x509crt, err := certcrypto.ParsePEMCertificate([]byte(cert.Certificate))
+
+	sec = &corev1.Secret{
+		Type: corev1.SecretTypeTLS,
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      system.TLSSecretName,
+			Namespace: system.Namespace,
+			Annotations: map[string]string{
```
cjellick, line 320: Of these annotations, only the hash is ever read. What is the intention for the other three? Just informational?

pkg/controller/dns/config.go
cjellick: changed my mind. this is good enough. dont need to also do in a daemon

pkg/dns/daemon.go
cjellick: changed my mind. not necessary. the controller is good enough for now i think

pkg/controller/dns/letsencrypt.go
```diff
@@ -0,0 +1,349 @@
...
+	conf.CADirURL = u.URL
+
+	client, err := lego.NewClient(conf)
+	if err != nil {
+		return nil, err
+	}
+
+	dnsProvider := NewDNSProvider(dnsendpoint, domain, token)
+
+	if err := client.Challenge.SetDNS01Provider(dnsProvider); err != nil {
```
cjellick, line 166: As mentioned in slack, I ran into a lot of trouble with my local DNS resolver trying to do the "pre check" for DNS. Not sure if you caught that bit of the convo, but this client attempts to ensure that the TXT record is there locally before telling let's encrypt to do it. The problem is that local dns resolvers can act really funky and for me, they never succeed. We agreed that if we just wait 60 seconds (the documented time for dns to proprogate throughout route53), we can just skip the local check entirely.

Here is the diff that allowed me to achieve this locally:
```
diff --git a/pkg/controller/dns/letsencrypt.go b/pkg/controller/dns/letsencrypt.go
index c5af25f..a17b690 100644
--- a/pkg/controller/dns/letsencrypt.go
+++ b/pkg/controller/dns/letsencrypt.go
@@ -21,6 +21,7 @@ import (
 	"github.com/go-acme/lego/v4/certcrypto"
 	"github.com/go-acme/lego/v4/certificate"
+	"github.com/go-acme/lego/v4/challenge/dns01"
 	"github.com/go-acme/lego/v4/lego"
 	"github.com/go-acme/lego/v4/registration"
 	"github.com/sirupsen/logrus"
@@ -143,6 +144,10 @@ func (u *LEUser) ToSecret() (*corev1.Secret, error) {
 	}, nil
 }

+func noOpCheck(_, _, _ string, _ dns01.PreCheckFunc) (bool, error) {
+	return true, nil
+}
+
 func (u *LEUser) GenerateWildcardCert(dnsendpoint, domain, token string) (*certificate.Resource, error) {
 	if u.Registration == nil {
 		return nil, fmt.Errorf("not generating LE cert: missing registration")
@@ -163,7 +168,7 @@ func (u *LEUser) GenerateWildcardCert(dnsendpoint, domain, token string) (*certi

 	dnsProvider := NewDNSProvider(dnsendpoint, domain, token)

-	if err := client.Challenge.SetDNS01Provider(dnsProvider); err != nil {
+	if err := client.Challenge.SetDNS01Provider(dnsProvider, dns01.WrapPreCheck(noOpCheck)); err != nil {
 		return nil, err
 	}

diff --git a/pkg/controller/dns/provider.go b/pkg/controller/dns/provider.go
index 93c1b20..612167d 100644
--- a/pkg/controller/dns/provider.go
+++ b/pkg/controller/dns/provider.go
@@ -40,7 +40,7 @@ func (d *DNSProvider) CleanUp(domain, token, keyAuth string) error {
 }

 func (d *DNSProvider) Timeout() (timeout, interval time.Duration) {
-	return 30 * time.Second, 5 * time.Second
+	return 3 * time.Minute, 1 * time.Minut
 }

```

cjellick (changes requested):

meldafrawi:
if user install acorn **without** `--tls-enabled --lets-encrypt --lets-encrypt-email`  and wanted to enable LE later, should the user just reinstall with the flags? will it require any restart on acorn api or controller ?

pkg/install/install.go
```diff
@@ -100,12 +101,36 @@ func DefaultImage() string {
...
+			if err != nil {
+				return err
+			}
+			opts.Config.LetsEncryptEmail = result
```
meldafrawi, line 124: @iwilltry42 
if email is required when using Let's Encrypt  `production`, should we check email address is valid before setting `opts.Config.LetsEncryptEmail`, and abort installation if not valid ? 

cc: @cjellick

cjellick:
Ok, Here is the summary of todos/changes/comments, adapted from what you sent me
- Update host prefix handling according to https://github.com/acorn-io/acorn/pull/693#discussion_r981859831
- When configured to use staging LE, set the email to staging-certs@acorn.io. If the user supplies an email anyway, respect it https://github.com/acorn-io/acorn/pull/693#discussion_r981815386
- Cert loop should not block/error-out DNS config handler. This is currently causing it to spam acorn-dns with new domain requests
- ~Cert Refresh will be handled in a Daemon as well~ Per our slack convo and my comments here, I changed my mind. The current approach is good enough, like you said
- Other PR comments, like changing the visibility of functions/variables in the letsencrypt.go file

If you wanted to get the above core functionality done and get the PR merged and then follow up on the below in a separate PR, I'd be fine with that, up to you.
- If the user opts-in to product let's encrypt, do the following:
  - Prompt the user to agree to the TOS per:  https://github.com/acorn-io/acorn/pull/693#pullrequestreview-1122851564
  - Add a message in the install output saying "You've enabled automatic TLS certificates with Let's Encrypt. This can take a few minutes to configure"
  - Add something to acorn info to say whether or not the certificate has been issued. Can be as simple as `letsEncrypt: production, pending`, `letsEncrypt: production, provisioned`
  - The URLs in `acorn ps` should start with https. Doing this to all endpoints actually is probably an oversimplification. So, if you can figure out a more elegant solution to do it on a per endpoint basis, great. But I don't want that to hold up shipping this feature this week. All that logic is here: https://github.com/acorn-io/acorn/blob/main/pkg/controller/appdefinition/status.go#L390-L397

.github/workflows/main-release.yaml
iwilltry42: Correct, removed.

pkg/controller/dns/letsencrypt.go
iwilltry42: They make debugging easier in case anything goes wrong, but I'm totally fine with removing them as they're only informational.

pkg/install/install.go
iwilltry42: We're checking it outside of this `if` statement so it's checking the validity for production **and** staging.
It will error out in both scenarios if the email is invalid: https://github.com/acorn-io/acorn/pull/693/files#diff-3c83c1219d23236aa4c92d672faa69bb463e2001adb87833ca0f7434faded420R127-R131

iwilltry42:
> if user install acorn **without** `--tls-enabled --lets-encrypt --lets-encrypt-email` and wanted to enable LE later, should the user just reinstall with the flags? will it require any restart on acorn api or controller ?

I think this was clarified via Slack already, but no, it won't require restarts as it's hooked up in the config handler and secret changes are propagated transparently. :magic_wand:

pkg/install/install.go
iwilltry42: Moved it even further down to always check it if it was set, either via flag (for both prod/staging) or via interactive input (prompt only for prod)

pkg/controller/dns/letsencrypt.go
iwilltry42: Unused anyway, removed.

iwilltry42: Changed.

pkg/controller/dns/letsencrypt.go
iwilltry42: Done.

pkg/controller/dns/letsencrypt.go
iwilltry42: Makes total sense, done!

pkg/controller/dns/provider.go
iwilltry42: Done.

pkg/dns/helpers.go
iwilltry42: For some reason, vscode adds this automatically on save. Removed it with vim for now :D

pkg/publish/ingress.go
iwilltry42: Whoops, forgot to add the `<appname>` part. Done.
Example: `http://my-webapp-dawn-moon-9b980585.v4avai.staging.on-acorn.io/`

pkg/apis/api.acorn.io/v1/types.go
```diff
@@ -316,7 +317,9 @@ type Config struct {
 
 	IngressClassName             *string        `json:"ingressClassName" usage:"The ingress class name to assign to all created ingress resources (default '')"`
 	ClusterDomains               []string       `json:"clusterDomains" name:"cluster-domain" usage:"The externally addressable cluster domain (default .on-acorn.io)"`
-	TLSEnabled                   *bool          `json:"tlsEnabled" name:"tls-enabled" usage:"If true HTTPS URLs will be rendered for HTTP endpoint URLs (default false)"`
+	LetsEncrypt                  *string        `json:"letsEncrypt" name:"lets-encrypt" usage:"enabled|disabled|staging. If enabled, uses the Let's Encrypt production environment to issue a wildcard TLS certificate. (default disabled)"`
+	LetsEncryptEmail             string         `json:"letsEncryptEmail" name:"lets-encrypt-email" usage:"Required if --lets-encrypt=production. The email address to use for Let's Encrypt registration(default '')"`
```
cjellick, line 321:
```suggestion
	LetsEncryptEmail             string         `json:"letsEncryptEmail" name:"lets-encrypt-email" usage:"Required if --lets-encrypt=enabled. The email address to use for Let's Encrypt registration(default '')"`
```

pkg/apis/api.acorn.io/v1/types.go
```diff
@@ -316,7 +317,9 @@ type Config struct {
...
+	LetsEncryptTOSAgree          *bool          `json:"letsEncryptTOSAgree" name:"lets-encrypt-tos-agree" usage:"Required if --lets-encrypt=production. If true, you agree to the Let's Encrypt terms of service (default false)"`
```
cjellick, line 322:
```suggestion
	LetsEncryptTOSAgree          *bool          `json:"letsEncryptTOSAgree" name:"lets-encrypt-tos-agree" usage:"Required if --lets-encrypt=enabled. If true, you agree to the Let's Encrypt terms of service (default false)"`
```

pkg/config/config.go
```diff
@@ -50,6 +51,20 @@ func complete(c *apiv1.Config, ctx context.Context, getter kclient.Reader) error
...
 	}
+	if c.LetsEncrypt == nil {
+		c.LetsEncrypt = &LetsEncryptOptionDefault
+	}
+	if c.LetsEncryptTOSAgree == nil {
+		c.LetsEncryptTOSAgree = new(bool)
+	}
+	if *c.LetsEncrypt == "enabled" {
+		if c.LetsEncryptEmail == "" {
+			return fmt.Errorf("letsencrypt email is required when using production")
```
cjellick, line 62:
```suggestion
			return fmt.Errorf("letsencrypt email is required when Let's Encrypt is enabled")
```

pkg/config/config.go
```diff
@@ -50,6 +51,20 @@ func complete(c *apiv1.Config, ctx context.Context, getter kclient.Reader) error
...
+		}
+		if !*c.LetsEncryptTOSAgree {
+			return fmt.Errorf("letsencrypt TOS must be agreed to when using production")
```
cjellick, line 65:
```suggestion
			return fmt.Errorf("letsencrypt TOS must be agreed to when Let's Encrypt is enabled")
```

pkg/controller/dns/letsencrypt.go
```diff
@@ -0,0 +1,354 @@
...
+	/*
+	 * Construct new LE User
+	 */
+	targetUsername := strings.TrimPrefix(domain, ".") // leading dot can be an issue
+	email := "staging-certs@acorn.io"
+	url := LetsEncryptURLStaging
+	if strings.EqualFold(*cfg.LetsEncrypt, "enabled") {
+		url = LetsEncryptURLProduction
+		if cfg.LetsEncryptEmail == "" {
+			return nil, fmt.Errorf("let's encrypt email is required for production")
```
cjellick, line 211:
```suggestion
			return nil, fmt.Errorf("let's encrypt email is required")
```

pkg/install/install.go
```diff
@@ -100,12 +101,51 @@ func DefaultImage() string {
...
+			if err != nil {
+				return err
+			}
+			if !ok {
+				return fmt.Errorf("you must agree to Let's Encrypt TOS to use their production environment")
+			}
+			opts.Config.LetsEncryptTOSAgree = &ok
+		}
+		if opts.Config.LetsEncryptEmail == "" {
+			result, err := pterm.DefaultInteractiveTextInput.WithMultiLine(false).Show("Enter your email address for Let's Encrypt (required for production)")
```
cjellick, line 131:
```suggestion
			result, err := pterm.DefaultInteractiveTextInput.WithMultiLine(false).Show("Enter your email address for Let's Encrypt")
```

pkg/install/install.go
```diff
@@ -100,12 +101,51 @@ func DefaultImage() string {
...
+			if err != nil {
+				return err
+			}
+			opts.Config.LetsEncryptEmail = result
+		}
+		pterm.Info.Println("You've enabled automatic TLS certificate provisioning with Let's Encrypt production. This can take a few minutes to configure.")
```
cjellick, line 137:
```suggestion
		pterm.Info.Println("You've enabled automatic TLS certificate provisioning with Let's Encrypt. This can take a few minutes to configure.")
```

pkg/controller/appdefinition/status.go
```diff
@@ -369,13 +375,34 @@ func podName(pod *corev1.Pod) string {
...
+		LabelSelector: klabels.SelectorFromSet(map[string]string{
+			labels.AcornManaged: "true",
+			labels.AcornAppName: app.Name,
+		}),
+	})
+	if err != nil {
+		return "", err
+	}
+
+	ingressTLSHosts := []string{}
```
cjellick, line 397: nitpick, a map would be more efficent. If this where a `map[string]interface{}`, line 423 would be a much more efficient lookup.

pkg/install/install.go
```diff
@@ -100,12 +101,51 @@ func DefaultImage() string {
...
+	}
+
+	// Validate E-Mail address provided for Let's Encrypt registration
+	if opts.Config.LetsEncryptEmail != "" {
```
cjellick, line 141: Found a bug with this: If I use the prompts to enter my email, it lets me just hit enter to submit a blank email.
That lets the install continue with a blank email and then the controller gets in an error loop.

So, this check probably needs to add an "or" for when letsEncrypt is enabled, like `if opts.Config.LetsEncryptEmail != "" || opts.COnfig.LetsEncrypt == "enabled {`

pkg/install/install.go
```diff
@@ -100,12 +101,51 @@ func DefaultImage() string {
...
 func Install(ctx context.Context, image string, opts *Options) error {
 	// I don't want these errors on the screen. Probably a better way to do this.
 	klog.SetOutput(io.Discard)
 	klogv2.SetOutput(io.Discard)
 	utilruntime.ErrorHandlers = nil
 
+	// Require E-Mail address when using Let's Encrypt production
+	if opts.Config.LetsEncrypt != nil && *opts.Config.LetsEncrypt == "enabled" {
+		if opts.Config.LetsEncryptTOSAgree == nil || !*opts.Config.LetsEncryptTOSAgree {
+			ok, err := pterm.DefaultInteractiveConfirm.Show("You are choosing to use Let's Encrypt production environment for TLS certificates. To do so, you must agree to their Terms of Service: https://community.letsencrypt.org/tos\nTip: use --lets-encrypt-tos-agree to skip this prompt\nDo you agree to Let's Encrypt TOS?")
```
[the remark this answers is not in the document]
cjellick: There is another thing I don't like about this: as soon as you enter 'y', it continues. It doesnt wait for you to hit enter. I think it should. Our other yes/no prompts (most obvious when you do `acorn uninstall -a`) wait until you hit enter.

This caused me to stumble because I hit y and then enter, but enter actually applied to the next prompt, which was for my email and it caused be to submit a blank email

pkg/install/install.go
iwilltry42: Good catch! Alternatively we could make `LetsEncryptEmail` a `*string` type flag as well and check for `nil` instead of `""`.
Went for or check for now :heavy_check_mark:

pkg/controller/appdefinition/status.go
iwilltry42: Yup :)

iwilltry42:
Rebased, fixed conflicts and re-generated docs.
`make generate` still does not generate the openapi stuff for me, so maybe one of you can look into this?

cjellick (approved):

PR #689, opened 2022-09-14
Fix deploy on AKS

meldafrawi:
create DS to add containerd config to allow http for local registry
fix https://github.com/acorn-io/acorn/issues/670

Signed-off-by: mohamed.eldafrawi <mohamed@acorn.io>

PR #708, opened 2022-09-22
add acorn test on azure GH actions

meldafrawi:
The following values are used as secrets:

- AZURE_CREDENTIALS 
- AZURE_RESOURCE_GROUP 

For the test to work we need to create azure service principle account, using:
```
 az ad sp create-for-rbac --name "{account_name}" --role contributor \
                            --scopes /subscriptions/{subscription-id}/resourceGroups/{resource-group} \
                            --sdk-auth
```

The credential json should be set to `AZURE_CREDENTIALS` secret

Signed-off-by: Mohamed Eldafrawi <mohamed@acorn.io>

### DIFF
--- a/.github/workflows/azure-aks-test.yml
+++ b/.github/workflows/azure-aks-test.yml
@@ -1,0 +1,72 @@
+name: test acorn on AKS
+on:
+  schedule:
+    - cron: '00 7 * * *'   # time in UTC
+
+jobs:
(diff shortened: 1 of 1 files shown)


meldafrawi:
should be merged after #689

cloudnautique (approved):
LGTM. Service account and dependent PR need to be wrapped up first.

Issue #466, opened 2022-08-05
Support `acorn run --update --name ...` functionality to achieve behavior like `kubectl apply`


hikhvar:
To install an app and than update it, one must do a `acorn run --name foo` and than on any follow up changes must run `acorn update foo`. It would be really cool to have something like `acorn apply --name foo` or similar to helm an `acorn update --install foo`. This would allow easy docs in new projects. We just have to document the single command to run and update the app if already present.

Moreover, this will ease the setup of CD steps in pipelines, since we must not distinguish in the pipeline between first install and later updates.

ibuildthecloud:
We plan to add a feature like this.  If you are not aware you can do `acorn run -o yaml ...` which will give you the YAML output of the run command.  So what your are asking for is basically the opposite where we take in that yaml.  (dumb hack right now is to actually do `acorn run -o yaml ....` and then edit the yaml and apply with `kubectl apply`.  We want a acorn native way to do that, but it is technically possible because acorn API is just a custom k8s api).

hikhvar:
I basically want to have a "acorn make it so" mode. In that mode I give acorn a acorn image or Acornfile and an application name. And than acorn will figure out if the current application have to be updated or created. Because I actually don't care if it's the first time the application is created (`acorn run`), or the application have to be updated ( `acorn update`). 

This is something I really hate with docker-compose for example. There is no quick way for "make it so".

ibuildthecloud:
Ah cool, I totally misunderstood your issue.  I too would like this feature.  There are three approaches I can think of to address this.

1.  If your run command has a `--name` parameter then we will create or update always.
2. Add a `--update` to run to enable the --update behavior.  This will implicitly require --name to be set
3. Add a `--create` to update to create if not found.

I personally like option 2.  Option 1 is nice but does remove a guardrail of when a user accidentally uses the same name twice and they really wanted two different apps.

hikhvar:
I think, `acorn run --update --name foo` will be a good solution. 
One can easily write this into the CI scripts, or a developer can easily define a shell alias `alias acornapply=acorn run --update --name` and than run `acornapply foo`.

cjellick:
coming in 0.4. updated title to reflect the current plan

cjellick:
Almost there, one more tweak that won't make it in time for v0.4.0. We will release v0.4.1 when this is in

cjellick:
back in! 😎

cjellick:
working as expected. 👍

PR #710, opened 2022-09-24
add acorn run --update/-u flag to update app if exists

iwilltry42:
- triggers acorn update logic if app exists
- does not work with --dev/-i
- requires --name/-n to be specified

Ref #466 

## Notes

- This is not 100% like `kubectl apply`, as that one does a 3-way patch merge, where it takes into account the `last-applied-configuration` annotation. This PR does not consider such annotations.

### DIFF
--- a/docs/docs/100-reference/01-command-line/acorn_update.md
+++ b/docs/docs/100-reference/01-command-line/acorn_update.md
@@ -25,6 +25,7 @@ acorn update [flags] APP_NAME [deploy flags]
       --profile strings           Profile to assign default values
   -p, --publish strings           Publish port of application (format [public:]private) (ex 81:80)
   -P, --publish-all               Publish all (true) or none (false) of the defined ports of application
+      --reset                     
   -s, --secret strings            Bind an existing secret (format existing:sec-name) (ex: sec-name:app-secret)
       --target-namespace string   The name of the namespace to be created and deleted for the application resources
   -v, --volume stringArray        Bind an existing volume (format existing:vol-name,field=value) (ex: pvc-name:app-data)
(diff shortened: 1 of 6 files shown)


cjellick (commented):
At first blush, this looks good to me. I'm going to get @ibuildthecloud's input on it

ibuildthecloud:
I really want this, but the approach should be different. "Run -u" and "update" should be slightly different. Basically "run -u" should be like k8s update and update is like k8s patch. 

update today will only change the fields you specify and not touch the other fields. Run -u should ensure all fields exactly match the arguments. So that means run -u will reset fields back to default values if the user didn't specify the flag.

iwilltry42:
@ibuildthecloud does this come close to what you'd expect?

pkg/client/client.go
```diff
@@ -101,6 +101,7 @@ type AppUpdateOptions struct {
 	DevMode         *bool
 	Image           string
 	TargetNamespace string
+	Reset           bool // Reset is used to indicate whether the update should be a patch (reset=false: only change specified fields) or a full update (reset=true: reset unspecified fields to defaults)
```
ibuildthecloud, line 104: Can you change this to `replace`

cjellick:
@ibuildthecloud mentioned the approach looks right but he wants a new command a different verb that is the equivalent of `update --patch=false`. We decided `acorn replace` should be that verb.

Can you add this @iwilltry42 ?


Moving this out to v0.4.1, which will basically be a milestone for this and one or two other issues that aren't quite ready.

docs/docs/100-reference/01-command-line/acorn_update.md
```diff
@@ -25,6 +25,7 @@ acorn update [flags] APP_NAME [deploy flags]
       --profile strings           Profile to assign default values
   -p, --publish strings           Publish port of application (format [public:]private) (ex 81:80)
   -P, --publish-all               Publish all (true) or none (false) of the defined ports of application
+      --reset
```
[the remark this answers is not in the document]
iwilltry42: Left it empty like the other update-only command. Changed now :+1:

pkg/client/client.go
iwilltry42: Done

Issue #515, opened 2022-08-10
Prompt install if first action is run dev mode


ibuildthecloud:
If you immediately run `acorn run -i .` you will get the following message:

```
ERRO[0000] Failed to build /Users/darren/src/examples/guestbook/Acornfile: no matches for kind "Builder" in version "api.acorn.io/v1" 
```
We should prompt the user if they want to install like we do with `acorn run` in non-dev mode.

iwilltry42:
> like we do with acorn run in non-dev mode.

This doesn't seem to be implemented right now (built from current HEAD:main):

```shell
$ acorn run -i .local 
ERRO[0000] Failed to build /home/thklein/git/github.com/acorn-io/acorn/.local/Acornfile: no matches for kind "Builder" in version "api.acorn.io/v1" 
                                                                                                                                                                                                               
$ acorn run .local   
Error: no matches for kind "Builder" in version "api.acorn.io/v1"

$ acorn --version 
acorn version v0.0.0-dev+0a880145
```

cjellick:
He must be referring to when you run an already built and published image:
```
acorn run ghcr.io/acorn-io/acorn-dns:v0.1.0
? Acorn is not installed, do you want to install it now?: No
```

meldafrawi:
Found nil pointer deference 
```
? Acorn is not installed, do you want to install it now?: Yes
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x48 pc=0x19229b1]

goroutine 1 [running]:
github.com/acorn-io/acorn/pkg/install.Install({0x2f14378, 0xc0005670c0}, {0xc00064a160, 0x1b}, 0x0)
	/home/acorn/go/src/github.com/meldafrawi/acorn/pkg/install/install.go:120 +0xb1
github.com/acorn-io/acorn/pkg/client.promptInstall[...]({0x2f14378?, 0xc0005670c0}, 0xc00108f600)
	/home/acorn/go/src/github.com/meldafrawi/acorn/pkg/client/ignore.go:38 +0x16b
github.com/acorn-io/acorn/pkg/client.IgnoreUninstalled.Info({{0x2f3e7a8?, 0xc0006bca40?}}, {0x2f14378?, 0xc0005670c0?})
	/home/acorn/go/src/github.com/meldafrawi/acorn/pkg/client/ignore.go:251 +0x71
[...]

PR #711, opened 2022-09-25
Prompt for Acorn install on run/dev if needed

iwilltry42:
- wraps call to Info() in installPrompt wrapper

Ref #515

### DIFF
--- a/pkg/cli/run.go
+++ b/pkg/cli/run.go
@@ -188,7 +188,10 @@ func (s *Run) Run(cmd *cobra.Command, args []string) error {
 	}
 
 	// Force install prompt if needed
-	_, _ = c.Info(cmd.Context())
+	_, err = c.Info(cmd.Context())
+	if err != nil {
+		return err
+	}
 
 	opts, err := s.ToOpts()
 	if err != nil {
--- a/pkg/cli/run.go
+++ b/pkg/cli/run.go
@@ -187,6 +187,9 @@ func (s *Run) Run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	// Force install prompt if needed
+	_, _ = c.Info(cmd.Context())
+
 	opts, err := s.ToOpts()
 	if err != nil {
 		return err
@@ -203,8 +206,6 @@ func (s *Run) Run(cmd *cobra.Command, args []string) error {
 	}
 
 	if s.Interactive && isDir {
-		// Force install prompt if needed
-		_, _ = c.Info(cmd.Context())
 
 		return dev.Dev(cmd.Context(), s.File, &dev.Options{
 			Args:   args,
--- a/pkg/client/ignore.go
+++ b/pkg/client/ignore.go
@@ -3,6 +3,7 @@ package client
 import (
 	"context"
 	"errors"
+	"fmt"
 	"net"
 
 	"github.com/AlecAivazis/survey/v2"
@@ -39,6 +40,8 @@ func promptInstall[V any](ctx context.Context, f twoFunc[V]) (V, error) {
 				return v, installErr
 			}
 			v, err = f()
+		} else {
+			return v, fmt.Errorf("action aborted because Acorn is not installed")
 		}
 	}
 	return v, err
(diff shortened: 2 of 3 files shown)


cjellick:
Your change is causing weird behavior for prebuilt images. It is causing acorn to ask me twice when I say no the first time:
```
acorn run ghcr.io/acorn-io/acorn-dns:v0.1.0
? Acorn is not installed, do you want to install it now?: No
? Acorn is not installed, do you want to install it now?: No
Error: no matches for kind "App" in version "api.acorn.io/v1"
```

Can you take a look and fix that?

iwilltry42:
I'll take a look, thank you!
It should just exit if answered with "No", right? Doesn't make sense to continue trying if Acorn is not installed and the user denies installing it :man_shrugging:

cjellick:
right.

Hit another problem (but one we might be able to ignore):
The test command i ran:
```
acorn run ghcr.io/acorn-io/acorn-dns:v0.1.0
```
never actually succeeds because that image require parameters (hence me saying in slack that we should have a hello-world image).

The problem is that the post-install checks get hung up on it. It sits here:
```
STATUS: ENDPOINTS[] HEALTHY[1/2] UPTODATE[2] [containers: default CrashLoopBackOff: back-off 2m40s restarting failed container=default pod=default-585f8d67c9-ljwc2_frosty-waterfall-45e4ba13-ad3(5fbdcbb9-5dc7-40b5-b043-b37e29c63726); frosty-waterfall is not ready]
STATUS: ENDPOINTS[] HEALTHY[1/2] UPTODATE[2] [containers: default Error: Exit Code 1; frosty-waterfall is not ready]
```

Must be some logic in the post-install check that is either waiting for the first app it finds to go healthy or for all apps it finds to go healthy. Can you look into this?

iwilltry42:
@cjellick 

> It is causing acorn to ask me twice when I say no the first time

Fixed. It will now show an error if first one was answered with `No`: 
```bash
$ acorn run ghcr.io/acorn-io/acorn-dns:v0.1.0
? Acorn is not installed, do you want to install it now?: No
Error: no matches for kind "Info" in version "api.acorn.io/v1"
```

I was thinking to provide a prettier error, but that would require some refactorings deeper down in the prompt logic. Not sure if that's worth it, as it's pretty clear that the error is because Acorn is not installed?

> post-install checks get hung up on it.

Fixed. Actually they succeeded, but the progress indicator was still there because it was missing an `s.Success()`

cjellick:
I don't see why we try the next action at all. Can we just abort if the user says no?

This change gives me the behavior I prefer:
```
diff --git a/pkg/client/ignore.go b/pkg/client/ignore.go
index 9185f64..e0af66c 100644
--- a/pkg/client/ignore.go
+++ b/pkg/client/ignore.go
@@ -3,6 +3,7 @@ package client
 import (
 	"context"
 	"errors"
+	"fmt"
 	"net"

 	"github.com/AlecAivazis/survey/v2"
@@ -40,6 +41,7 @@ func promptInstall[V any](ctx context.Context, f twoFunc[V]) (V, error) {
 			}
 			v, err = f()
 		}
+		return v, fmt.Errorf("action aborted. Acorn not installed")
 	}
 	return v, err
 }
```

Is there a reason we can't short circuit like this @ibuildthecloud ?

cjellick:
let's not wait for @ibuildthecloud. we'll beg forgiveness later. If you amek this last change, I'll merge

cjellick:
oh crud, that doesn't work. if you say yes, after the install for some reason it runs through that function again later and fails.

i guess that's what you were getting at

cjellick:
Oh, seems like putting it in an else statement suffices:
```
+		} else {
+			return v, fmt.Errorf("action aborted. Acorn not installed")
 		}
 	}
```


PS, not trying to micromanage your PRs, I just dont know a lot of the areas of the code well because I didnt right them, so i have to get hands-on to grok whats going on

cjellick (changes requested):
would like to see us short circuit if users says no to provide a better error message. see comments above.

iwilltry42:
@cjellick  added else-statement as per your suggestion which results in a more informational error message.
Should we style that up with pterm?

cjellick:
Seems fine to be as-is. I'm going to merge. If you want to do a follow up PR to pretty it up, fine by me. If not, also fine by me.

cjellick (approved):

Issue #363, opened 2022-07-28
Turn list argument into repeatable flag


iwilltry42:
Example argument:
```cue
args: {
  extra_food: ["hazelnut", "walnut"]
}
```

Expected usage:

`acorn run . --extra-food foo --extra-food bar`

Right now, all variants that I tried to pass in some sort of list (e.g. the one above, comma-separated, etc.) end up with an error similar to the one below:

```bash
acorn render . --extra-food foo --extra-food bar
Error: args.extra_food: 2 errors in empty disjunction:
args.extra_food: conflicting values "bar" and ["hazelnut","walnut"] (mismatched types string and list):
    /_internal_/args.cue:1:23
    Acornfile:2:15
args.extra_food: conflicting values "bar" and [...string] (mismatched types string and list):
    /_internal_/args.cue:1:23
    Acornfile:2:15
```

cjellick:
not going to try to squeeze this into v0.4.1. bumping out to v0.5

cjellick:
Wonder if we can revisit this and get it out in v0.6.0. If you have a PR, it probably needs rebased pretty badly 😅

cjellick:
This is low priority, much lower than the signing and attestation, but I'm bringing it up in case you need/want a change of base occasionally while working on that feature.

PR #712, opened 2022-09-26
add stringArray type flags

iwilltry42:
- e.g. args: { myarr: ["foo", "bar"] } can be overriden by --myarr baz --myarr bom which results in args: { myarr: ["baz", "bom"] }
- does only for for string arrays

Signed-off-by: Thorsten Klein <tk@thklein.io>

Ref #363 

## Notes

Does not work with int arrays. Just adding them to the Acornfile already results in the below error:
```
acorn run -f .local/Acornfile.ex .local --extra-food foo --extra-food bar --help
Error: args.blub: 6 errors in empty disjunction:
args.blub: conflicting values [1,2,4] and {...} (mismatched types list and struct):
    /_internal_/schema/v1/app.cue:280:52
    /_internal_/schema/v1/app.cue:283:18
    Acornfile:4:9
args.blub: conflicting values bool and [1,2,4] (mismatched types bool and list):
    /_internal_/schema/v1/app.cue:280:31
    Acornfile:4:9
args.blub: conflicting values float and [1,2,4] (mismatched types float and list):
    /_internal_/schema/v1/app.cue:280:23
    Acornfile:4:9
args.blub: conflicting values int and [1,2,4] (mismatched types int and list):
    /_internal_/schema/v1/app.cue:280:17
    Acornfile:4:9
args.blub: conflicting values string and [1,2,4] (mismatched types string and list):
    /_internal_/schema/v1/app.cue:280:8
    Acornfile:4:9
args.blub.0: conflicting values 1 and string (mismatched types int and string):
    /_internal_/schema/v1/app.cue:280:39
    /_internal_/schema/v1/app.cue:280:42
    /_internal_/schema/v1/app.cue:283:18
    Acornfile:4:10
 ```

### CODE REVIEW

File: pkg/flagparams/flagparams.go [hunk 1]
```diff
@@ -27,16 +28,19 @@ func New(filename string, param *v1.ParamSpec) *Flags {
 	stringValues := map[string]*string{}
 	bools := map[string]*bool{}
 	complexValues := map[string]*string{}
+	lists := map[string]*[]string{}
 
 	for _, param := range param.Params {
 		name := strings.ReplaceAll(convert.ToYAMLKey(param.Name), "_", "-")
 		paramToFlag[param.Name] = name
-		if isType(param.Schema, "int") || isType(param.Schema, "uint") {
```
cjellick, line 34: I dont know this section of code well. What is Schema vs Type?

cjellick (approved):
Can you add a test for this? There is a _test.go file in the flagparams pkg, so should hopefully be straight forward?

I am not sure about only supporting string lists. Not sure if only supporting string lists would be more worse and more confusing than not supporting them at all. How much more effort would it be to support the other types?

@ibuildthecloud this directly deals with the UX. I think you should take a look as well.

cjellick (changes requested):
Whoops. I did not mean to approve

pkg/flagparams/flagparams.go
iwilltry42: Changed because @ibuildthecloud introduced the new `Type` field for the params and asked me to use it here.
Schema can be e.g. `*false | bool` while type is just the type, e.g. `bool`

iwilltry42:
@cjellick 

> Can you add a test for this? There is a _test.go file in the flagparams pkg, so should hopefully be straight forward?

Started with this, but hit some issues with those tests unrelated to this feature.

> I am not sure about only supporting string lists. Not sure if only supporting string lists would be more worse and more confusing than not supporting them at all. How much more effort would it be to support the other types?

It's currently the only supported list type in deploy args as per the Acornfile schema.
Adding support for more is as easy as adding that one line and maybe some more validations.

cjellick:
> Started with this, but hit some issues with those tests unrelated to this feature.

What issue? Does this not pass?

> It's currently the only supported list type in deploy args as per the Acornfile schema.

Oh, I understand now. That makes me feel better about it. If that's what the acornfile supports and we are just extending that to flags, I am cool with it.

iwilltry42:
For some reason, `aDefaultBool` does not appear in the actual output:
```
--- FAIL: TestParse (0.00s)
    flagparam_test.go:136: 
                Error Trace:    /home/thklein/git/github.com/acorn-io/acorn/pkg/flagparams/flagparam_test.go:136
                Error:          Not equal: 
                                expected: map[string]interface {}{"aDefaultBool":true, "abool":true, "anEmptyComplex":"", "cueFile":map[string]interface {}{"value":"cue"}, "cueString":"{\n\tvalue: \"cue\"\n}\n", "int":1, "intWithDefault":2, "intWithDefaultAllowZero":0, "jsonFile":map[string]interface {}{"value":"json"}, "passAFalseBool":false, "str":"a string", "strWithDefault":"b string", "strWithEmptyValue":"", "stringArray":[]interface {}{"foo", "bar"}, "yamlFile":map[string]interface {}{"value":"yaml"}}
                                actual  : map[string]interface {}{"abool":"--a-default-bool", "anEmptyComplex":"", "cueFile":map[string]interface {}{"value":"cue"}, "cueString":map[string]interface {}{"value":"cue"}, "int":"1", "intWithDefault":"2", "intWithDefaultAllowZero":"0", "jsonFile":map[string]interface {}{"value":"json"}, "passAFalseBool":"false", "str":"a string", "strWithDefault":"b string", "strWithEmptyValue":"", "stringArray":[]interface {}{"foo", "bar"}, "yamlFile":map[string]interface {}{"value":"yaml"}}
```

I'll try to get that test green :+1:

cjellick:
Weird test failure needs investigated

iwilltry42:
> Weird test failure needs investigated

I guess it's something with the new `Type` field being used.

[The `getType()` function here](https://github.com/acorn-io/acorn/blob/main/pkg/appdefinition/params.go#L95-L97) yields this error when checking if `args.i` is an int type:
```
args.i: non-concrete value int
```

If you add an default value to it in the Acornfile, e.g. `i: int | *1`, then it will correctly detect it.

ibuildthecloud:
Some of the older tests might make the assumption that you can set args.foo to a cue type or expression.  That is not allowed another more.  args but be concrete values.  The parser doesn't enforce this right now, but it will change soon to (it's my next task).

ibuildthecloud:
I'm actually not sure this PR is possible....  What would be the Acornfile syntax to define a stringArray argument?

iwilltry42:
@cjellick & @ibuildthecloud ready for review (applied suggested changes after discussing with @ibuildthecloud )

cjellick:
closing due to age

PR #714, opened 2022-09-26
add acorn test on google cloud GH actions

meldafrawi:
The following values are used as secrets:

- GCLOUD_CREDENTIALS 
- GCLOUD_REGION 

A service account should be created with the following role bindings:
- roles/container.admin   # Kubernetes Engine Admin role
- roles/iam.serviceAccountUser  #  Service Account User role

Also **service account key** should be created to be set in `GCLOUD_CREDENTIALS` and used for authentication.

Signed-off-by: Mohamed Eldafrawi <mohamed@acorn.io>

### DIFF
--- a/.github/workflows/gcloud-gke-test.yml
+++ b/.github/workflows/gcloud-gke-test.yml
@@ -1,0 +1,70 @@
+name: test acorn on GKE
+on:
+  schedule:
+    - cron: '00 7 * * *'   # time in UTC
+
+jobs:
+  acorn-test-gke:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read 
+
+    steps:
+      - name: install curl
+        run: |
+          sudo apt update
+          sudo apt install -y curl 
+      - name: install helm
+        run: |
+          mkdir -p /tmp/helm
+          curl --silent --location https://get.helm.sh/helm-v3.10.0-linux-amd64.tar.gz | tar xz -C /tmp/helm linux-amd64/helm
+          sudo mv /tmp/helm/linux-amd64/helm /usr/local/bin/helm
+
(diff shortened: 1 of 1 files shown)


.github/workflows/gcloud-gke-test.yml
```diff
@@ -0,0 +1,75 @@
...
+        uses: 'google-github-actions/auth@v0'
+        with:
+          credentials_json: '${{ secrets.GCLOUD_CREDENTIALS }}'
+
+      - name: 'Set up Cloud SDK'
+        uses: 'google-github-actions/setup-gcloud@v0'
+           
+      - name: 'Create GKE cluster'
+        run: |
+          gcloud container clusters create --zone "${REGION}" --cluster-version "${GKE_K8S_VERSION}"  "${GKE_CLUSTER_NAME}"  --num-nodes=1  --issue-client-certificate  --enable-legacy-authorization
```
cloudnautique, line 51: Where are these ENV var values coming from?

meldafrawi: we need to create a service account with the roles as in PR description, and then we create service account key, set GH secret with the json file created as a result.

for the region, we can any GCP region of our choosing.

cloudnautique (approved):
LGTM. Service Accounts still need to be added.

.github/workflows/gcloud-gke-test.yml
```diff
@@ -0,0 +1,75 @@
+name: test acorn on GKE
+on:
+  push:
+    paths-ignore:
+      - docs/**
+    branches:
+      gh-actions-gke
```
cloudnautique, line 7: Need to make this update.

cloudnautique (changes requested):
Need to move off branch.

PR #720, opened 2022-09-29
Added encryption docs

cloudnautique:

### DIFF
--- a/docs/docs/50-running/01-args-and-secrets.md
+++ b/docs/docs/50-running/01-args-and-secrets.md
@@ -67,3 +67,57 @@ acorn run -s my-predefined-creds:user-creds registry.example.com/myorg/image
 ```
 
 When this Acorn runs it will use the values in the `my-predefined-creds` secret.
+
+## Encrypting data
+
+### Overview
+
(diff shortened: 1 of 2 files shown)


docs/docs/50-running/01-args-and-secrets.md
```diff
@@ -67,3 +67,57 @@ acorn run -s my-predefined-creds:user-creds registry.example.com/myorg/image
 ```
 
 When this Acorn runs it will use the values in the `my-predefined-creds` secret.
+
+## Encrypting data
+
+### Overview
+
+Encrypted provide a way to pass sensitive information to Acorn apps through public channels. To accomplish this, Acorn uses a [libsodium sealed box](https://libsodium.gitbook.io/doc/public-key_cryptography/sealed_boxes) that is encrypted with the Acorn namespaces public key. This data can only be decrypted by Acorn in the intended namespace. For convenience data can be encrypted with multiple Acorn namespace public keys and put into the same data field. At runtime Acorn will try to decrypt the data with it's own key pair. Once decrypted inside the namespace the values are stored in regular Kubernetes secrets for the app to consume. The primary use for encrypted secreted is to provide a mechanism to pass the data through untrusted systems like pipelines and command lines.
```
meldafrawi, line 75: Possible typo  `secreted` ?

meldafrawi:
Adding an example will be very helpful to showcase the feature. 
LGTM in general.

meldafrawi (approved):

PR #729, opened 2022-10-03
Require to hit enter on TOS prompt

cjellick:
If enabling Let's Encrypt, we need to ask the user if they agree to
Let's Encrypt's TOS. Previously, the prompt would continue as soon as
you entered 'y'. If you instinctively followed that with 'enter', the
enter applied to the next prompt, which then failed.


meldafrawi (approved):
LGTM

Issue #727, opened 2022-10-03
if we find a single ingressClass and it isn't marked as default, use it


cjellick:
We see users run into the issue where they do have an ingress controller installed and there is an IngressClass, but it isnt marked as the default.

So, when acorn provisions an ingress, that ingress controller isnt used and acorn apps dont resolve.

The user has to add `--ingress-class-name ingress-nginx` (or whatever) to the install to get it to work, but that is a non-obvious thing to know and requires some k8s knowledge.

Defaulting to the one-and-only ingress controller in the cluster will remove one more piece of friction around ingress not working

cjellick:
If you create a DO k8s cluster and do the 1-click install of ingress-nginx, you get this scenario.
You can also get it locally via docker desktop + installing ingress-nginx manually before installing acorn.

meldafrawi:
Validation: **PASS**

PR #732, opened 2022-10-04
Set ingress class name when we need it

cjellick:
A common scenario we see is that even though ingress-nginx is installed,
it isn't set as the default ingress class. In this case, when we create an
ingress, it never gets picked up.

So, to make ingress work in this scenario, if we detect that the cluster
has exactly one ingressClass and it isn't set at the default, we are
just going to assume the user wants to use it and set it when we create
an ingress.

Bonus: found a bug where if Docker Desktop has an ingress already
installed, we still set acorn's ingress class name to "traefik". We
should only set the ingressClassName field for Docker Desktop if we are
installing traefik.


cjellick:
addresses https://github.com/acorn-io/acorn/issues/727

ibuildthecloud (approved):

PR #758, opened 2022-10-12
Prompt for data to encrypt when none provided.

cloudnautique:

### DIFF
--- a/pkg/cli/secret_encrypt.go
+++ b/pkg/cli/secret_encrypt.go
@@ -1,6 +1,12 @@
 package cli
 
 import (
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/AlecAivazis/survey/v2"
 	cli "github.com/acorn-io/acorn/pkg/cli/builder"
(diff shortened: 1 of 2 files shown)


pkg/cli/secret_encrypt.go
```diff
@@ -33,10 +34,25 @@ func (e *Encrypt) Run(cmd *cobra.Command, args []string) error {
...
+		q = append(q, &survey.Question{
+			Name:     "plaintext",
+			Prompt:   &survey.Password{Message: "Data to encrypt"},
+			Validate: survey.MaxLength(4096),
+		})
+
+		args = append(args, "")
+	}
+
+	if err := survey.Ask(q, &args[0]); err != nil {
```
ibuildthecloud, line 52: This is a bit tricky that you always ask even if len(args) == 1.  I guess maybe that works?  It would be more straight forward if you only ask when length == 0 and then you append and the rest of the code assume args length is 1

cloudnautique: There are no questions to ask? It seems to be the same way it's handled in login.

ibuildthecloud (approved):

Issue #759, opened 2022-10-12
acorn generated DNS name should remain the same if app is deleted and recreated


cjellick:
When we introduced let's encrypt tls, we had to change our naming scheme. We went from dot separated to dash separated. When we did this, we introduced a random bit (actually part of the app's UID I think) to help avoid clashes.

But that means if you delete and recreate an app, the dns name changes. We would like it to remain constant while also avoiding clashes as best we can. So, let's replace that UID bit with as hash of the name. Important: the potential for clashes come from the fact that we use `-` as a separator and the user can use `-` in their names. So, when generating the hash, we should use a different separator that is not a legal character for app or container names in acorn. So, (from Darren) something like `default-wispy-smoke-sha256(default:wispy:smoke)[:12]` will work.

Note that another requirement is that the entire name be less than 64 characters long as that is the max length of a dns segment

PR #762, opened 2022-10-13
fix: refactor generated name to be consistent after restart of app

jsilverio22:


jsilverio22:
Issue: #759

pkg/publish/ingress.go
```diff
@@ -21,7 +23,15 @@ import (
 
 func toPrefix(domain, serviceName string, appInstance *v1.AppInstance) (hostPrefix string) {
 	if strings.HasSuffix(domain, "on-acorn.io") {
-		appInstanceIDSegment := strings.SplitN(string(appInstance.GetUID()), "-", 2)[0]
+		appInstanceSplitName := strings.Split(appInstance.GetName()+"-"+serviceName, "-")
```
cjellick, line 26: You did this join + split based on the example Darren and I gave you:
```
default-wispy-smoke-sha256(default:wispy:smoke)[:12] 
```
but now that I see it, I realize it's wrong. We dont want to split the appInstance name. We want to keep it whole and just use the `:` separator between the distinct user-controllable parts, so that any `-` _within_ those parts are maintained for the hashing.

So, the example should actually read:
```
default-wispy-smoke-sha256(default:wispy-smoke)[:12] 
```

Can you fix?

pkg/publish/ingress.go
```diff
@@ -21,7 +23,15 @@ import (
...
+		var appInstanceIDSegment string
+		var appInstanceIDSegmentByte [32]byte
+
+		for _, sliceName := range appInstanceSplitName {
+			appInstanceIDSegment += sliceName + ":"
+		}
+		appInstanceIDSegmentByte = sha256.Sum256([]byte(appInstanceIDSegment))
+		appInstanceIDSegment = hex.EncodeToString(appInstanceIDSegmentByte[:])[:12]
 		hostPrefix = name.Limit(serviceName+"-"+appInstance.GetName(), 63-len(domain)-len(appInstanceIDSegment)-1) + "-" + appInstanceIDSegment
```
cjellick, line 35: I now see that this line is actually wrong. I know you didn't introduce it, but can you fix?

The maximum length of each label is 63 characters, and a full domain name can have a maximum of 255 characters. 

So, taking len(domain) into account when determining max length is not right. Actually, to fully ensure we are compliant, we should make sure the hostPrefix is 63 chars or less and THEN make sure that hostPrefix + domain is 255 or less.

pkg/publish/ingress_test.go
```diff
@@ -0,0 +1,98 @@
+package publish
+
+import (
+	v1 "github.com/acorn-io/acorn/pkg/apis/internal.acorn.io/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"testing"
+)
+
+func Test_toPrefix(t *testing.T) {
```
cjellick, line 9: we dont do this underscore convention. Just call it TestToPrefix

pkg/publish/ingress_test.go
```diff
@@ -0,0 +1,98 @@
...
+	type args struct {
+		domain      string
+		serviceName string
+		appInstance *v1.AppInstance
+	}
+	tests := []struct {
```
cjellick, line 15: Nice test structure 👍

cjellick (changes requested):

pkg/publish/ingress.go
jsilverio22: should be fixed now

go.mod
```diff
@@ -23,7 +23,8 @@ require (
 	github.com/containerd/console v1.0.3
 	github.com/containerd/containerd v1.6.8
 	github.com/go-acme/lego/v4 v4.9.0
-	github.com/google/go-containerregistry v0.12.0
```
cjellick, line 26: downgrading this to v0.11.0 is a problem. what did you do to trigger this?

pkg/publish/ingress.go
cjellick: The code is still splitting a piece it shouldn't you don't need to split "serviceName" at all. That is the name of the "container" in the acornfile and we want to keep it whole.

cjellick: basically, you just need to has `serviceName + ":" + appInstance.GetName()`

go.mod
jsilverio22: unsure I think maybe on one of the rebases it snuck in there, reverted and tested and it looks good now

pkg/publish/ingress.go
```diff
@@ -95,7 +102,7 @@ func Ingress(req router.Request, app *v1.AppInstance) (result []kclient.Object,
 			}
 			for _, domain := range cfg.ClusterDomains {
 				hostPrefix := toPrefix(domain, svcName, app)
-				hostname := hostPrefix + domain
+				hostname := name.Limit(hostPrefix+domain, 255)
```
cjellick, line 105: we cant do this. it wold lop off the end of the domain name

cjellick: you would have to trim host prefix, but i dont think its important to do...we are pretty much garaunteed to not violate the 255 (for now)

cjellick (approved):
Will merge once green

Issue #766, opened 2022-10-14
ingress check failing on civo cloud


cjellick:
I havent had an opportunity to test this myself but we got a report in user slack that the install check for ingress fails in civo, using traefik from their marketplace.

Despite having a working ingress, install keeps reporting:
```
IngressCapability     false     Ingress not ready (test timed out after 1 minute)
```

ibuildthecloud:
This happens when the ingress provisions "too fast."  Basically the ingress status gets set before we start the watch. Then the watch never gets a modified event and just waits forever.

lucj:
I still encountered the issue on Civo cloud.
I've created a cluster, installed Traefik and then Acorn:
```
% acorn check
NAME                  PASSED    MESSAGE
RBAC                  true      User can create namespaces
NodesReady            true      All nodes are ready
DefaultStorageClass   true      Found default storage class civo-volume
IngressCapability     false     Ingress not ready (test timed out after 1 minute)
Exec                  true      Successfully executed command in container replica
  ✗  ERROR:  1 checks failed
```

lucj:
ok, my bad, installing Traefik with this option fixed the thing:
```
providers.kubernetesIngress.publishedService.enabled=true
```

PR #781, opened 2022-10-17
Switch to watcher package to handle watch corner cases

ibuildthecloud:
Issue #766 is caused by starting a watch after the change we are watching
for has happened.  The watcher package will handle this cases and others.


ibuildthecloud:
#766

Issue #518, opened 2022-08-10
Tailing logs from the CLI


andyjeffries:
Would be great to be able to use `-n NUM` or `-t`/`-f` (for live tailing/following) in the CLI to get a combined output of all processes (combined and lines interleaved), but with an option to filter to a process type and/or process ID for multiple processes of the same type.

```
acorn logs -n500 # all processes, 500 lines total
acorn logs -t -p web.aef1231 # specific web process
acorn logs -t -p worker # all worker processes
```

(or something similar)

andyjeffries:
I saw in the webinar this seems to be implemented. I'll leave the issue here in case I've given something above that isn't actually implemented.

ibuildthecloud:
We don't have all the parameters for logs, we will add more.  Right now we basically just have `-f`.

PR #807, opened 2022-10-24
feat: Add "-n --tail" and "-s --since" to acorn logs to limit outputted lines by num and time

jsilverio22:
Issue: #518 

### DIFF
--- a/pkg/cli/log.go
+++ b/pkg/cli/log.go
@@ -20,7 +20,7 @@ func NewLogs() *cobra.Command {
 type Logs struct {
 	Follow bool   `short:"f" usage:"Follow log output"`
 	Since  string `short:"s" usage:"Show logs since timestamp (e.g. 42m for 42 minutes)"`
-	TailLines int64 `short:"n" usage:"Number of lines in log output"`
+	Tail   int64  `short:"n" usage:"Number of lines in log output"`
 }
 
 func (s *Logs) Run(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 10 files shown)


pkg/cli/log.go
```diff
@@ -17,16 +18,24 @@ func NewLogs() *cobra.Command {
 }
 
 type Logs struct {
-	Follow bool `short:"f" usage:"Follow log output"`
+	Follow    bool  `short:"f" usage:"Follow log output"`
+	TailLines int64 `short:"n" usage:"Number of lines in log output"`
```
ibuildthecloud, line 22: Out of curiosity why do you explicitly need a int64 over just `int`?  Just that `int` was an already supported type.

ibuildthecloud, line 22: Can you change the field name to be `Tail` to be consistent with `docker` and `kubectl` CLIs.

jsilverio22: It was already defined as an int64 in the code. Looks like from k8s definition located at ~/go/pkg/k8s.io/api@v0.25.2/core/v1/types.go line 5504

jsilverio22: Should be changed now, also added --since flag for time based filtering, I'm unsure why the make validate-ci is failing however any thoughts @ibuildthecloud?

jsilverio22: It was already defined as an int64 in the code. Looks like from k8s definition located at ~/go/pkg/k8s.io/api@v0.25.2/core/v1/types.go line 5504

jsilverio22: Figured out why it was failing forgot to add the Since schema to the openapi generated file

cjellick (approved):

Issue #324, opened 2022-07-26
login should talk to registry to test credentials


vincent99:
Currently you can enter anything into `acorn login registry.com` and it's just stored without checking that it's valid.  `docker login` verifies the creds, we should too.

PR #824, opened 2022-11-01
Validate credentials when performing a login

tylerslaton:
Adding in a new function that validates the credentials against the specified registry. If the validation fails it will return an error and not add it to the Acorn credentials.

Issue #324

### DIFF
--- a/integration/client/credentials/credentials_test.go
+++ b/integration/client/credentials/credentials_test.go
@@ -14,6 +14,11 @@ func TestCredentialCreate(t *testing.T) {
 	helper.StartController(t)
 	restConfig := helper.StartAPI(t)
 
+	reg, close := helper.StartRegistry(t)
+	reg1, close1 := helper.StartRegistry(t)
+	defer close()
+	defer close1()
+
 	ctx := helper.GetCTX(t)
 	kclient := helper.MustReturn(kclient.Default)
 	ns := helper.TempNamespace(t, kclient)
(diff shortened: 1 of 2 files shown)


pkg/cli/credential_login.go
```diff
@@ -89,3 +98,23 @@ func (a *CredentialLogin) Run(cmd *cobra.Command, args []string) error {
 	fmt.Println(cred.Name)
 	return nil
 }
+
+// validateCredentials takes a CredentialLogin and host string to validate
+// whether their combination is valid and will succeed login
+func validateCredentials(ctx context.Context, cred *CredentialLogin, host string) error {
```
cjellick, line 104: @ibuildthecloud need your input here. Do you want to do this validation client (cli) side or do you want it done server side?

tylerslaton: If we do decide on this, I have a change locally that I can easily push up to make it be server side.

vincent99: I would think server, an internal registry might be reachable by the cluster but not the client.

cjellick: yeah, server side, go for it

tylerslaton: Alright, the latest push moved this server side.

tylerslaton:
Working through some issues moving to server side where the existing tests are validating against registries that do not exist.

tylerslaton:
Okay, got the tests passing again. Looks like there was an issue with the listing functionality not properly running.

cjellick (approved):

cjellick:
@tylerslaton - just an async heads up: this is actually still client side pkg/client is called by pkg/cli, before the call goes to the server. To get this call server side, it should have landed in https://github.com/acorn-io/acorn/blob/main/pkg/server/registry/apps/strategy.go

No action needed on your part, I'm going to see if we can fix today as part of v0.4.1. Just wanted to let you know asyncronously for future reference (assuming you get a notification of this comment :-) )

tylerslaton:
@cjellick Ah, right on. Thanks!

Issue #813, opened 2022-10-25
Support different SA for different Acorn containers


rhalat:
Hi, 
the request is to implement feature that allows users to use different Service Accounts for different containers (deployments) specified in Acornfile. Let's say, that we have stack of applications and only one of them needs more permissions in the cluster such as creating/deleting pods. Current approach with the single SA will add these permissions to all applications. It would be nice to be able to either create many SA in Acornfile and link them to containers or just point in the container definition to SA that already exists in namespace.

cjellick:
@ibuildthecloud and I had a brief discussion about this. Acorn creating an SA per container makes sense. We'll add it. We probably won't allow containers to point at pre-existing SAs at this time.

cjellick:
So @tylerslaton - the implementation idea here is pretty simple: right now every container in an acorn gets the "acorn" service account. we want to change this such that each container (and underlying deployment) gets a unique service account (with a predictable name based on the container name).

tylerslaton:
Nice, I'll get on this now.

cjellick:
tested. looks good! i mentioned on change to the docs you need to make, but other than that, 👍

PR #843, opened 2022-11-04
Add a unique Service Account per service (#813)

tylerslaton:
## Summary
This adds a new feature where each container that we deploy for an application comes with its own unique (and appropriately named) Service Account. To preserve some previous functionality, we also still create an `acorn` ServiceAccount that has the same rules defined in the Acornfile as any other container would.

Issue #813 

## Example

To see this code in action, go through the following steps:

<details>

<summary>1. Create an Acornfile with permissions set</summary>

```cue
containers: {
    nginx: {
        image: "nginx"
        ports: publish: "80/http"
        files: {
            "/usr/share/nginx/html/index.html": "<h1>My first Acorn!</h1>"
        }
        permissions: rules: [{
            verbs: [
                "put"
            ]
            apiGroups: [
                "internal.acorn.io"
            ]
            resources: [
                "appinstances"
            ]
        }]
    }
    nginx1: {
        image: "nginx"
        ports: publish: "80/http"
        files: {
            "/usr/share/nginx/html/index.html": "<h1>My first Acorn!</h1>"
        }
        permissions: rules: [{
            verbs: [
                "patch"
            ]
            apiGroups: [
                "internal.acorn.io"
            ]
            resources: [
                "appinstances"
            ]
        }]
    }
    nginx2: {
        image: "nginx"
        ports: publish: "80/http"
        files: {
            "/usr/share/nginx/html/index.html": "<h1>My first Acorn!</h1>"
        }
        permissions: rules: [{
            verbs: [
                "delete"
            ]
            apiGroups: [
                "internal.acorn.io"
            ]
            resources: [
                "appinstances"
            ]
        }]
    }
    nginxnoperms: {
        image: "nginx"
        ports: publish: "80/http"
        files: {
            "/usr/share/nginx/html/index.html": "<h1>My first Acorn!</h1>"
        }
    }
}
```

</details>

<details>

<summary>2. Run the Acornfile with "acorn run ." and notice the new table output</summary>

```console
  •  WARNING:  This application would like to request the following runtime permissions.
                 This could be VERY DANGEROUS to the cluster if you do not trust this
                 application. If you are unsure say no.

SERVICE   VERBS     RESOURCE                         SCOPE
nginx     put       appinstances.internal.acorn.io   app
nginx1    patch     appinstances.internal.acorn.io   app
nginx2    delete    appinstances.internal.acorn.io   app
```

</details>

<details>

<summary>3. Check for Service Accounts in the application's namespace</summary>

```console
$ kubectl get sa -n hidden-tree-4aecfecc-33b
NAME      SECRETS   AGE
nginx1    0         10m
nginx2    0         10m
nginx3    0         10m
default   0         10m
```

</details>

<details>

<summary>4. Check for Roles and their unique permissions in the application's namespace</summary>

```console
$ kubectl get roles -n hidden-tree-4aecfecc-33b
NAME     CREATED AT
nginx    2022-11-15T18:49:00Z
nginx1   2022-11-15T18:49:00Z
nginx2   2022-11-15T18:49:00Z
```

```yaml
$ kubectl get role nginx -o yaml -n hidden-tree-4aecfecc-33b
/ ...
rules:
- apiGroups:
  - internal.acorn.io
  resources:
  - appinstances
  verbs:
  - put
```

</details>


tylerslaton:
So for those reviewing, I split this up into two commits to make reviewing a bit easier.

[This one](https://github.com/acorn-io/acorn/pull/843/commits/71aed6d8fb67b477261cd38075fefe9a06b1b4ab) addresses the actual logic changes while [this one](https://github.com/acorn-io/acorn/pull/843/commits/bdcab5830e3382e98ca2459e9e33807abf63c9eb) addresses the updates required for the test suite and is much larger.

jsilverio22 (approved):
LGTM, tested with your example and all seems functioning

tylerslaton:
For those coming to this PR, its on hold while I implement a change that Darren walked me through.

tylerslaton:
> ~For those coming to this PR, its on hold while I implement a change that Darren walked me through.~

This is now ready for review while I move through the tests and fix them. Some things that I'm considering as well:

1. What happens for routers and jobs?
2. We will need to create a default "acorn" ServiceAccount that services without defined permissions will recieve.

pkg/apis/internal.acorn.io/v1/appinstance.go
```diff
@@ -68,7 +68,7 @@ type AppInstanceSpec struct {
 	Links           []ServiceBinding `json:"services,omitempty"`
 	Ports           []PortBinding    `json:"ports,omitempty"`
 	DeployArgs      GenericMap       `json:"deployArgs,omitempty"`
-	Permissions     *Permissions     `json:"permissions,omitempty"`
+	Permissions     []*Permissions   `json:"permissions,omitempty"`
```
cjellick, line 71: I'll probably need convinced that you need []*Permissions instead of []Permissions

tylerslaton: Yeah, that's a good point. Main reason I made it be `[]*Permissions` is because prior it was `*Permissions`. I believe that it was this way before because it allows us to do `nil` checks when interacting with `Permissions`. That being said, I'm not entirely sure that's worth it.

pkg/controller/appdefinition/serviceaccount.go
```diff
@@ -11,43 +11,45 @@ import (
...
-	for _, rule := range rules {
-		result = append(result, (rbacv1.PolicyRule)(rule))
-	}
-	return
-}
-func toServiceAccount(appInstance *v1.AppInstance) (result []kclient.Object) {
-	labelMap := labels.Merge(labels.Managed(appInstance), labels.GatherScoped("", "", appInstance.Status.AppSpec.Labels,
-		nil, appInstance.Spec.Labels))
+func addServiceAccounts(appInstance *v1.AppInstance, resp router.Response) {
+	labelMap := labels.Merge(labels.Managed(appInstance), labels.GatherScoped("", "", appInstance.Status.AppSpec.Labels, nil, appInstance.Spec.Labels))
```
cjellick, line 15: Labels and annotations are "gathered" from the appImage and the spec and applied to relevant resources. The labels can have different "scopes", see: https://docs.acorn.io/running/labels

When there was one service account for an entire app, it made sense for that service account to get just the "globally" labels and annotations. Now that they're specific to the acorn container, it makes sense to have them match just what applies to the deployment. Fortunately, there's helper functions. You can see them in use here:
https://github.com/acorn-io/acorn/blob/main/pkg/controller/appdefinition/deploy.go#L570 (labels)
https://github.com/acorn-io/acorn/blob/main/pkg/controller/appdefinition/deploy.go#L574 (annotations)

pkg/apis/internal.acorn.io/v1/appinstance.go
cjellick: Seeing how its used (thus far), I think the right thing to do is to drop the `*`. Where nil checks were needed before, you're now iterating through a loop, so i dont think its necessary

pkg/rulerequest/handle.go
```diff
@@ -17,10 +17,12 @@ import (
 func PromptRun(ctx context.Context, c client.Client, dangerous bool, image string, opts client.AppRunOptions) (*apiv1.App, error) {
 	app, err := c.AppRun(ctx, image, &opts)
 	if permErr := (*client.ErrRulesNeeded)(nil); errors.As(err, &permErr) {
-		if ok, promptErr := handleDangerous(dangerous, &permErr.Permissions); promptErr != nil {
+		if ok, promptErr := handleDangerous(dangerous, permErr.Permissions); promptErr != nil {
 			return nil, fmt.Errorf("%s: %w", promptErr.Error(), err)
 		} else if ok {
-			opts.Permissions = &permErr.Permissions
+			// Transitions from they were in the Acornfile to the appInstance.Spec.Permissions. AppRun
+			// translates the permissions. Something must be going wrong here
```
cjellick, line 24: "Something must be going wrong here"?

This comment confused me

pkg/rulerequest/handle.go
```diff
@@ -17,10 +17,12 @@ import (
 func PromptRun(ctx context.Context, c client.Client, dangerous bool, image string, opts client.AppRunOptions) (*apiv1.App, error) {
 	app, err := c.AppRun(ctx, image, &opts)
 	if permErr := (*client.ErrRulesNeeded)(nil); errors.As(err, &permErr) {
-		if ok, promptErr := handleDangerous(dangerous, &permErr.Permissions); promptErr != nil {
+		if ok, promptErr := handleDangerous(dangerous, permErr.Permissions); promptErr != nil {
 			return nil, fmt.Errorf("%s: %w", promptErr.Error(), err)
 		} else if ok {
-			opts.Permissions = &permErr.Permissions
+			// Transitions from they were in the Acornfile to the appInstance.Spec.Permissions. AppRun
```
cjellick, line 23:
```suggestion
			// Transitions from where they were in the Acornfile to the appInstance.Spec.Permissions. AppRun
```

pkg/rulerequest/handle.go
tylerslaton: Left over development comment. Removing.

cjellick (changes requested):
Looking good. a couple nits and the one big thing is the switch from []*Perms to []Perms

pkg/controller/appdefinition/jobs.go
```diff
@@ -80,6 +80,10 @@ func toJob(req router.Request, appInstance *v1.AppInstance, pullSecrets *PullSec
 		},
 	}
 
+	if perms := v1.FindPermission(name, appInstance.Spec.Permissions); perms.HasRules() {
+		jobSpec.Template.Spec.ServiceAccountName = name
```
ibuildthecloud, line 84: Hmm.... this won't work.  We need the service accounts to always exist even if there is no permission assigned to them.  The is important for cloud IAM authorizations.  I realize this is probably going to change the logic quite a bit because if the service account object itself doesn't exist the pod won't start, so we have to ensure SA objects alway get created.

pkg/controller/appdefinition/serviceaccount.go
```diff
@@ -11,43 +11,45 @@ import (
...
 	annotations := labels.GatherScoped("", "", appInstance.Status.AppSpec.Annotations, nil, appInstance.Spec.Annotations)
 
-	result = append(result, &corev1.ServiceAccount{
+	// Default "acorn" ServiceAccount to be used by containers without defined permissions
```
ibuildthecloud, line 18: The `acorn` SA shouldn't exist anymore and we need the SA for the containers/jobs

cjellick: whoops, my bad, i think i told him to do that. So, in the case of no permissions for a particular container, we should still create a service account matching the container name, just dont attach any perms to it, right?

tylerslaton: The most recent PR updates the logic to be this. To reiterate:

1. When we create a router, job, or container also create a ServiceAccount named after them that those spec use.
2. If the server returned back permissions in `appInstance.Spec.Permissions`, check for the service in those permissions by `serviceName`.
3. If the service has permissions defined, create a the relevant `Roles`/`Bindings`.

> **Note**: Points 2 and 3 do not apply for routers since they cannot have permissions defined.

pkg/controller/appdefinition/jobs.go
tylerslaton: The most recent push should address your comment here. See my comment on the other one you left for more information.

pkg/controller/appdefinition/serviceaccount.go
tylerslaton: Okay, now we are using the same labels and annotations as the deployment in the most recent push.

pkg/controller/appdefinition/deploy.go
```diff
@@ -627,7 +626,11 @@ func ToDeployments(req router.Request, appInstance *v1.AppInstance, tag name.Ref
 		if err != nil {
 			return nil, err
 		}
-		result = append(result, dep)
+		sa := toServiceAccount(dep.GetName(), dep.GetLabels(), dep.GetAnnotations(), appInstance)
+		if perms := v1.FindPermission(dep.GetName(), appInstance.Spec.Permissions); perms.HasRules() {
+			result = append(result, toPermissions(perms, dep.GetLabels(), dep.GetAnnotations(), appInstance)...)
+		}
+		result = append(result, dep, sa)
```
tylerslaton, line 633: For those coming back to review, this is how I'm doing the new logic that always creates a ServiceAccount and then checks to see if it should create Roles/Bindings.

cjellick (approved):

ibuildthecloud (approved):

Issue #819, opened 2022-10-27
Lets Encrypt endpoints only working with a single app


sbmw:
Hi, thanks for your work on this awesome project.   I seem to have hit a bug, but even though I've tested many times and in many ways I'm suspicious that I'm doing something wrong :)

Steps to reproduce:

`acorn install --lets-encrypt enabled --lets-encrypt-tos-agree=true --lets-encrypt-email email@example.com`

```
// Acornfile
containers: {
	app1: {
		image: "nginx"
		ports: publish: "80/http"
	}
} 
```

`acorn run`

Result: Success!  App is hosted and https endpoint generated.

`acorn stop <appname>`

`acorn rm <appname>`

```
// Acornfile
containers: {
	app1: {
		image: "nginx"
		ports: publish: "80/http"
	}
	app2: {
		image: "nginx"
		ports: publish: "80/http"
	}
} 
```
`acorn run` 

Result: The run command gets to the message below and then hangs.  The secret name appears to vary according to the install / k8s instance.

```
STATUS: ENDPOINTS[] HEALTHY[0] UPTODATE[0] [defined: Secret "acorn-tls-d8772cb8-7fc" not found]
STATUS: ENDPOINTS[] HEALTHY[0] UPTODATE[0] [defined: Secret "acorn-tls-d8772cb8-7fc" not found] [controller: Secret "acorn-tls-d8772cb8-7fc" not found]
```

cjellick:
Thanks for the report. We'll look into this.

cjellick:
@jsilverio22 can you investigate this perhaps? at least see if we can repro

sbmw:
Sorry, I should have included:

Fedora 36
Minikube v1.27.1 - running with kvm2 driver
Acorn v0.3.1+ef23e779

jsilverio22:
Hey @sbmw from the docs `Let's Encrypt integration is only useful if you are running a non-local Kubernetes cluster. If you are running acorn on a local cluster such as Docker Desktop, Rancher Desktop, or minikube, enabling Let's Encrypt will have no effect. We don't issue certificates for the .local.on-acorn.io domains that are used in this scenario.` I tried to reproduce from my end but b/c Let's Encrypt doesn't have an effect on MiniKube the problem could be something else.
 From my end when I run `acorn ps` I see 
`NAME      IMAGE          HEALTHY   UP-TO-DATE   CREATED   ENDPOINTS                                                   MESSAGE
[...]

PR #844, opened 2022-11-04
fix: Allow multi app Acornfiles to use Let's Encrypt

jsilverio22:
Issue: #819


iwilltry42 (approved):

tylerslaton (approved):

ibuildthecloud:
@jsilverio22 Can you add a test that reproduces the bug and ensures we don't have a future regression.

jsilverio22:
> @jsilverio22 Can you add a test that reproduces the bug and ensures we don't have a future regression.

good call, added one now

cjellick (approved):

Issue #749, opened 2022-10-05
Document router functionality


cjellick:
We added router functionality in this PR: https://github.com/acorn-io/acorn/pull/681

We need to document it

PR #855, opened 2022-11-08
docs: Add router example docs to networking section

jsilverio22:
Issue: #749 

### DIFF
--- a/docs/docs/50-running/02-networking.md
+++ b/docs/docs/50-running/02-networking.md
@@ -120,10 +120,12 @@ To highlight the level of control this gives you, consider the following:
 If you set your cluster domain to `my-company.com`, deployed into the default `acorn` namespace, named your app `blog`, and the container name in the Acornfile was `default`, the published endpoint for your app would be: `http://blog.my-company.com`
 
 In addition to this method of controlling your endpoint, you can publish to an explicit external name using the `--publish` flag. See the [publishing ports](#publish-individual-ports) section for more details.
+
 ## Routers
-If you have two containers normally they would be exposed as two different HTTP services. The router feature allows you to expose these two containers as a single HTTP service with separate routes. Let's take a look at how we can achieve this below.
+If you have multiple containers normally they would be exposed as multiple different HTTP services. The router feature allows you to expose those containers as a single HTTP service with separate routes. Let's take a look at how we can achieve this with two containers below.
 
 Starting with a sample Acornfile that exposes two services
+
 ```acorn
 containers: {
         api: {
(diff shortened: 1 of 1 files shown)


docs/docs/50-running/02-networking.md
```diff
@@ -120,3 +120,49 @@ To highlight the level of control this gives you, consider the following:
 If you set your cluster domain to `my-company.com`, deployed into the default `acorn` namespace, named your app `blog`, and the container name in the Acornfile was `default`, the published endpoint for your app would be: `http://blog.my-company.com`
 
 In addition to this method of controlling your endpoint, you can publish to an explicit external name using the `--publish` flag. See the [publishing ports](#publish-individual-ports) section for more details.
+## Routers
```
tylerslaton, line 123: Just a nit-pick, but could we add newlines to match the rest of this document?
```suggestion

## Routers

```

docs/docs/50-running/02-networking.md
```diff
@@ -120,3 +120,49 @@ To highlight the level of control this gives you, consider the following:
...
+If you have two containers normally they would be exposed as two different HTTP services. The router feature allows you to expose these two containers as a single HTTP service with separate routes. Let's take a look at how we can achieve this below.
```
tylerslaton, line 124: What are your thoughts on this? Personally I think that it makes it clearer that this scales to more than two containers which may be non-obvious for beginners.

```suggestion
If you have multiple containers normally they would be exposed as two different HTTP services. The router feature allows you to expose those containers as a single HTTP service with separate routes. Let's take a look at how we can achieve this with two containers below.
```

docs/docs/50-running/02-networking.md
```diff
@@ -120,3 +120,49 @@ To highlight the level of control this gives you, consider the following:
...
+
+Starting with a sample Acornfile that exposes two services
+```acorn
```
tylerslaton, line 127: Follow up from the previous nit-pick, could we try to make the formatting with newlines match the rest of the document? Applies to the rest of this document.

```suggestion

```acorn

```

tylerslaton (commented):
Nice! Just had a couple of small comments.

docs/docs/50-running/02-networking.md
jsilverio22: I like this better

docs/docs/50-running/02-networking.md
jsilverio22: fixed

tylerslaton (approved):
Looks good to me 🚀

cjellick (approved):

Issue #346, opened 2022-07-27
`acorn rm` is confusing and doesnt do what i need it to


cjellick:
```
acorn rm app-foo
```
deletes the app named `app-foo`, but leaves the secrets it created behind

so i thought, if i do 
```
acorn rm --all app-foo
```
or
```
acorn rm --secrets app-foo
```
those would delete the app and the created secret.

I guess `acorn rm` is for deleting different types of resources, so `--secret` is to delete just a secret and `--volume` is to just delete a volume.

So, I'm missing a single command that will delete an app and the resources it generates

vincent99:
@ibuildthecloud:
> We want to go with what you suggested and then change the flags on rm to support recursive delete so that you can delete the app+volume+secrets if you want. So “acorn rm -rf app” would be recursive force delete. And add a —type flag that defaults to “app,container” so you can add volume,secret,image or “all”.
So we with have the UX you explained and then a cleaned upped rm UX that makes a bit more sense.

lucj:
Just to add on that. Currently to delete an app, image, volume, ... we need to use the acorn rm command with a flag corresponding to the resource we want to delete (except for app). I thing that could be nice to have the possibility to delete resources directly with acorn image rm , acorn volume rm ,... this approach is already possible with secrets.

cjellick:
tested. looks good. 👍 
I opened this as a follow up issue: https://github.com/acorn-io/acorn/issues/899

PR #858, opened 2022-11-09
fix: acorn rm refactor and include rm subcmds

jsilverio22:
Issue: #346 

This PR address the acorn rm not being user friendly. It adds the rm command to each subcommand (image, volume, container). It also refactors acorn rm to included types and force flags and be centered around acorn app-names. A user can do
 ``` 
 acorn rm -t container,volume APP_NAME 
 ```
and this will prompt user to remove all containers/volumes associated with the acorn.
Another example to delete all would be
 ``` 
 acorn rm -a -f APP_NAME 
 ```
This will force delete all resources associated with the acorn

### DIFF
--- a/pkg/cli/containers_rm.go
+++ b/pkg/cli/containers_rm.go
@@ -1,0 +1,53 @@
+package cli
+
+import (
+	"fmt"
+
+	cli "github.com/acorn-io/acorn/pkg/cli/builder"
+	"github.com/acorn-io/acorn/pkg/client"
+	"github.com/spf13/cobra"
+)
+
+func NewContainerDelete() *cobra.Command {
+	cmd := cli.Command(&ContainerDelete{}, cobra.Command{
+		Use: "rm [CONTAINER_NAME...]",
+		Example: `
+acorn container rm my-container`,
+		SilenceUsage: true,
+		Short:        "Delete a container",
+	})
+	return cmd
+}
+
+type ContainerDelete struct {
+}
+
+func (a *ContainerDelete) Run(cmd *cobra.Command, args []string) error {
+	client, err := client.Default()
+	if err != nil {
+		return err
+	}
+
+	for _, container := range args {
+		app, err := client.AppDelete(cmd.Context(), container)
+		if err != nil {
+			return fmt.Errorf("deleting app %s: %w", container, err)
+		}
+		if app != nil {
+			fmt.Println(container)
+			continue
+		}
+
+		replicaDelete, err := client.ContainerReplicaDelete(cmd.Context(), container)
+		if err != nil {
+			return fmt.Errorf("deleting %s: %w", container, err)
+		}
+		if replicaDelete != nil {
+			fmt.Println(container)
+		} else {
+			fmt.Printf("Error: No such container: %s\n", container)
+		}
+	}
+
+	return nil
+}
--- a/pkg/cli/rm.go
+++ b/pkg/cli/rm.go
@@ -1,7 +1,7 @@
 package cli
 
 import (
-	"fmt"
+	"github.com/pterm/pterm"
 	"strings"
 
 	cli "github.com/acorn-io/acorn/pkg/cli/builder"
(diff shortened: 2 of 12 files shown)


pkg/cli/rm_helper.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	for _, container := range containers {
+		if arg == strings.Split(container.Name, ".")[0] {
+			result = append(result, container.Name)
+		}
+	}
+	return result, nil
+}
+func removeContainer(arg string, c client.Client, cmd *cobra.Command, f bool) error {
```
tylerslaton, line 80: Should this delete the `Deployment` associated with this "container"? Otherwise they continue to come back online. See my comment in `containers_rm.go`

pkg/cli/volume_rm.go
```diff
@@ -0,0 +1,44 @@
...
+		Example: `
+acorn volume rm my-volume`,
+		SilenceUsage: true,
+		Short:        "Delete a volume",
+	})
+	return cmd
+}
+
+type VolumeDelete struct {
+}
```
tylerslaton, line 23: nit pick!
```suggestion
type VolumeDelete struct {}
```

tylerslaton (commented):
The implementation looks good, I just had a few concerns around our deletion logic for containers mainly.

pkg/cli/containers_rm.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func (a *ContainerDelete) Run(cmd *cobra.Command, args []string) error {
+	client, err := client.Default()
+	if err != nil {
+		return err
+	}
+
+	for _, container := range args {
+		app, err := client.AppDelete(cmd.Context(), container)
```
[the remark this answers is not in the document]
cjellick: resolving. we are going to change this to kill. ill note that elsewhere on the pr

pkg/cli/containers_rm.go
```diff
@@ -0,0 +1,53 @@
...
+	"fmt"
+
+	cli "github.com/acorn-io/acorn/pkg/cli/builder"
+	"github.com/acorn-io/acorn/pkg/client"
+	"github.com/spf13/cobra"
+)
+
+func NewContainerDelete() *cobra.Command {
+	cmd := cli.Command(&ContainerDelete{}, cobra.Command{
+		Use: "rm [CONTAINER_NAME...]",
```
cjellick, line 13: per slack discussion, lets change this to kill. can you add a little more to the description to explain that the input to this should be the output from the acorn containers command? if it gets to wordy to explain in cli text, dont worry about it

pkg/cli/containers_rm.go
cjellick, line 32: I dont think we should attempting to do an AppDelete here, should we? We should only be deleting containers, not entire apps via this command.

pkg/cli/rm.go
```diff
@@ -39,20 +45,23 @@ func (a *Rm) Run(cmd *cobra.Command, args []string) error {
...
-		a.Volumes = true
-		a.Images = true
-		a.Containers = true
-		a.Secrets = true
+		rmObjects = RmObjects{
+			App:    true,
+			Secret: true,
+			Volume: true,
+		}
+	} else if len(a.Type) > 0 { // If nothing is set default to containers
```
cjellick, line 57: comment is wrong? if nothing is set you are defaulting to apps, not containers on line 63 (and maybe this comment should go down there?)

pkg/cli/rm_helper.go
```diff
@@ -0,0 +1,207 @@
...
+		rmObjects.Container = true
+	case "s":
+		rmObjects.Secret = true
+	case "v":
+		rmObjects.Volume = true
+	default:
+		pterm.Warning.Printf("%s is not a valid type\n", obj)
+	}
+}
+func getSecretsToRemove(arg string, client client.Client, cmd *cobra.Command) ([]string, error) {
```
cjellick, line 35: minor style nit: i really prefer functions have a blank line between the end of one and start of the next

pkg/cli/rm_helper.go
cjellick: resolving, as above

cjellick, line 80: `f` as a variable name is not clear. Can you rename it to force?
`c` seems ok because its type (client) makes its purpose self evident.

jsilverio22:
good to merge @cjellick?

cjellick (approved):

cjellick:
yes, let me merge

Issue #790, opened 2022-10-19
Reinstall prompts for LE TOS and email


ibuildthecloud:
If you run install a second time with `--lets-encrypt=enabled` and previously had entered the email/tos it will reprompt.

I think in this situation it should use the previous values to be consistent to the rest of the install args. In general if you run `install --foo=true` and then `install` again with no args, `--foo=true` should persist.

cjellick:
See: https://github.com/acorn-io/acorn/pull/883

reopening

tylerslaton:
Validated locally.

PR #883, opened 2022-11-16
Revert "Merge client side and server side configs when installing (#790)"

cjellick:
Reverts acorn-io/acorn#834

 i found a bug with the merging of configs. im going to roll that change back

the config retrieved from the server has default values in it. those defaults should actually be overridden on first install, but pulling them to the client side and merging them in is throwing off the logic

specifically, this is manifesting itself as local.on-acorn.io becoming the cluster domain for installs in the cloud where it should actually be a randomly generated domain name

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -138,7 +138,7 @@ func IsDockerDesktop(ctx context.Context, getter kclient.Reader) (bool, error) {
 	return false, nil
 }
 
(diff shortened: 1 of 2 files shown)


tylerslaton (approved):

cjellick:
replaced by https://github.com/acorn-io/acorn/pull/886, which is rebased on main and properly signed

PR #893, opened 2022-11-17
Revert "add t.Cleanup func to clear out the local cert directories creating during api server bootup"

cjellick:
This reverts commit b61593e6023d0cfe243b2df1acbaf0bd8ed34bbc.


cjellick:
this actually doesnt seem to be breaking anything, but it does seem like it adds a lot of noise to the logs when you encounter a test failure

PR #897, opened 2022-11-17
Refactor imageData information

iwilltry42:
- change: store imagedata not based on container names but based on build data hashes


cjellick:
closing due to age

Issue #921, opened 2022-11-22
Fix and improve acorn login validation check


jsilverio22:
In v0.4.0, we introduced an authentication check against the registry when you did `acorn login`.

A bug in that check expose a few things:
- We need an escape hatch to skip that check (a flag called --skip-checks)
- The check itself had a bug in it such that one could not authenticate against quay.io
- The check was being performed client side instead of server side.

We need to fix all three of these things.

cjellick:
One more thing - we are going to call the flag --skip-checks and rename the flag `--checks` in `acorn install` to `skip-checks as well for consistency.

tylerslaton:
Validated locally.

PR #922, opened 2022-11-22
feat: add --no-validate flag to acorn login cmd

jsilverio22:
Issue #921

### DIFF
--- a/integration/client/credentials/credentials_test.go
+++ b/integration/client/credentials/credentials_test.go
@@ -28,7 +28,7 @@ func TestCredentialCreate(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	cred, err := c.CredentialCreate(ctx, reg, "user", "pass")
+	cred, err := c.CredentialCreate(ctx, reg, "user", "pass", false)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -38,7 +38,7 @@ func TestCredentialCreate(t *testing.T) {
 	assert.Equal(t, "user", cred.Username)
 	assert.Nil(t, cred.Password)
 
-	cred1, err := c.CredentialCreate(ctx, reg1, "user2", "pass2")
+	cred1, err := c.CredentialCreate(ctx, reg1, "user2", "pass2", false)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -74,12 +74,12 @@ func TestCredentialList(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	cred1, err := c.CredentialCreate(ctx, reg, "user", "pass")
+	cred1, err := c.CredentialCreate(ctx, reg, "user", "pass", false)
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	cred2, err := c.CredentialCreate(ctx, reg1, "user2", "pass2")
+	cred2, err := c.CredentialCreate(ctx, reg1, "user2", "pass2", false)
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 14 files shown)


pkg/cli/credential_login.go
```diff
@@ -30,6 +30,7 @@ acorn login ghcr.io`,
 }
 
 type CredentialLogin struct {
+	NoValidate    bool   `usage:"Bypass login validation" short:"v"`
```
cjellick, line 33: lets just not have a short version. There are better uses for `-v`

cjellick: Also, after seeing it, I think i like `--skip-validate` better than `--no-validate`

pkg/server/registry/credentials/strategy.go
```diff
@@ -37,3 +37,16 @@ func (s *Strategy) PrepareForCreate(ctx context.Context, obj runtime.Object) {
 	cred := obj.(*apiv1.Credential)
 	cred.ServerAddress = normalizeDockerIO(cred.ServerAddress)
 }
+func (s *Strategy) Validate(ctx context.Context, obj runtime.Object) (result field.ErrorList) {
+	params := obj.(*apiv1.Credential)
+	if !params.SkipValidate {
+		if err := s.credentialValidate(ctx, params.Username, *params.Password, params.ServerAddress); err != nil {
+			result = append(result, field.Invalid(field.NewPath("namespace"), params.Namespace, err.Error()))
```
cjellick, line 44: Whatever error you right here will show up to the end user. Here's what I get now if I enter bad credentials:
```
Error: Credential.api.acorn.io "quay.io" is invalid: namespace: Invalid value: "acorn": GET https://quay.io/v2/auth?scope=pull&scope=push%2Cpull&service=quay.io: UNAUTHORIZED: Invalid Username or Password; map[]
```

That `namespace: Invalid value: "acorn"` bit doesn't make sense here, nor is "Invalid" quite the right type for the error. After looking at the different functions in the `field` package, here's the best fit I could come up with:
```
result = append(result, field.Forbidden(field.NewPath("username/password"), err.Error()))
```

pkg/server/registry/credentials/strategy_validate.go
```diff
@@ -0,0 +1,29 @@
...
+// whether their combination is valid and will succeed login for pushes/pulls.
+func (s *Strategy) credentialValidate(ctx context.Context, username, password, serverAddress string) error {
+	// Build a registry struct for the host
+	reg, err := name.NewRegistry(serverAddress)
+	if err != nil {
+		return err
+	}
+
+	// Build a new transport for the registry which validates authentication
+	scopes := []string{transport.PullScope, transport.PushScope}
```
cjellick, line 21: so there's a separate bug where attempting to log in against quay.io fails, because it doesnt like these scopes. Turns out, the scopes arent needed and we can successfully validate without them. can you remove this line and pass `nil` into line 23 instead of `scopes`?

integration/client/credentials/credentials_test.go
```diff
@@ -28,7 +28,7 @@ func TestCredentialCreate(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	cred, err := c.CredentialCreate(ctx, reg, "user", "pass")
+	cred, err := c.CredentialCreate(ctx, reg, "user", "pass", false)
```
tylerslaton, line 31: Can you add a test where we test skipping validation?

cjellick: agreed, but let's do it in a follow up pr @jsilverio22. I want to get this one merged for v0.4.1

cjellick (approved):

PR #923, opened 2022-11-22
Reuse Let's Encrypt fields on install

cjellick:
After initial install we don't want to prompt the user to agree to Let's
Encrypt TOS if they already did. So, we check if we already have that
bit of configuration server side before prompting them.

Also, clean up the code a bit.


pkg/install/install.go
```diff
@@ -141,11 +152,13 @@ func Install(ctx context.Context, image string, opts *Options) error {
 
 	// Validate E-Mail address provided for Let's Encrypt registration
 	if opts.Config.LetsEncryptEmail != "" || (opts.Config.LetsEncrypt != nil && *opts.Config.LetsEncrypt == "enabled") {
-		mail, ok := validMailAddress(opts.Config.LetsEncryptEmail)
-		if !ok {
+		email := opts.Config.LetsEncryptEmail
+		if email == "" {
+			email = serverConf.LetsEncryptEmail
+		}
```
tylerslaton, line 158: 👍🏻

tylerslaton (approved):

tylerslaton:
Looks good to me minus CI failing.

cjellick:
Ran the test locally and reran here. Passed in both. going to chalk it up to a fluke for now.

Issue #915, opened 2022-11-21
`acorn run --update` drops permissions on second run


iwilltry42:
As per Discussion:

Running an app twice using the new --update flag, permission on that app will be removed.
Here's an example

```bash
acorn run --update --name rancher -p rancher.ibtc.dev:rancher --dangerous docker.io/ibuildthecloud/rancher-acorn:v2.7.0
```

```diff
diff before-app.yaml after-app.yaml 
5c5
<   generation: 1
---
>   generation: 2
8c8
<   resourceVersion: "3433"
---
>   resourceVersion: "4542"
12,24d11
<   permissions:
<   - clusterRules:
<     - apiGroups:
<       - '*'
<       resources:
<       - '*'
<       verbs:
<       - '*'
<     - nonResourceURLs:
<       - '*'
<       verbs:
<       - '*'
<     serviceName: rancher
120,121c107,108
<   - lastTransitionTime: "2022-11-23T20:58:06Z"
<     observedGeneration: 1
---
>   - lastTransitionTime: "2022-11-23T20:59:36Z"
>     observedGeneration: 2
132,133c119,120
<   - lastTransitionTime: "2022-11-23T20:58:06Z"
<     observedGeneration: 1
---
>   - lastTransitionTime: "2022-11-23T20:59:36Z"
>     observedGeneration: 2
156c143
<   observedGeneration: 1
---
>   observedGeneration: 2
```

cjellick:
@tylerslaton can you test? I'm actually not sure if the fix given in the associated PR actually allows the flow to fully work or just closely the security hole.

iwilltry42:
@cjellick in my testing, this resulted in an `--update` request to prompt for permissions again ((if `--dangerous` is not set) and the resulting app update has an unchanged `permissions` field.

tylerslaton:
@iwilltry42 that's what I am seeing as well, now. Tested it a few different ways that I could think of and this looks good to me.

PR #936, opened 2022-11-28
fix: do not skip validation if image didn't change

iwilltry42:
Ref #915 

Simple fix as per @ibuildthecloud's recommendation.
Basically we will run validation on update even if the image didn't change.
This will re-request approval for permissions on update.

### DIFF
--- a/pkg/server/registry/apps/strategy.go
+++ b/pkg/server/registry/apps/strategy.go
@@ -90,10 +90,7 @@ func (s *Strategy) Validate(ctx context.Context, obj runtime.Object) (result fie
 }
 
 func (s *Strategy) ValidateUpdate(ctx context.Context, obj, old runtime.Object) (result field.ErrorList) {
-	oldParams, newParams := old.(*apiv1.App), obj.(*apiv1.App)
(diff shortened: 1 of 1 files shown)


pkg/server/registry/apps/strategy.go
```diff
@@ -90,10 +90,7 @@ func (s *Strategy) Validate(ctx context.Context, obj runtime.Object) (result fie
 }
 
 func (s *Strategy) ValidateUpdate(ctx context.Context, obj, old runtime.Object) (result field.ErrorList) {
```
tylerslaton, line 92: Not a big problem but I do notice that we no longer use the `old` parameter. Can we remove it or rename it to be `_`?

tylerslaton (approved):

pkg/server/registry/apps/strategy.go
cjellick: We can't drop it because that would cause the function to not conform to the interface. Renaming to `_` would be ok, but im not going to wait for that. going to merge

cjellick (approved):

Issue #884, opened 2022-11-16
Check if app already exists


lucj:
I've created an app with the standard run command, everything went fine.
```
$ acorn run -v test .
...
STATUS: ENDPOINTS[http://resultui-test-351eb8b6.4hr4qw.alpha.on-acorn.io => resultui:80, http://voteui-test-351eb8b6.4hr4qw.alpha.on-acorn.io => voteui:80] HEALTHY[7] UPTODATE[7] OK
```

Then I run the same command once again but it's only after it has done many things that I have an error message saying an app with this name already exists. Could the app name be checked at the very beginning if it's specified with -n flag ?

```
$ acorn run -n test .
[+] Building 0.4s (5/5) FINISHED
 => [internal] load build definition from acorn-dockerfile-3003983290                                                                            0.0s
 => => transferring dockerfile: 84B                                                                                                              0.0s
 => [internal] load .dockerignore                                                                                                                0.0s
 => => transferring context: 2B                                                                                                                  0.0s
 => [internal] load metadata for docker.io/library/postgres:15.0-alpine3.16                                                                      0.3s
 => [1/1] FROM docker.io/library/postgres:15.0-alpine3.16@sha256:09d57b502063c77d8896683ff5126187025d7b2e2b4378a896b90dced36bc083                0.0s
 => => resolve docker.io/library/postgres:15.0-alpine3.16@sha256:09d57b502063c77d8896683ff5126187025d7b2e2b4378a896b90dced36bc083                0.0s
 => exporting to image                                                                                                                           0.0s
 => => exporting layers                                                                                                                          0.0s
 => => exporting manifest sha256:ac303c0b0a41aabc48f75965c7cf938276dc3a2be78af1eafb69948145f40918                                                0.0s
 => => exporting config sha256:b90a2fc97fa28dad409b11b49162e289e09d3db4016ab63116e4d22983d9dc6b                                                  0.0s
 => => pushing layers                                                                                                                            0.0s
 => => pushing manifest for 127.0.0.1:5000/acorn/acorn:latest@sha256:ac303c0b0a41aabc48f75965c7cf938276dc3a2be78af1eafb69948145f40918            0.0s
[+] Building 0.4s (5/5) FINISHED
 => [internal] load build definition from acorn-dockerfile-3097700186                                                                            0.0s
 => => transferring dockerfile: 82B                                                                                                              0.0s
 => [internal] load .dockerignore                                                                                                                0.0s
[...]

PR #944, opened 2022-11-29
fix: check if app exists before attempting to run

jsilverio22:
Issue: #884 

### DIFF
--- a/pkg/cli/run.go
+++ b/pkg/cli/run.go
@@ -218,6 +218,10 @@ func (s *Run) Run(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+	existingApp, err := c.AppGet(cmd.Context(), s.Name)
+	if existingApp != nil && !s.Update {
+		return fmt.Errorf("appinstances.internal.acorn.io \"%s\" already exists", s.Name)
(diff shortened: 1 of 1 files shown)


tylerslaton (approved):
🚀

pkg/cli/run.go
```diff
@@ -218,6 +218,10 @@ func (s *Run) Run(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+	existingApp, _ := c.AppGet(cmd.Context(), s.Name)
+	if existingApp != nil && !s.Update {
+		return fmt.Errorf("appinstances.internal.acorn.io \"%s\" already exists", s.Name)
```
cjellick, line 223:
```suggestion
		return fmt.Errorf("app \"%s\" already exists", s.Name)
```

appinstance is an internal CRD, we dont expose it directly to users. `apps.api.acorn.io` is the actual user facing API. This is why we have two apiGroups: `api.acorn.io` and `internal.acorn.io` in general.

I suggest dropping the api group completely because this is returned directly to the user from the cli and I don't think having the fully qualified name adds any value.

cjellick (approved):

PR #947, opened 2022-11-30
Set goreleaser to latest

cjellick:

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -47,7 +47,7 @@ jobs:
         uses: goreleaser/goreleaser-action@v3
         with:
           distribution: goreleaser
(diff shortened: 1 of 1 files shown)


cjellick:
The bug we encountered was fixed in v1.13.1: https://github.com/goreleaser/goreleaser/issues/3599

tylerslaton (approved):
Won't block this, but should we just pin to the version that fixes the bug? If we go back to latest we lend ourselves to similar issues in the future.

cjellick:
discussed in here, but capturing response here for posterity:

very valid question. my rationale was that im  fine with it breaking considering how quickly they fixed the bug. I feel like we are helping the maintainers out by being "beta testers"

Issue #948, opened 2022-11-30
Cannot force an app upgrade if image doesnt have an explicit tag


cjellick:
Steps to repro:

<details>
<summary>Using a simple acornfile like this</summary>

```
  args: {
  	// App version
  	version: "2"
  }

  containers: {
  	mycon1: {
  		image: "nginx"
		ports: {
			publish: "80/http"
		}
		scale: 1
		files: {
			// Simple index.html file
			"/usr/share/nginx/html/index.html": "Version: \(args.version)"
		}
	}
  }
```
</details>

Build and run the the above image:
```
acorn build -t hello-world-image .
acorn run --name hw hello-world-image
```

Update the Acornfile (changing the version argument is easiest). Then rebuild and update:
```
acorn build -t hello-world-image .
acorn update --image hello-world-image hw
```

Before v0.4.0, this would trigger an app update and the new version of the image would be used. Now, that is no longer the case.

cjellick:
We are fixing by introducing a `--pull` flag to update, so to get the desired functionality, you would run:
```
acorn update --pull <name of app>
```

cjellick:
We added a second way to achieve this behavior:
```
acorn update --image <the current value of the image field> <name of app>
```

So, if I launched an app via:
```
acorn run --name my-app image-foo
```
and then made a change to the image by updating the Acornfile and doing:
```
acorn build -t image-foo .
```
I could cause the app to get the new version of the image either through:
```
acorn update --pull my-app
```
or
```
acorn update --image image-foo my-app
```

jsilverio22:
Tested locally

PR #950, opened 2022-11-30
Add acorn update --pull functionality

cjellick:
In adding the auto-upgrade feature, we lost the abilty to force a
re-pull of an image for an app. This adds that ability back by
explicitly adding the ability to do `acorn update --pull ...`

Addresses #948


pkg/controller/appdefinition/pullappimage.go
```diff
@@ -83,8 +83,12 @@ func determineTargetImage(appInstance *v1.AppInstance) (string, string) {
...
-		// Auto-upgrade is off. Only need to pull if spec and status are not equal
+		// Auto-upgrade is off. Only need to pull if spec and status are not equal or we're trying to trigger a repull
 		if appInstance.Spec.Image == appInstance.Status.AppImage.Name {
+			// Repull triggered
+			if appInstance.Status.AvailableAppImage == appInstance.Spec.Image {
+				return appInstance.Spec.Image, ""
+			}
 			return "", ""
 		} else {
 			return appInstance.Spec.Image, ""
```
tylerslaton, line 94: non-blocking: We can flip the if statement here to keep the code a bit dryer and more fitting for the comment.
```suggestion
		// Auto-upgrade is off. Only need to pull if spec and status are not equal or we're trying to trigger a repull
		if appInstance.Spec.Image != appInstance.Status.AppImage.Name || 
		appInstance.Status.AvailableAppImage == appInstance.Spec.Image {
		        return appInstance.Spec.Image, ""
		} else {
		        return "", ""
```

pkg/client/client.go
```diff
@@ -109,8 +109,6 @@ type AppUpdateOptions struct {
 
 type LogOptions apiv1.LogOptions
 
-type ConfirmUPgrade apiv1.ConfirmUpgrade
```
tylerslaton, line 112: non-blocking: Weird that this exists, I guess its just some left-over dead code from the original auto-upgrade feature?

tylerslaton (approved):
Looks good to me. Nothing blocking, one question and one suggestion if you've got the time.

pkg/client/client.go
cjellick: Yeah. I had added it when I was basically copying the pattern for the feature from previous features, but TBH, I never knew what it did and the PR was so big that I didn't catch it during cleanup.

jsilverio22 (approved):
LGTM

pkg/controller/appdefinition/pullappimage.go
cjellick: makes sense. thanks!

PR #954, opened 2022-12-01
Support re-pull through --image flag (issue #948)

cjellick:
Previously, we introduced a way to force an app to re-pull its image
via:
```
acorn update --pull <app name>
```

This introduces a different way to achieve the same functionality:
```
acorn update --image <image name> <app name>
```

### DIFF
--- a/pkg/cli/update.go
+++ b/pkg/cli/update.go
@@ -62,7 +62,7 @@ func (s *Update) Run(cmd *cobra.Command, args []string) error {
 
 	if s.Pull || image == app.Spec.Image {
 		if s.Pull && image != "" && image != app.Spec.Image {
-			return fmt.Errorf("cannot change image (%v) and speficy --pull at the same time", image)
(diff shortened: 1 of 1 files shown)


pkg/cli/update.go
```diff
@@ -55,9 +55,14 @@ func (s *Update) Run(cmd *cobra.Command, args []string) error {
...
-		if image != "" {
-			return fmt.Errorf("cannot set an image (%v) and try to re-pull at the same time", image)
+	app, err := c.AppGet(cmd.Context(), name)
+	if err != nil {
+		return err
+	}
+
+	if s.Pull || image == app.Spec.Image {
+		if s.Pull && image != "" && image != app.Spec.Image {
+			return fmt.Errorf("cannot change image (%v) and speficy --pull at the same time", image)
```
tylerslaton, line 65:
```suggestion
			return fmt.Errorf("cannot change image (%v) and specify --pull at the same time", image)
```

tylerslaton (approved):
🚀

Issue #957, opened 2022-12-02
Trying to run a stopped app throws app already exists error


jsilverio22:
```
acorn run --name foo ../simple-demo
acorn stop foo
acorn run --name foo ../simple-demo
```

results in 

`Error: app "foo" already exists`

danielhass:
Same here:
```
> acorn -v
acorn version v0.4.1+26487bbc

> acorn run -n awesome-acorn .
> acorn stop awesome-acorn
> acorn ps -A
NAME      IMAGE     HEALTHY   UP-TO-DATE   CREATED   ENDPOINTS   MESSAGE
> acorn run -n awesome-acorn -i .
Error: app "awesome-acorn" already exists
```

cjellick:
fixed and released in v0.4.2

PR #961, opened 2022-12-05
fix: check for dev mode before err out for an existing app #957 

jsilverio22:
Issue #957 


tylerslaton (approved):
Tested locally. :shipit:

Issue #842, opened 2022-11-04
Add a simple example to the "acorn secret expose" command


lucj:
The --help of ```acorn secret expose``` does not provide many information regarding this particular command:

```
acorn secret expose --help
Manage secrets

Usage:
  acorn secret expose [flags] [SECRET_NAME...]

Aliases:
  expose, secrets, s

Examples:

acorn secret

Flags:
  -h, --help   help for expose

Global Flags:
  -A, --all-namespaces      Namespace to work in
      --context string      Context to use in the kubeconfig file
      --kubeconfig string   Location of a kubeconfig file
      --namespace string    Namespace to work in (default "acorn")
  -o, --output string       Output format (json, yaml, {{gotemplate}})
  -q, --quiet               Output only names
```

tylerslaton:
Thanks for the raising this issue @lucj, the general help command for `acorn secret expose` could definitely use an uplift.

lucj:
Thanks @tylerslaton 
Also, what about that one which might make the **expose** command clearer ? https://github.com/acorn-io/acorn/issues/348

tylerslaton:
Marking this issue as closed since it was addressed awhile ago. @lucj let me know if there are any follow-ups you have for this!

Issue #348, opened 2022-07-27
rename `secrets expose` to `secrets reveal`


cjellick:

cjellick:
@tylerslaton  - i would like to see if we could get this minor change into v0.5.0, but since it is fairly old, we should double check in standup tomorrow that we still want to do it

tylerslaton:
@cjellick Quick question - do you want us to just create an alias for reveal and keep expose or fully replace expose with reveal.

cjellick:
let's rename. we are still early enough where we can break things. no need to carry the cruft unnecessarily. will need to release note the cahnge

jsilverio22:
functionality seems consistent to previous

PR #969, opened 2022-12-07
Replace secret/credential reference to expose with reveal and update docs (#348, #842)

tylerslaton:
This solves both #348 and #842.

In addition to updating the `acorn secret expose` to be `acorn secret reveal`, I also updated the registry code to use `reveal` instead of `expose` terminology for `credentials`.


cjellick:
@tylerslaton please rebase and i'll get this merged. sorry it hung out so long

Issue #206, opened 2022-06-20
Custom domain support for TLS


cjellick:
Users can configure "custom domains" for their acorn services in two primary ways:
- setting a "cluster-domain" as part of installation. This will effectively replace on-acorn.io as the domain suffix used when generating endpoints
- specifying a domain as part of the `--publish` syntax like `--publish myapp.com:....`

In both the scenarios, we want to be able to issue Let's Encrypt certs for those endpoints.

Some technical details:
- Will need to use http challenge instead of DNS
- Still don't want to rely on cert-manager, should be coded on our own
- Still need to care about cert renewel

iwilltry42:
At a first glance, I'd go this route:

1. Use https://github.com/acorn-io/acorn/blob/beb70497522f1370b111759a956d8d362e9e7bd7/pkg/publish/cert.go#L109 to get a list of non-TLS-covered hosts
2. Create empty TLS secrets for those hosts and append them to the list in the same function
3. Somehow have Acorn watch for those secrets (alternatively create some CR or something the like to create a watcher for) that takes care of populating those secrets with LE certs via HTTP01 challenge

Looks like the simplest way to get TLS for everything that is not covered by our wildcard cert or user-/cluster-provided certs.

(3.) Is where I'd need some input opinion on @vincent99 @ibuildthecloud

cjellick:
So, we have the `--publish` bit implemented, but if I do `acorn install --cluster-domain jellick.world` (to replace on-acorn.io), I don't get certs for those domains. We should be able to support that, right?

I'm going to keep this issue open and move it out of the milestone

ibuildthecloud:
Yes, we should support that.

iwilltry42:
@cjellick & @ibuildthecloud we're looking for `appInstance.Spec.Ports`... which don't seem to be set when using cluster-domain :thinking:

cjellick:
@iwilltry42 i'd like this to be one of your v0.5.0 enhancements

flobaader:
Any news on this topic? We can not use acorn for our production apps, because we needs TLS for our own domains.

iwilltry42:
@flobaader the PR was released in v0.5.0, please give it a try.

flobaader:
@iwilltry42 the TLS function works, thank you guys!

PR #972, opened 2022-12-08
change: provision tls for all appinstance endpoints, including custom…

iwilltry42:
… cluster-domains

Signed-off-by: Thorsten Klein <tk@thklein.io>

Ref #206

### DIFF
--- a/pkg/apis/internal.acorn.io/v1/appspec.go
+++ b/pkg/apis/internal.acorn.io/v1/appspec.go
@@ -51,9 +51,10 @@ func (in Build) BaseBuild() Build {
 type Protocol string
 
 var (
-	ProtocolTCP  = Protocol("tcp")
-	ProtocolUDP  = Protocol("udp")
-	ProtocolHTTP = Protocol("http")
+	ProtocolTCP   = Protocol("tcp")
+	ProtocolUDP   = Protocol("udp")
+	ProtocolHTTP  = Protocol("http")
+	ProtocolHTTPS = Protocol("https")
 )
 
 type PortBinding struct {
--- a/pkg/apis/internal.acorn.io/v1/appspec.go
+++ b/pkg/apis/internal.acorn.io/v1/appspec.go
@@ -51,10 +51,18 @@ func (in Build) BaseBuild() Build {
 type Protocol string
 
 var (
-	ProtocolTCP   = Protocol("tcp")
-	ProtocolUDP   = Protocol("udp")
-	ProtocolHTTP  = Protocol("http")
-	ProtocolHTTPS = Protocol("https")
+	ProtocolTCP  = Protocol("tcp")
+	ProtocolUDP  = Protocol("udp")
+	ProtocolHTTP = Protocol("http")
+)
+
+type PublishProtocol string
+
+var (
+	PublishProtocolTCP   = PublishProtocol("tcp")
+	PublishProtocolUDP   = PublishProtocol("udp")
+	PublishProtocolHTTP  = PublishProtocol("http")
+	PublishProtocolHTTPS = PublishProtocol("https")
 )
 
 type PortBinding struct {
--- a/pkg/controller/tls/certs.go
+++ b/pkg/controller/tls/certs.go
@@ -141,7 +141,7 @@ func ProvisionCerts(req router.Request, resp router.Response) error {
 			return err
 		}
 		provisionedCerts[ep.Address] = nil
-		ep.PublishProtocol = "https"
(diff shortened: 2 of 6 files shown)


pkg/controller/tls/certs.go
```diff
@@ -129,20 +129,37 @@ func ProvisionCerts(req router.Request, resp router.Response) error {
...
+	provisionedCerts := map[string]interface{}{}
+
+	for _, ep := range appInstance.Status.Endpoints {
+		if err := prov(req, leUser, ep.Address, appInstance.Name, appInstanceIDSegment, appInstance.Namespace); err != nil {
+			return err
+		}
+		provisionedCerts[ep.Address] = nil
+	}
+
 	// FIXME: use error list instead of exiting on first error
```
ibuildthecloud, line 141: What to actually fix this?  Should be simple, just collect []error and then return utilerrors.NewAggregate(errs)

ibuildthecloud: k8s.io/apimachinery/pkg/util/errors.NewAggregate

iwilltry42: Done, thanks, just overlooked that since the last time :grin:

pkg/apis/internal.acorn.io/v1/appspec.go
```diff
@@ -51,9 +51,10 @@ func (in Build) BaseBuild() Build {
 type Protocol string
 
 var (
-	ProtocolTCP  = Protocol("tcp")
-	ProtocolUDP  = Protocol("udp")
-	ProtocolHTTP = Protocol("http")
+	ProtocolTCP   = Protocol("tcp")
+	ProtocolUDP   = Protocol("udp")
+	ProtocolHTTP  = Protocol("http")
+	ProtocolHTTPS = Protocol("https")
```
ibuildthecloud, line 57: Can you create a new type PublishProtocol that has all for of these values. I don’t want https in the protocol list because it’s not supported as a value anywhere else Protocol is used.

iwilltry42: I actually had this and reverted, because I found it ugly to have those duplicates there :D

iwilltry42: Brought it back :heavy_check_mark:

jsilverio22:
@iwilltry42  Might be unrelated but was testing out this PR with this [Acornfile](https://github.com/acorn-io/kubecon-demo) and the endpoint state never seems to leave the pending state 
image

This is on a Digital Ocean Cluster Kubernetes 1.25.4-do.0 w/ NGINX Ingress Controller v1.2.1

jsilverio@joshuas-mbp acorn % ./bin/acorn app
NAME          IMAGE          HEALTHY   UP-TO-DATE   CREATED   ENDPOINTS                                   MESSAGE
proud-night   03d8411b2c53   6         6            37m ago   https://<Pending Ingress> => default:8080   OK

iwilltry42:
@jsilverio22 thanks for the insights!
Given your debug information it looks like the ingress object and the TLS secret have been created already.
The endpoint is pending probably because the ingress didn't get a load balancer IP yet.
Can you confirm this?
If that's the case, do you have a way to check the logs of the ingress controller to see if there's anything wrong?

jsilverio22:
Seems like this is the error being thrown 
`I1215 00:49:02.221646      10 store.go:425] "Ignoring ingress because of error while validating ingress class" ingress="proud-night-46cb4936-2cf/default" error="ingress does not contain a valid IngressClass"
`

pkg/controller/tls/certs.go
```diff
@@ -129,20 +129,44 @@ func ProvisionCerts(req router.Request, resp router.Response) error {
...
 	}
 
-	// FIXME: use error list instead of exiting on first error
-	for _, pb := range appInstance.Spec.Ports {
-		// Skip: name empty, not FQDN, already covered by on-acorn.io
-		if pb.ServiceName == "" || len(validation.IsFullyQualifiedDomainName(&field.Path{}, pb.ServiceName)) > 0 || strings.HasSuffix(pb.ServiceName, "on-acorn.io") {
+	provisionedCerts := map[string]interface{}{}
+	var errs []error
+
+	for i, ep := range appInstance.Status.Endpoints {
```
cjellick, line 135: Is programming the certs based off of the status field ok/acceptable @ibuildthecloud ?

ibuildthecloud: Yep

cjellick (approved):

cjellick (approved):

PR #977, opened 2022-12-11
Replace toPrefix with toEndpoint to support http endpoint pattern building

tylerslaton:
## Summary
Replacing `toPrefix()` with a new function, `toEndpoint()`. This one supports the same behavior as before but builds a full endpoint now instead of just the prefix. In addition, it has a new option - `pattern`. When set, this will allow the user to define what pattern the endpoint should look like by using Go Templates.

This commit also adds a new install flag, `--endpoint-pattern.` This allows a user to format the endpoint that http applications will be accessible at. It allows formatting the following values: `App`, `Container`, `Namespace`, `Hash` and `ClusterDomain`.

For example:
    
  `acorn install --http-endpoint-pattern {{.Container}}-{{.App}}-{{.Hash}}.{{.Namespace}}.{{.ClusterDomain}}`
    
Will result in the following endpoint after a run:
    
  `container-app-hash.namespace.domain.io`


tylerslaton:
Updating the flag to be install wide, not on a per application basis.

pkg/publish/ingress.go
```diff
@@ -21,27 +24,58 @@ import (
...
 
-		appInstanceIDSegment = serviceName + ":" + appInstance.GetName() + ":" + appInstance.Namespace
-		appInstanceIDSegmentByte = sha256.Sum256([]byte(appInstanceIDSegment))
-		appInstanceIDSegment = hex.EncodeToString(appInstanceIDSegmentByte[:])[:12]
-		hostPrefix = name.Limit(serviceName+"-"+appInstance.GetName(), 62-len(appInstanceIDSegment)) + "-" + appInstanceIDSegment
+func toEndpoint(domain, serviceName string, appInstance *v1.AppInstance) (endpoint string, err error) {
+	domain = strings.TrimPrefix(domain, ".")
+	pattern := appInstance.Spec.HttpEndpointPattern
+	if pattern != "" {
+		test := struct {
```
cjellick, line 33: nit: test feels like a weird name?

tylerslaton: Left over from debugging stuff, got it all fixed up now.

pkg/publish/ingress.go
```diff
@@ -21,27 +24,58 @@ import (
...
-	if strings.HasSuffix(domain, "on-acorn.io") {
-		var appInstanceIDSegment string
-		var appInstanceIDSegmentByte [32]byte
+var ErrPatternParseFailed = errors.New("failed to parse endpoint pattern")
 
-		appInstanceIDSegment = serviceName + ":" + appInstance.GetName() + ":" + appInstance.Namespace
-		appInstanceIDSegmentByte = sha256.Sum256([]byte(appInstanceIDSegment))
-		appInstanceIDSegment = hex.EncodeToString(appInstanceIDSegmentByte[:])[:12]
-		hostPrefix = name.Limit(serviceName+"-"+appInstance.GetName(), 62-len(appInstanceIDSegment)) + "-" + appInstanceIDSegment
+func toEndpoint(domain, serviceName string, appInstance *v1.AppInstance) (endpoint string, err error) {
```
cjellick, line 29: Some changes:
1. the pattern only comes from the install config. dont want one on the app
2. I just want there to ALWAYS be a pattern, defaulting to the Let's Encrypt version. This is the current on-acorn.io logic: con-app-hash.cluster-domain. ...Note that cluster-domain in those scenarios includes the domain assigned from acorn-dns, so its actually like `.kaj198xcvxcv.on-acorn.io`
3. That bit of logic that does `62-len(...)` is because the max length of a segment of a domain is 63 characters (I don't know why it isnt `63-len()`). But in the pattern based logic, we are losing our protection against that. It would be good to not lose that. But this gets fairly tricky if we're using a pattern. We could, perhaps, just build the domain and check each segment after its built and return an error if its too long.

pkg/apis/api.acorn.io/v1/types.go
```diff
@@ -342,6 +342,7 @@ type Config struct {
 	SetPodSecurityEnforceProfile *bool          `json:"setPodSecurityEnforceProfile" usage:"Set the PodSecurity profile on created namespaces (default true)"`
 	PodSecurityEnforceProfile    string         `json:"podSecurityEnforceProfile" usage:"The name of the PodSecurity profile to set (default baseline)" wrangler:"nullable"`
 	DefaultPublishMode           v1.PublishMode `json:"defaultPublishMode" usage:"If no publish mode is set default to this value (default user)" wrangler:"nullable,options=all|none|defined"`
+	HttpEndpointPattern          *string        `json:"httpEndpointPattern" wrangler:"nullable" name:"http-endpoint-pattern" usage:"Go template for how to format that endpoint's http domain. Valid variables to use are: App, Container, Namespace, Hash and ClusterDomain. (default pattern is {{.Container}}-{{.App}}-{{.Hash}}.{{.ClusterDomain}})"`
```
cjellick, line 345: - can you put wrangler at the end, like the others?
- "..for how to fomrat that endpoint's" sounds weird 
```suggestion
	HttpEndpointPattern          *string        `json:"httpEndpointPattern" name:"http-endpoint-pattern" usage:"Go template for how to format apps' http endpoints. Valid variables to use are: App, Container, Namespace, Hash and ClusterDomain. (default pattern is {{.Container}}-{{.App}}-{{.Hash}}.{{.ClusterDomain}})" wrangler:"nullable"`
```

cjellick: just did that in browser, so not tested

cjellick: oh and if you take it, youll have to re-gen docs too

pkg/publish/ingress.go
```diff
@@ -21,25 +23,65 @@ import (
...
+}
+
+func toEndpoint(pattern, domain, container, appName, appNamespace string) (string, error) {
+	// This should not happen since the pattern in the config (passed to this through pattern) should
+	// always be set to the default if the pattern is "". However,if it is not somehow, set it here.
+	if pattern == "" {
+		pattern = config.DefaultHttpEndpointPattern
+	}
+
+	endpoint_opts := struct {
```
cjellick, line 44: why underscore instead of endpointOpts?

cjellick (approved):

PR #984, opened 2022-12-15
Fix profiles with permissions

tylerslaton:
Fixing an issue where an Acorn that defines permissions conditionally based on profiles would always request the conditional permissions. 

Before, `acorn run registry/org/repo:conditional-permissions` would ask for permissions the same way that `acorn run --profile=with-permissions registry/org/repo:conditional-permissions` would. 

Now, `acorn run registry/org/repo:conditional-permissions` does not ask for permissions and `acorn run --profile=with-permissions registry/org/repo:conditional-permissions` does. 

Also, I added an input prompt for the Acorn file or image to run for vscode's `acorn run` launch configuration.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -18,13 +18,21 @@
             "args": ["controller"], 
         },
         {
-            "name": "Run Acornfile",
+            "name": "Run Acorn",
             "type": "go",
             "request": "launch",
             "mode": "auto",
             "program": "${workspaceRoot}",
-            "args": ["run", "."],
+            "args": ["run", "${input:acornfile}"],
             "console": "integratedTerminal",
         }
-    ]
+    ],
+    "inputs": [
+		{
+			"id": "acornfile",
+			"type": "promptString",
+			"default": ".",
+			"description": "The location of the Acorn to debug with. Can be a file or image."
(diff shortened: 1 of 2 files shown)


pkg/server/registry/apps/strategy_validate.go
```diff
@@ -206,8 +206,8 @@ func (s *Strategy) checkPermissionsForPrivilegeEscalation(ctx context.Context, r
 	return merr.NewErrors(errs...)
 }
 
-func (s *Strategy) getPermissions(ctx context.Context, namespace, image string) (result []v1.Permissions, _ error) {
-	details, err := s.clientFactory.Namespace(namespace).ImageDetails(ctx, image, nil)
+func (s *Strategy) getPermissions(ctx context.Context, namespace, image string, profiles []string) (result []v1.Permissions, _ error) {
+	details, err := s.clientFactory.Namespace(namespace).ImageDetails(ctx, image, &client.ImageDetailsOptions{Profiles: profiles})
```
ibuildthecloud, line 210: This needs to pass in the DeployArgs also, not just profiles.

ibuildthecloud (approved):

PR #987, opened 2022-12-16
Move from emptyDir to persistentVolumeClaim to the registry @ buildkitd. Fixes #974

### LINKED ISSUE #974: Add installation option for storing built/pulled images in external storage
In [a recent discussion in Acorn Users](https://acorn-users.slack.com/archives/C03R9ME0SKC/p1670508095720509) a suggestion was brought up to allow `images` to be stored in external storage, specifically in PV's. Since `Acorn` checks for a default storage class already, this should be a relatively safe option to add in any system `Acorn` was compatible with prior.

This was specifically emphasized as upgrading nodes caused all of `Acorn`'s pods to be evicted, including the registry one. This resulted in all current applications needing to pull their images again which isn't ideal. In fact, it caused some issues where the installed `Acorn` applications [were hitting errors](https://acorn-users.slack.com/archives/C03R9ME0SKC/p1670515378985499?thread_ts=1670508095.720509&cid=C03R9ME0SKC). This would become a non-issue if those images were persistent.

angelbarrera92:
Solves #974 issue by creating a `persistentVolumeClaim` of 10Gi instead of using an `emptyDir`.

I could iterate the pr further if required, something like:
- Using an external container registry.
- Extend the configuration of the internal registry (external object storage).
- make it configurable while installing acorn: `acorn --install --registry-opts....`.

Let me know if this PR is enough, or if it would be better to implement something else. As it is, it covers what is mentioned in the #974 issue.

Thanks!

### DIFF
--- a/pkg/build/buildkit/template.go
+++ b/pkg/build/buildkit/template.go
@@ -11,6 +11,7 @@ import (
 	"github.com/acorn-io/baaah/pkg/apply"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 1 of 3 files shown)


ibuildthecloud:
Oh this will conflict with I just put in. https://github.com/acorn-io/acorn/pull/988

I think the best scalable solution is to use an external registry and not even use the internal.  The referenced PR adds support for an external registry.  By default we will still run a registry in the cluster.  For that registry what I wanted to do was if the cluster has a default storage class, then create a PVC for the registry.  I didn't want builds to completely fail if there is no storage class.

angelbarrera92:
> 

Isn't the default storage class a strict requirement to install acorn? Anyway, I can run a check to decide what kind of volume to use in the registry:

- persistentVolumeClaim if there is a default storage class set.
- emptyDir if there is no storage class marked as default.

WDYT?

PR #1004, opened 2023-01-03
Add CLI tests for install

tylerslaton:
Adding some tests for `acorn install` to help prevent any potential regressions.

### DIFF
--- a/pkg/cli/install_test.go
+++ b/pkg/cli/install_test.go
@@ -1,0 +1,165 @@
+package cli
+
+import (
+	"io"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/acorn-io/acorn/pkg/cli/testdata"
+	"github.com/acorn-io/acorn/pkg/client"
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestInstall(t *testing.T) {
+	type fields struct {
+		Quiet  bool
+		Output string
+		All    bool
+	}
+	type args struct {
+		cmd    *cobra.Command
+		args   []string
+		client *testdata.MockClient
+	}
+	var _, w, _ = os.Pipe()
+	tests := []struct {
+		name           string
+		fields         fields
+		args           args
+		wantErr        bool
+		wantOut        []string
+		commandContext client.CommandContext
+	}{
+		{
+			name: "acorn install: Valid",
+			fields: fields{
+				All:    false,
+				Quiet:  false,
+				Output: "",
+			},
+			commandContext: client.CommandContext{
+				ClientFactory: &testdata.MockClientFactory{},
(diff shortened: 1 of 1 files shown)


pkg/cli/install_test.go
```diff
@@ -0,0 +1,165 @@
...
+		client *testdata.MockClient
+	}
+	var _, w, _ = os.Pipe()
+	tests := []struct {
+		name           string
+		fields         fields
+		args           args
+		wantErr        bool
+		wantOut        []string
+		commandContext client.CommandContext
```
thedadams, line 33: Since this field is the same for each test case, does it make sense to remove the `commandContext` field here and have a single `commandContext` for the whole test?

pkg/cli/install_test.go
```diff
@@ -0,0 +1,165 @@
...
+func TestInstall(t *testing.T) {
+	type fields struct {
+		Quiet  bool
+		Output string
+		All    bool
+	}
+	type args struct {
+		cmd    *cobra.Command
+		args   []string
+		client *testdata.MockClient
```
thedadams, line 24: Same question here: since this is the same for the entire test, can we use just one for the whole test and simplify the "test table"?

pkg/cli/install_test.go
```diff
@@ -0,0 +1,165 @@
...
+)
+
+func TestInstall(t *testing.T) {
+	type fields struct {
+		Quiet  bool
+		Output string
+		All    bool
+	}
+	type args struct {
+		cmd    *cobra.Command
```
thedadams, line 22: It looks like this field is never used. Can it be removed?

This combined with my other comments would also get rid of the `tt.args.args` stutter on line 143.

pkg/cli/install_test.go
```diff
@@ -0,0 +1,165 @@
...
+				out = []byte(err.Error())
+			} else {
+				out, _ = io.ReadAll(r)
+			}
+
+			for _, wantOut := range tt.wantOut {
+				testOut, err := os.ReadFile(wantOut)
+				if err == os.ErrNotExist {
+					testOut = []byte(wantOut)
+				}
```
thedadams, line 160: It seems that we are never expecting the `wantOut` here to be a file. I think it would be good if this was simplified.

thedadams (changes requested):
Just some simplifications that I think would be valuable.

pkg/cli/install_test.go
tylerslaton: Yeah this and all the other comments you hard a definitely valid. The main reason I structured them this was is in the case that we need to edit these values in the future. However, if we do end up needing to edit the values we can update these tests at that time.

tylerslaton:
Closing this PR until I can come back to it with a better implementation.

PR #1012, opened 2023-01-05
fix: refactor NewTag to ParseReference (#1014 #1015)

jsilverio22:
Issue: #1014 #1015
test: add unit tests for duplicate tags and matching images


pkg/client/image.go
```diff
@@ -19,15 +19,16 @@ func (c *client) ImageTag(ctx context.Context, imageName, tag string) error {
...
 	tagResult := &apiv1.ImageTag{}
 	err = c.RESTClient.Post().
 		Namespace(image.Namespace).
 		Resource("images").
 		Name(image.Name).
 		SubResource("tag").
 		Body(&apiv1.ImageTag{
-			Tags: []string{tag},
+			Tags:       []string{tag},
+			ObjectMeta: image.ObjectMeta,
```
ibuildthecloud, line 30: Why are you copying the ObjectMeta? For sure you shouldn't copy the TypeMeta because that will set the TypeMeta from image to imagetag which are different types.

pkg/server/registry/images/tag.go
```diff
@@ -58,9 +59,25 @@ func (t *TagStrategy) ImageTag(ctx context.Context, namespace, imageName string,
...
+		if err != nil {
+			return image, err
+		}
+	}
 
+	image.Tags = append(image.Tags, normalizeTags(tags)...)
+	return image, t.client.Update(ctx, image)
+}
+
+func checkDuplicateTags(image *apiv1.Image, imageList *apiv1.ImageList, tags *[]string) ([]apiv1.Image, error) {
```
ibuildthecloud, line 77: Why is this a `*[]string`?  It's generally bad form to take a pointer to a slice unless you are doing something very specific and tricky.

pkg/server/registry/images/tag.go
```diff
@@ -69,26 +86,48 @@ func (t *TagStrategy) ImageTag(ctx context.Context, namespace, imageName string,
...
+			continue
+		}
+		if imageParsedTag.Name() == tagToFind {
+			*tags = append((*tags)[:i], (*tags)[i+1:]...)
+		}
+	}
+}
+
+func normalizeTags(tags []string) []string {
+	result := []string{}
```
ibuildthecloud, line 124: Small nit, but it's good practice to not allocate empty slices as they just immediately get thrown away once you append to them.  So it's just wasted memory allocation.
```suggestion
	var result []string
```

pkg/server/registry/images/tag.go
ibuildthecloud: Oh, I see on L118.  That's not good, too obscure to modify a slice like that.  What is better is just collect all the tags up front so something like

```golang
set := sets.NewString()
for _, tag := range append(image.Tags, tags...) {
   set.Insert(normalize(tag))
}

// then remove tags from other images that match

image.Tags = set.List()
```

ibuildthecloud (dismissed):

pkg/client/image.go
jsilverio22: I had thought the reasoning for that weird err might have been something with the image version mismatching in object meta but that wasn't the case. removed both

pkg/server/registry/images/tag.go
jsilverio22: Should be easier to follow now

pkg/server/registry/images/tag.go
```diff
@@ -58,37 +59,35 @@ func (t *TagStrategy) ImageTag(ctx context.Context, namespace, imageName string,
...
+	image.Tags = set.List()
 	return image, t.client.Update(ctx, image)
 }
+
+func normalizeTags(tags []string) []string {
+	var result []string
+	for _, tag := range tags {
+		imageParsedTag, err := name.NewTag(tag, name.WithDefaultRegistry(""))
+		if err != nil {
+			return nil
```
cjellick, line 88: you shouldn't just swallow the error here. you should return it and bubble it up. in fact, it would catch the scenario where i put in a bad tag like:
```
acorn tag d53545e106fb 'foo:a@asdfasdfadsf'
```

With your current code, that would just silently drop the tag and return no useful feedback to the user. if you actually return the error, the user will get a useful message

cjellick (changes requested):

pkg/server/registry/images/tag.go
```diff
@@ -58,37 +59,54 @@ func (t *TagStrategy) ImageTag(ctx context.Context, namespace, imageName string,
...
+			return nil, err
 		}
-		for i, tag := range img.Tags {
-			imageParsedTag, err := name.NewTag(tag, name.WithDefaultRegistry(""))
-			if err == nil {
-				tag = imageParsedTag.Name()
-			} else {
-				logrus.Errorf("invalid tag [%s] found, ignoring: %v", tag, err)
+		for i, tag := range res {
+			if img.Name == image.Name {
```
cjellick, line 81: this check should go outside this inner loop. if `img.Name == image.Name`, we should never have to enter this inner for loop. as the logic currently stands, you would still go through the loop the maximum number of times.

pkg/server/registry/images/tag.go
cjellick: can you please add tests cases to assert this fails

cjellick (changes requested):
Three outstanding things:
1. Move the image.name check outside the inner loop - https://github.com/acorn-io/acorn/pull/1012/files#r1064030607
2. Add tests for bad tag input - https://github.com/acorn-io/acorn/pull/1012#discussion_r1064031036
3. Change the ImagTag function to interact with ImageInstance instead of Image. This one is a non-blocking issue. If you don't understand fully, let me know and we can skip it for now and address later in the week

cjellick (approved):

cjellick (approved):

PR #1013, opened 2023-01-05
Add ability to test external API Server and Registry

tylerslaton:
Adding some updates to get the nightly EKS tests passing.

1. Update API server to only get created when TEST_ACORN_API_SERVER is set to external
2. Add a local-path StorageClass so that tests requiring volumes can provision correctly
3. Update the registry helper to create one that is accessible to the cluster when testing external API servers.

Signed-off-by: tylerslaton <mtslaton1@gmail.com>

### DIFF
--- a/integration/helper/controller.go
+++ b/integration/helper/controller.go
@@ -2,6 +2,8 @@ package helper
 
 import (
 	"context"
+	"errors"
+	"fmt"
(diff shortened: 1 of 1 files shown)


tylerslaton:
Closing for now due to inactivity.

Issue #1018, opened 2023-01-09
Report successful EKS nightly test runs to Slack


tylerslaton:
Right now we're reporting failures. We would like to also report success to a separate channel (for reduced noise) to confirm the tests are still running and passing.

cjellick:
Working 🙌

PR #1019, opened 2023-01-09
Report successful runs of nightly EKS tests to slack

tylerslaton:
#1018 

Once this PR merges, I will update the repo secrets to include the success and failure channel ids.

### DIFF
--- a/.github/workflows/aws-eks-test.yml
+++ b/.github/workflows/aws-eks-test.yml
@@ -56,11 +56,20 @@ jobs:
 
       - name: report failure to slack
         if: failure()
-        id: slack
+        id: slack-failure
         uses: slackapi/slack-github-action@v1.23.0
         with:
-          channel-id: '${{ secrets.SLACK_BOT_CHANNEL }}'
-          slack-message: "Nightly EKS test failed: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+          channel-id: '${{ secrets.SLACK_BOT_FAILURE_CHANNEL }}'
+          slack-message: "❌ Nightly EKS test failed: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/aws-eks-test.yml
```diff
@@ -56,11 +56,20 @@ jobs:
...
         with:
-          channel-id: '${{ secrets.SLACK_BOT_CHANNEL }}'
-          slack-message: "Nightly EKS test failed: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+          channel-id: '${{ secrets.SLACK_BOT_FAILURE_CHANNEL }}'
+          slack-message: "❌ Nightly EKS test failed: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+        env:
+          SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
+
+       - name: report success to slack
+        id: slack-success
```
JeffNeff, line 68:
```suggestion
        if: success()
        id: slack-success
```
Will this not produce false positive "success" messages without checking the status?

JeffNeff (commented):
<3

.github/workflows/aws-eks-test.yml
tylerslaton: Great question! [GitHub defaults to having this behavior unless otherwise specified](https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions). So, if a previous job fails then it will not run the next one. Thus, if a step (our test step in this case) fails then subsequent steps will not run which is why we need `if: failure()` and `if: always()`. In the case that it passes, this step will to run due to its default `if: success()` but will not if a failure occurs in a previous step.

thedadams (approved):

cjellick (approved):

.github/workflows/aws-eks-test.yml
JeffNeff: Thank you for that information!

Issue #43, opened 2022-05-19
Add resource request/constraint functionality (...for memory)


cjellick:
Add the ability to constraint the pod resources (essentially set pod resource requests and limits).

Update 1/19/23 - we are going to use this issue to just track memory, no other resources like cpu

cjellick:
good use case: prove that we can properly launch and manage AI/ML applications that require GPU support

sirredbeard:
> good use case: prove that we can properly launch and manage AI/ML applications that require GPU support

Like Determined

cjellick:
@sangee2004 the best way to understand the scope of this one is to refer to the docs @tylerslaton added. they do a good job encapsulating the functionality

- https://docs.acorn.io/installation/options#memory
- https://docs.acorn.io/authoring/containers (actually at the time of writing this, theres a separate memory page in the authoring section, but i want @tylerslaton  to move it to be in the containers page)
- https://docs.acorn.io/running/memory

sangee2004:
Tested with acorn version v0.5.0-rc3-18-gfbe46a2+fbe46a25

This feature works as described in the docs. Issues found when testing this feature is tracked as separate issues.  Related blocking issue is tracked in https://github.com/acorn-io/acorn/issues/1089.

sirredbeard:
Should I open another issue to track resources like CPU and GPU?

PR #1028, opened 2023-01-11
Add ability to set workload memory constraints via limit/request (#43)

tylerslaton:
## Summary
This commit adds three ways to set the limit/request restriction on memory. In order or precedence, they are:

1. "acorn run --memory service=512M" where service is the name of a service in the Acornfile and 512M is 512 megabytes.
2. A new property in the Acornfile for defined containers - memory. It takes a Cuelang integer that equates to bytes.
3. "acorn install --default-service-memory 512M" where 512M is the memory that Acorn will default to if no other option is set.

Co-authored-by: Joshua Silverio <joshua@acorn.io>

## In-progress
- [x] Working on getting updates to work as expected with the new `--memory` flag.
- [x] Update approach for `acorn install -m`
- [x] Address test failures


docs/docs/100-reference/01-command-line/acorn_install.md
```diff
@@ -27,6 +27,8 @@ acorn install
       --cluster-domain strings                The externally addressable cluster domain (default .on-acorn.io)
       --controller-replicas int               acorn-controller deployment replica count
       --default-publish-mode string           If no publish mode is set default to this value (default user)
+      --default-service-memory int
```
tylerslaton, line 30: We're going to be fixing this as it is not currently the desired behavior. There will only be one flag here once this is fixed.

pkg/cli/install.go
```diff
@@ -40,6 +46,14 @@ func (i *Install) Run(cmd *cobra.Command, args []string) error {
...
 	}
 
+	i.DefaultServiceMemory = DEFAULT_SERVICE_MEMORY
+	if i.DefaultServiceMemoryString != "" {
+		quantity, err := resource.ParseQuantity(i.DefaultServiceMemoryString)
+		if err != nil {
+			return err
+		}
+		i.DefaultServiceMemory = quantity.Value()
+	}
```
tylerslaton, line 56: Trying to brainstorm through ways to improve this to not be as hacky. We like the UX of `acorn install -m 512M` vs `acorn install -m 536,870,912` so a string feels natural but it causes a couple of issues (namely two flags being present).

docs/docs/100-reference/01-command-line/acorn_install.md
cjellick, line 30: service isnt the word we use. we use container.

docs/docs/100-reference/01-command-line/acorn_run.md
```diff
@@ -87,6 +87,7 @@ acorn run [flags] IMAGE|DIRECTORY [acorn args]
       --interval string           If configured for auto-upgrade, this is the time interval at which to check for new releases (ex: 1h, 5m)
   -l, --label strings             Add labels to the app and the resources it creates (format [type:][name:]key=value) (ex k=v, containers:k=v)
       --link strings              Link external app as a service in the current app (format app-name:container-name)
+  -m, --memory strings            Overwrite memory limit for a container (format service=memory) (ex foo=512M)
```
cjellick, line 90: dont like the word "overwrite" here. Just use "set"

pkg/apis/internal.acorn.io/v1/appinstance.go
```diff
@@ -55,24 +56,25 @@ const (
...
+	PublishMode         PublishMode                            `json:"publishMode,omitempty"`
+	TargetNamespace     string                                 `json:"targetNamespace,omitempty"`
+	Links               []ServiceBinding                       `json:"services,omitempty"`
+	Ports               []PortBinding                          `json:"ports,omitempty"`
+	DeployArgs          GenericMap                             `json:"deployArgs,omitempty"`
+	Permissions         []Permissions                          `json:"permissions,omitempty"`
+	AutoUpgrade         *bool                                  `json:"autoUpgrade,omitempty"`
+	NotifyUpgrade       *bool                                  `json:"notifyUpgrade,omitempty"`
+	AutoUpgradeInterval string                                 `json:"autoUpgradeInterval,omitempty"`
+	Memory              map[string]corev1.ResourceRequirements `json:"memory,omitempty"`
```
cjellick, line 77: In our design discussion, we said that this should just be a simple int, not a the core type.

So, the conversion to a resource requirement shouldnt happen until you're in the controller and translating AppInstance to a deployment.

pkg/cli/install.go
cjellick: This is because we want the API to be an int, but the flag to be a string, right?

tylerslaton: Yes, exactly. Its cumbersome to type out all 9 digits for a megabyte and go doesn't support the suffixes [natively like Cue does](https://cuelang.org/docs/tutorials/tour/types/numbers/).

docs/docs/100-reference/01-command-line/acorn_install.md
cjellick: Hm. good point. i still dont love using that word. i'd say `--default-memory` rather than service. Workload is another possibility. let's touch on in standup.

tylerslaton: For those coming to read this later, we decided to reference the Acorn concept of Jobs and Containers as workloads.

tylerslaton: The fix for this is now out. Our builder library is now able to parse int64 types in the config from a string flag. This is enabled by adding the `quantity="true"` reflection tag to a int64 type.

The main thing that needs to be changed is to update the doc generation library to write this flag as a accepting a string.

pkg/cli/install.go
tylerslaton: I think we reached a good middle ground here with the latest commits.

pkg/apis/api.acorn.io/v1/types.go
```diff
@@ -325,6 +325,7 @@ type Config struct {
 	PublishBuilders              *bool          `json:"publishBuilders" name:"publish-builders" usage:"Publish the builders through ingress to so build traffic does not traverse the api-server"`
 	BuilderPerNamespace          *bool          `json:"builderPerNamespace" name:"builder-per-namespace" usage:"Create a dedicated builder per namespace"`
 	InternalRegistryPrefix       string         `json:"internalRegistryPrefix" name:"internal-registry-prefix" usage:"The image prefix to use when pushing internal images (example ghcr.io/my-org/)"`
+	DefaultWorkloadMemory        int64          `json:"defaultWorkloadMemory" name:"default-workload-memory" quantity:"true" usage:"Set the memory limit default for installed Acorns." short:"m" default:"256M"`
```
cjellick, line 328:
```suggestion
	DefaultWorkloadMemory        int64          `json:"defaultWorkloadMemory" name:"default-workload-memory" quantity:"true" usage:"Set the memory limit default for acorn workloads." short:"m" default:"256M"`
```

Suggesting this change in an attempt to clarify that this isnt the quantity for the entire acorn, but for each container in it.

cjellick: also, drop the word limit as it implies just the k8s "limit", but this is for limit and request. Basically, don tuse the limit qualifier

pkg/apis/internal.acorn.io/v1/memory.go
```diff
@@ -0,0 +1,33 @@
+package v1
+
+import (
+	"cuelang.org/go/cue/literal"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+func ParseMemoryMap(memory map[string]string) (map[string]int64, error) {
+	resources := map[string]int64{}
+	for service, value := range memory {
```
cjellick, line 10: even though its not user facing, id prefer you not use the term service. i just want to avoid the conflation

jsilverio22: Workload in place of service here as well?

tylerslaton: This isn't necessarily tied to a workload anymore since anything can call it to create the memory map. Updated it to just be `key`.

pkg/apis/internal.acorn.io/v1/memory.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+func ParseQuantityString(memory string) (int64, error) {
+	numInfo := literal.NumInfo{}
+	err := literal.ParseNum(memory, &numInfo)
+	if err != nil {
+		return -1, err
+	}
+
+	quantity, err := resource.ParseQuantity(numInfo.String())
```
thedadams, line 27: Cue actually uses `"github.com/cockroachdb/apd/v2.Decimal"` under the hood for its parsing. My concern with using `Quantity` here is that there may be a difference with the way the two parse the `numInfo.String()` value.

Would there be a problem with using `numInfo.Decimal` here instead?

pkg/cli/builder/builder.go
```diff
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"unsafe"
 
+	v1 "github.com/acorn-io/acorn/pkg/apis/internal.acorn.io/v1"
```
thedadams, line 11: nit: 
```suggestion
	"github.com/acorn-io/acorn/pkg/apis/internal.acorn.io/v1"
```

schema/v1/app.cue
```diff
@@ -92,6 +92,7 @@ package v1
 	ports:                          #PortSingle | *[...#Port] | #PortMap
 	[=~"probes|probe"]:             #Probes
 	[=~"depends[oO]n|depends_on"]:  string | *[...string]
+	memory?: 												int
```
ibuildthecloud, line 95: support mem here too.

pkg/controller/appdefinition/testdata/volumes/contextdir/expected.yaml
```diff
@@ -45,6 +45,11 @@ spec:
       serviceAccountName: container-name
       containers:
         - name: container-name
+          resources:
```
ibuildthecloud, line 48: If there's no global default and the user hasn't set a memory value there should be no change to the output here.  There should be no "resources" field at all.

pkg/cli/builder/builder.go
```diff
@@ -207,8 +216,35 @@ func assignOptBool(app *cobra.Command, maps map[string]reflect.Value) error {
...
+		}
+
+		v.Set(reflect.ValueOf(quantity))
+	}
+	return nil
+}
+
 func assignOptString(app *cobra.Command, maps map[string]reflect.Value) error {
 	for k, v := range maps {
+
```
thedadams, line 247: nit: remove line

pkg/cli/install_test.go
```diff
@@ -0,0 +1,71 @@
+package cli
```
thedadams, line 1: These changes seem unrelated to memory requests and limits.

pkg/cli/run_test.go
```diff
@@ -22,3 +27,100 @@ func TestRunArgs_Env(t *testing.T) {
...
+			args: args{
+				args:   []string{"-m found.container=notallowed", "found"},
+				client: &testdata.MockClient{},
+			},
+			wantErr: true,
+			wantOut: "illegal number start \"notallowed\"",
+		},
+	}
+	for _, tt := range tests {
+		r, w, _ := os.Pipe()
```
ibuildthecloud, line 110: The for body should be wrapped with the below snippet so that tt.name is passed to the go test runner and the test name shows up.  Additionally tools will allow you to run just that one subtest easier (goland does this).
```
t.Run(tt.name, func(t *testing.T) {
})
```

pkg/cli/install.go
```diff
@@ -8,6 +8,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const (
+	DEFAULT_WORKLOAD_MEMORY = int64(536870912) // 512M
```
ibuildthecloud, line 12: Don't use ALL_CAPS syntax.  Go style prefers `UpperCamelCase`  Also if it's not public use normal `camelCase`.  In short constants don't have a special case style.

pkg/client/app.go
```diff
@@ -137,6 +138,9 @@ func ToAppUpdate(ctx context.Context, c Client, name string, opts *AppUpdateOpti
 	if opts.AutoUpgradeInterval != "" {
 		app.Spec.AutoUpgradeInterval = opts.AutoUpgradeInterval
 	}
+	if len(opts.Memory) != 0 {
```
thedadams, line 141: nit: It seems that always setting this would be harmless. Is that the case?

pkg/controller/appdefinition/deploy.go
```diff
@@ -406,9 +407,32 @@ func toContainer(app *v1.AppInstance, tag name.Reference, deploymentName, contai
 		LivenessProbe:  toProbe(container, v1.LivenessProbeType),
 		StartupProbe:   toProbe(container, v1.StartupProbeType),
 		ReadinessProbe: toProbe(container, v1.ReadinessProbeType),
+		Resources:      toResources(req, app, container, containerName),
 	}
 }
 
+func toResources(req router.Request, app *v1.AppInstance, container v1.Container, containerName string) corev1.ResourceRequirements {
+	var bytes int64
```
thedadams, line 415: nit: `bytes` here may be confusing because of the `byte` type in Go. Suggest using `memBytes` instead.

pkg/controller/appdefinition/testdata/cronjob/expected.yaml
```diff
@@ -47,6 +47,11 @@ spec:
           terminationGracePeriodSeconds: 5
           containers:
             - name: oneimage
+              resources:
```
thedadams, line 50: It seems strange to me that all of these tests would produce memory limits and requests of `"0"`. Can we leave out the memory requests and limits if they will output `"0"`?

thedadams (changes requested):
Curious to know why the memory requests and limits are getting stored on the Acorn objects as `int64`s. The memory requests/limits are taken in as `string`s, converted to `int64`s to be stored on the Acorn objects, then converted to `Quantities` to be put on the Kubernetes deployment. Could we take in the Cue `string`s and convert them directly to `Quantities` on the Acorn objects? The idea here is that they would be stored on the Acorn objects in human-readable form (like `12M`) instead of large numbers.

Secondly, we (the whole team) should have a discussion about setting memory requests and limit to the same value.

pkg/controller/appdefinition/deploy.go
```diff
@@ -406,9 +407,32 @@ func toContainer(app *v1.AppInstance, tag name.Reference, deploymentName, contai
...
+		bytes = container.Memory
+	} else {
+		//Attempt to grab default from the apiv1.Config
+		cfg, err := config.Get(req.Ctx, req.Client)
+		if err == nil {
+			bytes = cfg.DefaultWorkloadMemory
+		}
+	}
+
+	quantity := resource.NewQuantity(bytes, resource.DecimalSI)
```
thedadams, line 430: Note from huddle: This should be BinarySI

pkg/apis/internal.acorn.io/v1/memory.go
thedadams: After huddle conversation: use `strconv.ParseInt` here to get an `int64`.

pkg/cli/builder/builder.go
tylerslaton: Go format keeps bringing this `v1` back when I go to remove it so I believe that it is here to stay unless I'm missing something.

pkg/client/app.go
tylerslaton: By removing the if statement here, it makes it so that if a user does not pass anything for `-m` when updating an Acorn it will overwrite the existing setting to be nothing. So this line (and the if statements before) are allowing a user to only set the options that they want to update.

tylerslaton: I do think that there is an update to be made here though as there is a situation where a user wants to remove all `appInstance.Spec` memory. I'll think through something to solve that issue.

pkg/cli/builder/builder.go
thedadams: Ah, yep. I forgot that Go treats `v*` packages as versioned packages.

pkg/client/app.go
tylerslaton: Okay so with a recent change that @jsilverio22 did we are now able to unset all memory for a Acorn by running `acorn update -m 0 foo`. This will make it so whatever memory the `foo` acorn had set will be overwritten so this code should be fine as is.

pkg/cli/install_test.go
tylerslaton: Agreed. Removed it from this PR and will address it later in this one:

https://github.com/acorn-io/acorn/pull/1004

schema/v1/app.cue
```diff
@@ -92,6 +92,7 @@ package v1
 	ports:                          #PortSingle | *[...#Port] | #PortMap
 	[=~"probes|probe"]:             #Probes
 	[=~"depends[oO]n|depends_on"]:  string | *[...string]
+	[=~"mem|memory"]: 							int
```
cjellick, line 95: looks like everyhting else in this file is separated by spaces, not tabs. can you change this to match so it aligns properly in diffs?

pkg/apis/api.acorn.io/v1/types.go
```diff
@@ -332,6 +332,8 @@ type Config struct {
 	PublishBuilders              *bool          `json:"publishBuilders" name:"publish-builders" usage:"Publish the builders through ingress to so build traffic does not traverse the api-server"`
 	BuilderPerProject            *bool          `json:"builderPerProject" name:"builder-per-project" usage:"Create a dedicated builder per project"`
 	InternalRegistryPrefix       *string        `json:"internalRegistryPrefix" name:"internal-registry-prefix" usage:"The image prefix to use when pushing internal images (example ghcr.io/my-org/)"`
+	DefaultWorkloadMemory        int64          `json:"defaultWorkloadMemory" name:"default-workload-memory" quantity:"true" usage:"Set the default memory for acorn workloads." short:"m" default:"256Mi"`
```
cjellick, line 335: How about `WorkloadMemoryDefault` to achieve symmetry and help-text nearness (since its sorted alphabetically) to `WorkloadMemoryMaxiumum`?

pkg/apis/internal.acorn.io/v1/memory.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"cuelang.org/go/cue/literal"
+)
+
+func ParseMemoryMap(memory map[string]string) (Memory, error) {
+	resources := Memory{
+		All:       0,
+		Workloads: map[string]int64{},
+	}
+	allSet := false
```
cjellick, line 14: Can you explain whats going on with this allSet logic? I'm confused

pkg/apis/api.acorn.io/v1/types.go
tylerslaton: Sounds good to me!

pkg/apis/internal.acorn.io/v1/memory.go
tylerslaton: This is how we allow a user to do `acorn run -m 512Mi` and `acorn run -m nginx:512Mi` with the same flag.

pkg/controller/appdefinition/deploy.go
```diff
@@ -196,17 +198,26 @@ func toContainers(app *v1.AppInstance, tag name.Reference, name string, containe
...
-	containers = append(containers, toContainer(app, tag, name, name, container))
+	newContainer, err := toContainer(req, app, tag, name, name, container)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	containers = append(containers, newContainer)
 	for _, entry := range typed.Sorted(container.Sidecars) {
-		newContainer := toContainer(app, tag, name, entry.Key, entry.Value)
+		newContainer, err := toContainer(req, app, tag, name, entry.Key, entry.Value)
```
thedadams, line 208: nit: Don't reallocate on every loop iteration if you already have the necessary.
```suggestion
		newContainer, err = toContainer(req, app, tag, name, entry.Key, entry.Value)
```

pkg/controller/appdefinition/deploy.go
```diff
@@ -407,6 +418,65 @@ func toContainer(app *v1.AppInstance, tag name.Reference, deploymentName, contai
...
+
+	// For bytes, 0 is equivalent to "unlimited"
+	if maximum != 0 {
+		if memBytes > maximum {
+			return nil, fmt.Errorf(
+				"workload %v exceeds the workload-memory-maximum by %v",
+				containerName,
+				resource.NewQuantity(memBytes-maximum, resource.BinarySI).String())
+		} else if memBytes == 0 {
+			return nil, fmt.Errorf(
```
thedadams, line 464: What are your thoughts on just setting `memBytes = maximum` in this case?

thedadams (commented):
I know there are other changes coming here, but just some thoughts on this review round.

I think this is shaping up nicely!

pkg/cli/builder/builder.go
```diff
@@ -147,6 +149,13 @@ func Command(obj Runnable, cmd cobra.Command) *cobra.Command {
 			case reflect.Int:
 				optInt[name] = v
 				flags.IntP(name, alias, defInt, usage)
+			case reflect.Int64:
+				// In the case that a quantity tag is found and set to true, we want to create a string flag
+				// for it that will get parsed into an *int64. See assignQuantities().
+				if fieldType.Tag.Get("quantity") == "true" {
```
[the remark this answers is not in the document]
tylerslaton: Yeah I was debating that when adding this code. I've seen reviewers go in different directions so I left it out. That being said I'm on-board with including it as the change scope is minimal and makes our lives easier in the future that we need an `*int64` flag.

pkg/client/app.go
```diff
@@ -137,6 +138,14 @@ func ToAppUpdate(ctx context.Context, c Client, name string, opts *AppUpdateOpti
 	if opts.AutoUpgradeInterval != "" {
 		app.Spec.AutoUpgradeInterval = opts.AutoUpgradeInterval
 	}
+	if len(opts.Memory.Workloads) != 0 {
+		app.Spec.Memory.Workloads = opts.Memory.Workloads
+	}
+	if opts.Memory.All != nil {
+		app.Spec.Memory.All = opts.Memory.All
+		// TODO: needed?
+		app.Spec.Memory.Workloads = opts.Memory.Workloads
```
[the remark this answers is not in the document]
tylerslaton: That's how I was reading it too, completely forgot that I had a TODO comment and it got committed here. I ended up overhauling how we set memory on the `appSpec` in general so that's definitely worth another look if you've got the time.

pkg/apis/internal.acorn.io/v1/memory.go
```diff
@@ -0,0 +1,47 @@
...
+		}
+		if allSet {
+			resources.All = &quantity
+			allSet = false
+			continue
+		}
+		resources.Workloads[key] = &quantity
+	}
+	return resources, nil
+}
```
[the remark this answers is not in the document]
tylerslaton: This function got overhauled heavily as apart of the new changes. Mind giving that a look and seeing if you agree with the implementation there?

docs/docs/50-running/55-memory.md
```diff
@@ -0,0 +1,36 @@
+---
+title: Memory
+---
+You can configure Acorn apps to have a set memory upon startup.
+
+There are three primary places where a memory default can be set in order of precedence,
+
+1. Through an `acorn run` flag `--memory workloadName=512M` where workloadName is the name of a workload in the Acornfile and 512M is 512 megabytes. (if no workloadName is specified i.e. `--memory 1G` will apply 1 Gigabyte of memory to all workloads defined in the Acornfile)
```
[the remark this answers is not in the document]
tylerslaton: Fixed!

pkg/controller/appdefinition/deploy.go
tylerslaton: I see the merit in this since it helps the user get to their (probably) desired state without having to change anything. 

My main concern - what happens if the user is expecting to never run into memory issues by setting their limit to 0 and suddenly their app starts getting `OOM`'d due to the maximum be set as their value? This current approach prevents them from ever getting to that state since their desired state cannot be guaranteed.

pkg/apis/internal.acorn.io/v1/memory.go
tylerslaton: This function got heavily overhauled. Mind giving it another look and seeing if its still confusing? The approach is all new.

thedadams (commented):
Spoke about these items "offline"

docs/docs/100-reference/01-command-line/acorn_install.md
```diff
@@ -42,6 +42,8 @@ acorn install
       --record-builds                         Keep a record of each acorn build that happens
       --set-pod-security-enforce-profile      Set the PodSecurity profile on created namespaces (default true)
       --skip-checks                           Bypass installation checks
+  -m, --workload-memory-default string        Set the default memory for acorn workloads. Expects binary suffixes, such as Ki, Mi, Gi, Ti, Pi (default "0")
```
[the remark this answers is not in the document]
jsilverio22: I think this is described in more detail on the docs side still worth it to add here?

pkg/cli/run.go
```diff
@@ -127,6 +127,7 @@ type RunArgs struct {
 	NotifyUpgrade   *bool    `usage:"If true and the app is configured for auto-upgrades, you will be notified in the CLI when an upgrade is available and must confirm it"`
 	AutoUpgrade     *bool    `usage:"Enabled automatic upgrades."`
 	Interval        string   `usage:"If configured for auto-upgrade, this is the time interval at which to check for new releases (ex: 1h, 5m)"`
+	Memory          []string `usage:"Set memory for a workload in the format of workload=memory. Only specify an amount to set all workloads. (ex foo=512M or 512M)" short:"m"`
```
jsilverio22, line 130: same nit as Donnie from above Mi vs M

jsilverio22 (approved):

pkg/controller/appdefinition/deploy.go
tylerslaton: Resolved in offline conversation. We're going with your original approach and will update the docs to reflect the phrasing of "Setting memory to 0 will give that Acorn the maximum memory it is allowed to take".

pkg/apis/internal.acorn.io/v1/memory.go
```diff
@@ -0,0 +1,122 @@
...
+
+var (
+	ErrInvalidAcornMemory   = errors.New("invalid memory from acorn")
+	ErrInvalidSetMemory     = errors.New("invalid memory set by user")
+	ErrInvalidDefaultMemory = errors.New("invalid memory default")
+)
+
+type MemoryError struct {
+	Err      error
```
jsilverio22, line 58: https://github.com/acorn-io/acorn/pull/1028/files#annotation_8115132309 saying there is an invalid field type here

pkg/apis/internal.acorn.io/v1/memory.go
```diff
@@ -0,0 +1,122 @@
...
+	quantity, err := strconv.ParseInt(numInfo.String(), 10, 64)
+	if err != nil {
+		return -1, err
+	}
+
+	return quantity, nil
+}
+
+var (
+	ErrInvalidAcornMemory   = errors.New("invalid memory from acorn")
```
jsilverio22, line 52:
```suggestion
	ErrInvalidAcornMemory   = errors.New("invalid memory from Acornfile")
```

jsilverio22: acorn might be hard to determine where invalidity is coming from

pkg/apis/internal.acorn.io/v1/memory.go
```diff
@@ -0,0 +1,122 @@
...
+		err = ErrInvalidAcornMemory
+		memBytes = *container.Memory
+	}
+
+	// For maximum memory, 0 is equivalent to "unlimited"
+	if memMaximum != 0 {
+		var (
+			maxQuantity     = resource.NewQuantity(memMaximum, resource.BinarySI).String()
+			defaultQuantity = resource.NewQuantity(memBytes, resource.BinarySI).String()
+			bytesQuantity   = resource.NewQuantity(memBytes, resource.BinarySI).String()
```
jsilverio22, line 91: small: looks like defaultQuantity and bytesQuantity are being set to same value here maybe eliminate one call to NewQuantity()

jsilverio22 (commented):
nice job adding all the changes 👍

cjellick:
https://github.com/acorn-io/acorn/issues/43

pkg/apis/internal.acorn.io/v1/memory.go
tylerslaton: Yeah I added a comment for deep copy gen to ignore this but I guess it didn't end up getting committed. Should be fixed now 👍🏻

pkg/apis/internal.acorn.io/v1/memory.go
```diff
@@ -0,0 +1,123 @@
...
+				Err: fmt.Errorf(
+					"%w: workload %v with memory of %v exceeds the workload-memory-maximum of %v",
+					err, containerName, bytesQuantity, maxQuantity),
+				Quantity: bytesQuantity}
+		} else if memBytes == 0 {
+			// For bytes, 0 is viewed as the maximum allowed memory". As such,
+			// update to the current maximum.
+			memBytes = memMaximum
+		}
+	}
```
thedadams, line 113: I think this logic belongs in `validator.go`. The function should be responsible for parsing the memory information, but let the validator validate it.

Once the validator does its job, then any subsequent call to this method for parsing can safely assume the values meet the expected criteria, or that the app was in compliance when it was deployed and should continue to function properly.

Additionally, this would allow the app to continue to function as expected if the default memory for a cluster is updated and causes a workflow to be out of compliance. The offending app would only be flagged as invalid on update, which I think is the desired outcome.

Lastly, putting this code in the validator will also get rid of the need for the `MemoryError` type.

pkg/apis/internal.acorn.io/v1/memory.go
```diff
@@ -0,0 +1,123 @@
...
+	return quantity, nil
+}
+
+var (
+	ErrInvalidAcornMemory   = errors.New("invalid memory from Acornfile")
+	ErrInvalidSetMemory     = errors.New("invalid memory set by user")
+	ErrInvalidDefaultMemory = errors.New("invalid memory default")
+)
+
+// +k8s:deepcopy-gen=false
```
thedadams, line 57: I was iffy on this type before you added this comment. Now I think it would be valuable to figure out a way to not need this.

thedadams (changes requested):

pkg/server/registry/apps/validator.go
```diff
@@ -298,6 +329,34 @@ func (s *Validator) getPermissions(ctx context.Context, image string, app *apiv1
 	return result, nil
 }
 
+func (s *Validator) getWorkloads(ctx context.Context, image string, app *apiv1.App) (map[string]v1.Container, error) {
+	result := map[string]v1.Container{}
```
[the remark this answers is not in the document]
tylerslaton: That's valid, what's the reasoning behind this best practice just so I know for the future?

pkg/server/registry/apps/validator.go
```diff
@@ -298,6 +329,34 @@ func (s *Validator) getPermissions(ctx context.Context, image string, app *apiv1
...
+	details, err := s.clientFactory.Namespace("", app.Namespace).ImageDetails(ctx, image,
+		&client.ImageDetailsOptions{
+			Profiles:   app.Spec.Profiles,
+			DeployArgs: app.Spec.DeployArgs})
+
+	if err != nil {
+		return result, err
```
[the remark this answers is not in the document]
tylerslaton: Yep! Looks like a typo snuck in there.

pkg/controller/appdefinition/deploy.go
```diff
@@ -425,6 +434,27 @@ func toContainer(app *v1.AppInstance, tag name.Reference, deploymentName, contai
...
+
+func toResources(req router.Request, app *v1.AppInstance, container v1.Container, containerName string) (*corev1.ResourceRequirements, error) {
+	cfg, err := config.Get(req.Ctx, req.Client)
+	if err != nil {
+		return nil, err
+	}
+
+	resources, memErr := v1.MemoryToRequirements(app.Spec.Memory, containerName, container, cfg.WorkloadMemoryDefault, cfg.WorkloadMemoryMaximum)
+	if memErr.Err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
tylerslaton: Yep!

pkg/server/registry/apps/validator.go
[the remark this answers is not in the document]
thedadams: In Go, slices and maps use an array under the hood. When `append` or map assignment determines this array needs to grow, a new array array with twice the size is allocated and the data copied over. This can be expensive if done over and over, and you can save quite a bit of time/compute by allocating as much as possible upfront.

pkg/apis/internal.acorn.io/v1/memory.go
tylerslaton: Yeah I don't love this code. The only reason this struct exists is to pass off the invalid value so that API server's message is a bit friendlier.

pkg/apis/internal.acorn.io/v1/memory.go
tylerslaton: Talked offline but the most recent push fixes this to be in the vision of what you were mentioning.

pkg/server/registry/apps/validator.go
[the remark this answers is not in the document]
tylerslaton: That makes a ton of sense. Thanks for the explanation! I'll remember this for the future.

I went ahead and fixed this in the most recent push.

pkg/apis/internal.acorn.io/v1/memory.go
tylerslaton: I'm happy to report that the `MemoryError` struct is gone!

thedadams (approved):
Looks good! Great job dealing with all the discussions from various folks!

docs/docs/50-running/55-memory.md
```diff
@@ -0,0 +1,103 @@
...
+        ports: publish: "80/http"
+        files: {
+            "/usr/share/nginx/html/index.html": "<h1>My first Acorn!</h1>"
+        }
+        memory: 512Mi
+    }
+}
+```
+
+> **Note**: This can be overwritten by `acorn run --memory`.
```
cjellick, line 67: use docusaurus admonitions https://docusaurus.io/docs/markdown-features/admonitions
```suggestion
:::note

This can be overwritten by `acorn run --memory`.

:::
```

docs/docs/50-running/55-memory.md
```diff
@@ -0,0 +1,103 @@
...
+acorn install --workload-memory-maximum 1Gi`
+```
+
+This will set it so all Acorns on this cluster will be unable to install should they exceed `1Gi` of memory.
+
+## No limit
+It is possible to set any of these flags to have no limit on memory by simply setting their value to `0`. However, running an Acorn with its memory set to `0` while the `--workload-memory-maximum` is not set to `0` will roughly translate to "use as much memory as allowed". For example, say that we install Acorn with a non-zero `--workload-memory-maximum`.
+
+```shell
+acorn run --workload-memory-maximum 512Mi
```
cjellick, line 94:
```suggestion
acorn install --workload-memory-maximum 512Mi
```

docs/docs/50-running/55-memory.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+Then we try to run an Acorn with its memory set to 0.
+
+```shell
+acorn run --memory 0 foo
+```
+
+When the `foo` Acorn gets provisioned, all of its containers will have their memory set to `512Mi` (the `--workload-memory-maximum` we set prior).
```
cjellick, line 103: if i omit  a memory setting completely in this case, will it get the max or be unset?

cjellick (changes requested):

docs/docs/50-running/55-memory.md
tylerslaton: In this case it will get the max.

tylerslaton: Misread the question - it will get whatever the default is here.

cjellick (approved):

Issue #1023, opened 2023-01-11
Trying to specify a container for `acorn logs` or `acorn exec` is confusing


cjellick:
When using `acorn logs` or `acorn exec`, it is unclear how to specify the container name.

Like, to get logs for an app named happy-tree, you'd do `acorn logs happy-tree`. If you want to get the logs for a container named `web` within happy-tree's acorn image, you'd think the command would be something like `acorn logs happy-tree.web`, but that's not how it works. You need to get the container name from the output of `acorn containers` and the command is going to look like:

```
acorn logs happy-tree.my-app-7657d9bbbf-hmhz6
```

I think we just need some tweaks to the help commands of logs and exec to explain container names can be obtained from the output the containers command. It would also be nice to have a few examples, like we do in the help text of the `acorn run` command.

sangee2004:
Tested with build from master - jan/16 -  acorn version v0.4.1-112-gb7531fe+b7531fe7

Auto complete is available for containers.

So we should be able to use auto complete feature for fetching the desired containers in `acron logs` and `acorn exec` commands.

Autocomplete  defaults to apps , so if the intent was for containers we need to backtrack one space and then enter "." to get to the list of containers with in the app.

Issue #841, opened 2022-11-03
Add autocomplete on resource names


ibuildthecloud:
Right now we have just basic CLI bash completion which auto-completes command names.  I would like something like `acorn exec <TAB>` to print the list of apps/containers. Each auto complete would have to be added to each command specific to the commands behavior.

cjellick:
Marking this is "High" because I think it would have a big positive user impact

thedadams:
To setup the completions on `bash`:
1. Run `brew install bash-completion@2`
2. Add `[[ -r "/opt/homebrew/etc/profile.d/bash_completion.sh" ]] && . "/opt/homebrew/etc/profile.d/bash_completion.sh"` to your `~/.bash_profile` and restart your terminal
3. Run `source <(acorn completion bash)`

Completion should now work, and dynamic completion should work if you have Acorns installed.

sangee2004:
Tested with master build - acorn version v0.4.1-112-gb7531fe+b7531fe7

Auto completes works  for container , secrets , volumes ,images  and credentials.

PR #1031, opened 2023-01-13
Add dynamic CLI completion

thedadams:
These changes add dynamic completion to the CLI. The majority of the changes are in the first commit. Each subsequent commit simply adds to the changes for the new types (with a few caveats).

There are a couple commands (`logs` and `exec`) that take either an app name or a container name. The completion will suggest app names unless there is a '.' in the name to be completed. In this case, it can be determined that the user is looking for a container name and the completion will adjust accordingly.

For the logs command specifically, the user can use the `-c` flag to specify a container name from the Acornfile (this is also dynamically completed). If the user has specified such a container, then the completion will only suggest apps with containers of the given name.

Anywhere there could be an image completion, the completion will always prefer tags to digests. That is, if an image has a tag, then the completion will not suggest the digest for that image. The exception to this is if the user is trying to get completions for something that matches the digest but doesn't match the tag. In this case, the digest will be suggested.

For the push command, it doesn't make sense to suggest images without tags.

Finally, during my testing of image completion, there was one spot where I found two log messages that were incorrect. I fixed them in the commit for images. Also, the `Use` description for the `images` command seemed incorrect to me, so I updated it and ran `make gen-docs`.

Issues:
https://github.com/acorn-io/acorn/issues/841
https://github.com/acorn-io/acorn/issues/1023

### DIFF
--- a/pkg/cli/completion.go
+++ b/pkg/cli/completion.go
@@ -207,3 +207,19 @@ func imagesCompletion(allowDigest bool) completionFunc {
 		return result, nil
 	}
 }
+
+func credentialsCompletion(ctx context.Context, c client.Client, toComplete string) ([]string, error) {
+	credentials, err := c.CredentialList(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	var result []string
+	for _, credential := range credentials {
+		if strings.HasPrefix(credential.Name, toComplete) {
+			result = append(result, credential.Name)
+		}
+	}
+
+	return result, nil
+}
--- a/pkg/cli/completion.go
+++ b/pkg/cli/completion.go
@@ -239,3 +239,19 @@ func volumesCompletion(ctx context.Context, c client.Client, toComplete string) 
 
 	return result, nil
 }
+
+func secretsCompletion(ctx context.Context, c client.Client, toComplete string) ([]string, error) {
+	secrets, err := c.SecretList(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	var result []string
+	for _, secret := range secrets {
+		if strings.HasPrefix(secret.Name, toComplete) {
+			result = append(result, secret.Name)
+		}
+	}
+
+	return result, nil
+}
--- a/pkg/cli/completion.go
+++ b/pkg/cli/completion.go
@@ -1,0 +1,177 @@
+package cli
+
+import (
+	"context"
+	"strings"
+
+	"github.com/acorn-io/acorn/pkg/client"
+	"github.com/acorn-io/baaah/pkg/typed"
+	"github.com/spf13/cobra"
+	"k8s.io/utils/strings/slices"
+)
+
+type completionFunc func(context.Context, client.Client, string) ([]string, error)
+
+// These define instances when the completion should not happen and the default completion (usually file/directory
+// completion should occur). For example, exec should only have one argument, everything after that is completed as
+// default by the user's terminal.
+type noCompletionOption func([]string) bool
+
+type completion struct {
+	client           client.ClientFactory
+	completionFunc   completionFunc
+	successDirective cobra.ShellCompDirective
+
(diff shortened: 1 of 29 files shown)


pkg/cli/testdata/MockClient.go
```diff
@@ -336,17 +377,20 @@ func (m *MockClient) VolumeDelete(ctx context.Context, name string) (*apiv1.Volu
...
+	}, {
 		TypeMeta:   metav1.TypeMeta{},
 		ObjectMeta: metav1.ObjectMeta{Name: "found-image-no-tag"},
-	}, apiv1.Image{
+		Digest:     "lkjhgfdsa0987654321",
+	}, {
 		TypeMeta:   metav1.TypeMeta{},
 		ObjectMeta: metav1.ObjectMeta{Name: "found-image-two-tags1234567"},
-		Tags:       []string{"testtag1:latest", "testtag2:latest"},
+		Tags:       []string{"testtag1:latest", "testtag2:v1"},
```
thedadams, line 392: I wanted an image with a tag that wasn't `latest`, so I made that change here. This caused some of the existing test to be updated.

pkg/cli/images.go
```diff
@@ -16,14 +16,16 @@ import (
 )
 
 func NewImage(c client.CommandContext) *cobra.Command {
-	cmd := cli.Command(&Image{client: c.ClientFactory}, cobra.Command{
-		Use:     "image [flags] [APP_NAME...]",
+	image := &Image{client: c.ClientFactory}
+	cmd := cli.Command(image, cobra.Command{
+		Use:     "image [flags] [IMAGE_REPO:TAG|IMAGE_ID]",
```
thedadams, line 21: The `Use` here was inconsistent with what I see in the code and from the CLI. I made that adjust here and ran `make gen-docs`.

pkg/cli/completion.go
```diff
@@ -0,0 +1,257 @@
...
+
+	result, err := a.completionFunc(cmd.Context(), c, toComplete)
+	if err != nil {
+		return nil, cobra.ShellCompDirectiveError
+	}
+
+	return removeExistingArgs(result, args), a.successDirective
+}
+
+func appsThenContainersCompletion(ctx context.Context, c client.Client, toComplete string) ([]string, error) {
```
tylerslaton, line 83: Thoughts on moving these completion functions out of here and into their relevant `cli/<command>.go files`? This would make it so that this file doesn't expand too much overtime and keep more relevant code together. Then the other boilerplate settings (like `withSuccessDirective` and `completion`) would stay here.

tylerslaton (commented):
Awesome feature! Only one real question I had.

pkg/cli/completion.go
thedadams: The main reason I didn't do that is that some of these are used in multiple places. For example, the apps completion is used in a number of commands (apps, start, stop, rm, etc).

Secondly, these functions can be unit tested independent of the commands themselves, which is how I implemented the tests.

I'm open to moving them though. After reading my explanation, do you still feel like moving them would be better?

tylerslaton: Nope! I'm with you on the benefits of it being like it is now.

thedadams: Awesome! Thanks for the review, @tylerslaton!

pkg/cli/testdata/MockClient.go
```diff
@@ -14,24 +14,43 @@ import (
 )
 
 type MockClientFactory struct {
+	AppList        []apiv1.App
```
jsilverio22, line 17: This is very helpful for testing!

jsilverio22 (approved):
Great to have this functionality now. LGTM

tylerslaton (approved):
Great work here!

cjellick (approved):

Issue #1041, opened 2023-01-16
Disable the feature that allows users to set labels and annotations


cjellick:
We introduced a feature that allows labels and annotations to be set in an acornfile or as part of `acorn run` and have those be propagated down through the native kubernetes objects. That feature is described here:
- https://docs.acorn.io/authoring/labels
- https://docs.acorn.io/running/labels

We need an install flag that can turn this feature completely off. We want to be able to turn it off because it poses a security concern in Hub. Setting labels/annotations could allow a user to opt pods into permissions or abilities that they shouldn't have.

sangee2004:
Tested with `acorn version v0.5.0-rc3-9-g4e16d33+4e16d334`

 
1.  Default value  for  `ignore-user-labels-and-annotations` is false. User is allowed to create apps with user labels and annotations.
2. Once this flag id set to `true`, all existing pods with user labels and annotations get redeployed with no user labels and annotations.
3. When this flag is set to `false` , all existing pods with user labels and annotations get redeployed with all the user labels and annotations that were ignored  because of this flag being set to `true`
4. With this flag set to `true`, creating/updating new apps with the labels and annotation ignores labels and annotations silently and pods don’t get the  labels/annotations passed in the Acorn file using `acorn run` and `acorn run —update`.
  5. With flag set to `true`,  creating/updating new apps with the labels and annotation ignores labels and annotations silently and pods don’t get the  labels/annotations set using `—label` and `—annotation`  flags in `acorn run` and `acorn update`.   All the above behaviors are also seen with secrets and volumes.

PR #1054, opened 2023-01-18
Add flag to disable propagation of user-defined labels and annotations

thedadams:
The user is still free to attempt to add labels and annotations to the various objects, but the controller will ignore them when creating the other Kubernetes objects. If the flag is later turned off (so labels and annotations are allowed to propagate) then new objects will be created with the desired labels and annotations.

Two tests are also added to test the labels and annotations functionality with and without propagation.

The flag added is sort of long, but I wanted to ensure that it was explicit about what it was doing.

Issue: https://github.com/acorn-io/acorn/issues/1041


pkg/controller/appdefinition/deploy.go
```diff
@@ -32,6 +32,22 @@ import (
...
 
+func RemoveLabelsAndAnnotationsConfig(h router.Handler) router.Handler {
+	return router.HandlerFunc(func(req router.Request, resp router.Response) error {
+		appInstance := req.Object.(*v1.AppInstance)
+		cfg, err := config.Get(req.Ctx, req.Client)
+		if err != nil {
+			return err
+		}
+
+		if cfg.IgnoreUserLabelsAndAnnotations != nil && *cfg.IgnoreUserLabelsAndAnnotations {
```
cjellick, line 43: A tip @ibuildthecloud gave me: For nil bools feel free to add GetIgnoreUserLabelsAndAnnotations() method like the GetDevMode() method on appInstanceSpec. Makes the code more readable.

pkg/controller/appdefinition/testdata/deployspec/labels/existing.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
```
cjellick, line 1: Why did you need to add this file?

cjellick: wait, this whole dir is new. its a whole new test?

cjellick: resolving, disregard

cjellick (changes requested):
Just one small nitpck

pkg/controller/appdefinition/deploy.go
thedadams: I took a slightly different approach here because `config.Get` "completes" the config object with defaults.

jsilverio22 (approved):

pkg/controller/appdefinition/deploy.go
jsilverio22: for my benefit..this is just to check for bools in AppInstanceSpec are there and set?

cjellick: Which part @jsilverio22 ? Really anywhere we have nilable bools, you can add a helper to do this type of logic, EXCEPT for config because as donnie is pointing out, it isnt necessary because the config.Get is responsible for ensure its not nil

cjellick (approved):

pkg/controller/appdefinition/deploy.go
jsilverio22: I was looking at the GetDevMode(). But that makes sense bundling it up for clarity and why it is not necessary in this case.

PR #1055, opened 2023-01-18
Cut generated app url hash suffix to 8 characters

StrongMonkey:
The current generated url uses the first 12 characters of a hash of serviceName + appName + namespace, which is a bit longer and undesirable. Cut down to 8 to make url a little bit nicer.

This change will probably need to be made into changelog for the next release as it changes the format of user's app url.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -32,7 +32,7 @@ var (
 	DefaultImageCheckIntervalDefault = "5m"
 
 	// Default HttpEndpointPattern set to enable Let's Encrypt
-	DefaultHttpEndpointPattern = "{{printf \"%s-%s-%s\" .Container .App .Hash | truncate}}.{{.ClusterDomain}}"
+	DefaultHttpEndpointPattern = "{{hashConcat 8 .Container .App .Namespace | truncate}}.{{.ClusterDomain}}"
 )
 
 func complete(ctx context.Context, c *apiv1.Config, getter kclient.Reader) error {
--- a/pkg/publish/ingress.go
+++ b/pkg/publish/ingress.go
@@ -78,13 +78,14 @@ func toEndpoint(pattern, domain, container, appName, appNamespace string) (strin
 		App:           appName,
 		Container:     container,
 		Namespace:     appNamespace,
-		Hash:          getAppHash(container, appName, appNamespace),
+		Hash:          hash(8, strings.Join([]string{container, appName, appNamespace}, ":")),
 		ClusterDomain: strings.TrimPrefix(domain, "."),
 	}
 
 	var templateBuffer bytes.Buffer
 	t := template.Must(template.New("").Funcs(map[string]any{
-		"truncate": truncate,
+		"truncate":   truncate,
+		"hashConcat": hashConcat,
 	}).Parse(pattern))
 	if err := t.Execute(&templateBuffer, endpointOpts); err != nil {
 		return "", fmt.Errorf("%w %v: %v", ErrInvalidPattern, pattern, err)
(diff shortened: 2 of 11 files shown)


pkg/config/config.go
```diff
@@ -32,7 +32,7 @@ var (
 	DefaultImageCheckIntervalDefault = "5m"
 
 	// Default HttpEndpointPattern set to enable Let's Encrypt
-	DefaultHttpEndpointPattern = "{{printf \"%s-%s-%s\" .Container .App .Hash | truncate}}.{{.ClusterDomain}}"
+	DefaultHttpEndpointPattern = "{{getAppFullName .Container .App .Namespace 8 | truncate}}.{{.ClusterDomain}}"
```
StrongMonkey, line 35: Made a new function to make it easiler to pass parameter and limit of hash

pkg/publish/ingress.go
```diff
@@ -104,13 +104,19 @@ func toEndpoint(pattern, domain, container, appName, appNamespace string) (strin
 	return templateBuffer.String(), nil
 }
 
-func getAppHash(serviceName, appName, appNamespace string) string {
+/*
+getAppFullName takes containerName, appName and namespace, convert it to the format of
+`${containerName}-${appName}-${hash of `containerName:appName:namespace`}, with
+the ability to specify the length of the hash suffix
+*/
+func getAppFullName(serviceName, appName, appNamespace string, limit int) string {
```
StrongMonkey, line 112: made it a go template function so that user can override the limit as they want.

StrongMonkey:
This will reconsile all existing workloads to short name if users don't put a default value.

It will break if user uses cname record to point the old DNS name. 

To revert the behavior, pass `{{getAppFullName .Container .App .Namespace 12 | truncate}}.{{.ClusterDomain}}"` to `--http-endpoint-pattern`

ibuildthecloud (approved):

pkg/publish/ingress.go
```diff
@@ -78,13 +78,13 @@ func toEndpoint(pattern, domain, container, appName, appNamespace string) (strin
 		App:           appName,
 		Container:     container,
 		Namespace:     appNamespace,
-		Hash:          getAppHash(container, appName, appNamespace),
```
ibuildthecloud, line 81: Can you keep this field here and have it call hashConcat with size 8.  Otherwise there's no way to update production which is using the `Hash` field in the template pattern.

ibuildthecloud (changes requested):

pkg/publish/ingress.go
StrongMonkey: Oh shoot, I will just keep it

StrongMonkey: Ok, I added this back.

tylerslaton (commented):
Changes look good! Would you mind adding test cases to `ingress_test.go` that accounts for the various states of `hashConcat`? For example - what happens if we give it something other than 8 or some invalid input.

ibuildthecloud (dismissed):
Can you address the comment to add a unit test for edge case of hashConcat

StrongMonkey:
@tylerslaton unit tests are added 👍

tylerslaton (approved):
Won't block as this is a minor thing, but what happens if I set something like:

```
{{hashConcat 8 .Foo .Bar .Baz | truncate}}-{{.Namespace}}.{{.ClusterDomain}}
```

Would be good to see that in a test as a well so we know the UX of erroneous templates passed is consistent.

StrongMonkey:
@tylerslaton Good point! Added

cjellick (approved):

PR #1046, opened 2023-01-17
Add registry to acorn image print outs and fix query based off specific tag in acorn image (#1043 #1044)

jsilverio22:
This PR  helps provide clarity to `acorn image` outputs by including the registry to console. It also address a bug that would occur when running `acorn image <tag>` where all tags attached to that image would be outputted. The expected behavior of only displaying the queried tag should be achieved now.

Issues #1043 #1044 
Signed-off-by: Joshua Silverio <joshua@acorn.io>

Issue #868, opened 2022-11-14
CLI not testable


jsilverio22:
Acorn CLI has no tests / test framework

PR #1062, opened 2023-01-18
refactor cli unit tests to specify app, credential, container, image, secret or volume to return in test (#868)

jsilverio22:
Per comments on PR #1046 this PR will allow devs to specify items to use in CLI unit tests. For more special cases, the switch statement can still be used however for general best practice defining the Item and adding to the MockClientFactory in your tests will allow for easier testing of CLI.

Issue: #868

### DIFF
--- a/pkg/cli/completion_test.go
+++ b/pkg/cli/completion_test.go
@@ -129,13 +129,29 @@ func TestAppsThenContainersCompletion(t *testing.T) {
 }
 
 func TestAppsCompletion(t *testing.T) {
+	var DefaultImageList = []apiv1.Image{{
+		TypeMeta:   metav1.TypeMeta{},
+		ObjectMeta: metav1.ObjectMeta{Name: "found-image1234567"},
+		Tags:       []string{"testtag:latest"},
+		Digest:     "1234567890asdfghkl",
+	}, {
+		TypeMeta:   metav1.TypeMeta{},
+		ObjectMeta: metav1.ObjectMeta{Name: "found-image-no-tag"},
+		Digest:     "lkjhgfdsa0987654321",
+	}, {
+		TypeMeta:   metav1.TypeMeta{},
+		ObjectMeta: metav1.ObjectMeta{Name: "found-image-two-tags1234567"},
+		Tags:       []string{"testtag1:latest", "testtag2:v1"},
+		Digest:     "lkjhgfdsa1234567890",
+	}}
 	names := []string{"test-1", "acorn-1", "acorn-2", "hub", "test-2"}
 	apps := make([]apiv1.App, 0, len(names))
 	for _, name := range names {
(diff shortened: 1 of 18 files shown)


pkg/cli/completion_test.go
```diff
@@ -129,13 +129,29 @@ func TestAppsThenContainersCompletion(t *testing.T) {
 }
 
 func TestAppsCompletion(t *testing.T) {
+	var DefaultImageList = []apiv1.Image{{
+		TypeMeta:   metav1.TypeMeta{},
```
thedadams, line 133: nit: You don't need this like if `TypeMeta` is getting set to its "zero" value. Here and other tests altered in these changes.

pkg/cli/testdata/MockClient.go
```diff
@@ -382,76 +451,24 @@ func (m *MockClient) VolumeDelete(ctx context.Context, name string) (*apiv1.Volu
 }
 
 func (m *MockClient) ImageList(ctx context.Context) ([]apiv1.Image, error) {
-	return []apiv1.Image{{
```
thedadams, line 385: Personally, I am a fan of leaving this in and just adding the `if` check at the beginning. That way, if a user only wants to use a basic list of images, they can do that with minimal effort.

You could also remove the `DefaultImagesList` from the tests in this case. What are your thoughts on this?

thedadams (approved):
A couple small comments. Overall I really like these changes. Thanks!

pkg/cli/testdata/MockClient.go
jsilverio22: I was going back and forth on this so, I don't mind adding it back in

pkg/cli/completion_test.go
jsilverio22: yeah, removed now

pkg/cli/testdata/MockClient.go
```diff
@@ -400,58 +472,17 @@ func (m *MockClient) ImageList(ctx context.Context) ([]apiv1.Image, error) {
...
 func (m *MockClient) ImageDelete(ctx context.Context, name string, opts *client.ImageDeleteOptions) (*apiv1.Image, error) {
+	if m.ImageItem != nil {
+		return m.ImageItem, nil
+	}
 	switch name {
-	case "dne":
-		return nil, nil
-	case "found-image1234567":
-		return &apiv1.Image{}, nil
 	case "found-image-two-tags1234567":
```
tylerslaton, line 486: Maybe we could export `const` values for the common switch values that these can use? That would make the package a bit more discoverable for new devs coming in to write tests. That way as a developer I could name my object `testdata.FoundWithTwoTags` or something akin to that. Thoughts?

tylerslaton (approved):

PR #1079, opened 2023-01-20
Wait until digests are accessible

ibuildthecloud:
Due to storage backends that are eventually consistent it is possible
that a digest that was just written is not immediately available for
read or we get a 404 after a successful read.  For digests that we just
created we will retry for 5 seconds trying to get the result.

### DIFF
--- a/pkg/build/assemble.go
+++ b/pkg/build/assemble.go
@@ -3,6 +3,7 @@ package build
 import (
 	"context"
 	"fmt"
+	"time"
 
 	v1 "github.com/acorn-io/acorn/pkg/apis/internal.acorn.io/v1"
 	"github.com/acorn-io/baaah/pkg/typed"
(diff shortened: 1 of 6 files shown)


pkg/build/assemble.go
```diff
@@ -172,13 +174,31 @@ func imagePlatform(img ggcrv1.Image) (*ggcrv1.Platform, error) {
 	}, nil
 }
 
+// retryGet will keep trying to get a digest for 5 seconds until it succeeds. This is specifically used for digests
+// we just created. For example, in ECR this call will sometimes return 404 while I assume S3 is becoming eventually
+// consistent. It is possible that you do GET and find the response and then do GET and get a 404. So just
+// keep trying until we get it.
+func retryGetImage(d name.Digest, opts []remote.Option) (result ggcrv1.Image, err error) {
+	for i := 0; i < 5; i++ {
```
tylerslaton, line 182: Are we sure 5 seconds is long enough? Maybe we could up it to more than we need just in case any network issues occur.

tylerslaton: Oop, this merged while I was reviewing. Not a huge thing regardless.

PR #1080, opened 2023-01-20
Break out reading logs into separate role

ibuildthecloud:
Right now the view role includes reading logs, which depending on your
circumstance you may or may not want.  This change pulls the logs
action into a separate role "view-logs" so a user must be assigned both
view and view-logs to have the previous behavior of view.

### DIFF
--- a/pkg/roles/roles.go
+++ b/pkg/roles/roles.go
@@ -7,10 +7,11 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


cjellick (approved):

cjellick:
I'm fine if you want to merge this for v0.5.0 @ibuildthecloud

Issue #1037, opened 2023-01-13
Ability to provide a custom certificate authority bundle to the acorn control plane


cjellick:
When an organization uses an internal registry, it is often secured with a TLS certificate signed by a private Certificate Authority. When that's the case, clients will only trust the certificate presented by the registry if the signing authority is in the CA bundle that the client has access to. In containerized applications, the CA is usually whatever is part of the base OS of the container. This is the case with the acorn control plane components: api-server, controller, builders.

So, to allow these components to properly integrate with the registry, we need to give the person installing/administering the acorn control plane a way to override the default CA bundle.

Proposed solution is for the administrator to put their CA bundle into a secret that will be recognized by the acorn installation process and bound into the acorn control plane components.

StrongMonkey:
@sangee2004 this is how I test this during development, but take it as a grain of salt.

### Prepare a registry

First you will need to make sure the registry runs in static IP/hostname so that you can put the SAN into your self-signed certificate. I run registry in Digitialocean.

1. Create a digitalocean instance. Generate a certificate with the public IP of the instance.

```
openssl req -x509 -nodes -days 825 -newkey rsa:4096 \
    -addext "subjectAltName = IP:$replace_your_ip" \
    -keyout server.key -out server.crt
```

2. Run the registry service with the certs mounted.

```
docker run -d \
  --restart=always \
  --name registry \
  -v /path/to/server.crt:/certs/server.crt -v /path/to/server.key:/certs/server.key \
  -e REGISTRY_HTTP_TLS_CERTIFICATE=/certs/server.crt \
  -e REGISTRY_HTTP_TLS_KEY=/certs/server.key -e REGISTRY_HTTP_ADDR=0.0.0.0:443\
  -p 443:443 \
  registry:2
```

### Generate a **full** CA chain with the private CA include.

Since we are replacing the whole CA chain in acorn controlplane, you will need to grab the full chain, otherwise acorn won't be able to trust other CAs. To do so, I run ubuntu container and just copy the certs out.


```
docker run -it ubuntu sleep 10000

docker cp $ubuntu_container_id:/etc/ssl/certs/ca-certificates.crt ~/ca-certificates.crt
```

### Append your private CA

Now append the private CA into the cert you just copied, and create secrets based on the file

```
cat ~/server.crt >> ~/ca-certificates.crt

kubectl -n acorn-image-system create secret generic cabundle --from-file=ca-certificates.crt=~/ca-certificates.crt

[...]

PR #1083, opened 2023-01-20
Add ability to add custom bundle

StrongMonkey:

Add the ability for acorn controlplane to override default CA certificates. The main use case is for user to configure an internal registry that is signed by private CA, and have acorn trust that registry.

User needs to add the secrets into the cluster, then configure acorn to use the bundle. 

As @ibuildthecloud  mentioned before, he wants acorn to automatically consume the secret when it exists. I spent some work trying to figure this out but couldn't find out a reliable way to do this in k8s without having to overriding the exist CA store if secret doesn't exist. For now we just add a new install option to tell acorn to consume the secret.

Will provide detailed instruction on how I test private registry in development under https://github.com/acorn-io/acorn/issues/1037


pkg/imagesystem/buildertemplate.go
```diff
@@ -11,189 +11,209 @@ import (
...
-								Name: "socket",
-								VolumeSource: corev1.VolumeSource{
-									EmptyDir: &corev1.EmptyDirVolumeSource{},
-								},
+					},
+					Volumes: []corev1.Volume{
+						{
+							Name: "socket",
+							VolumeSource: corev1.VolumeSource{
+								EmptyDir: &corev1.EmptyDirVolumeSource{},
```
StrongMonkey, line 190: Nothing really changes here, just moving it out so we can modify the object based on custom bundle.

pkg/install/install.go
```diff
@@ -481,7 +486,7 @@ func Namespace() ([]kclient.Object, error) {
 	return objectsFromFile("namespace.yaml")
 }
 
-func Deployments(runtimeImage string, apiServerReplicas, controllerReplicas int) ([]kclient.Object, error) {
+func Deployments(runtimeImage string, apiServerReplicas, controllerReplicas int, useCustomCabundle bool) ([]kclient.Object, error) {
```
ibuildthecloud, line 489: the lowercase form should be `useCustomCABundle`

pkg/install/install.go
```diff
@@ -430,7 +435,7 @@ func printObject(image string, opts *Options) error {
 	return err
 }
 
-func applyDeployments(ctx context.Context, imageName string, apiServerReplicas, controllerReplicas int, apply apply.Apply, c kclient.Client) error {
+func applyDeployments(ctx context.Context, imageName string, apiServerReplicas, controllerReplicas int, useCustomBundle bool, apply apply.Apply, c kclient.Client) error {
```
ibuildthecloud, line 438: can you keep this consistent as `useCustomCABundle`

pkg/install/controller.yaml
```diff
@@ -28,3 +28,4 @@ spec:
           securityContext:
             runAsUser: 1000
       serviceAccountName: acorn-system
+
```
ibuildthecloud, line 31: Unneeded change?

pkg/apis/api.acorn.io/v1/types.go
```diff
@@ -335,6 +335,7 @@ type Config struct {
 	IgnoreUserLabelsAndAnnotations *bool          `json:"ignoreUserLabelsAndAnnotations" name:"ignore-user-labels-and-annotations" usage:"Don't propagate user-defined labels and annotations to dependent objects"`
 	WorkloadMemoryDefault          *int64         `json:"workloadMemoryDefault" name:"workload-memory-default" quantity:"true" usage:"Set the default memory for acorn workloads. Accepts binary suffixes (Ki, Mi, Gi, etc) and \".\" and \"_\" seperators" short:"m" default:"0"`
 	WorkloadMemoryMaximum          *int64         `json:"workloadMemoryMaximum" name:"workload-memory-maximum" quantity:"true" usage:"Set the maximum memory for acorn workloads. Accepts binary suffixes (Ki, Mi, Gi, etc) and \".\" and \"_\" seperators" default:"0"`
+	UseCustomCABundle              *bool          `json:"useCustomCabundle" name:"use-custom-cabundle" usage:"Use custom cabundle that is stored in secrets for all acorn controlplane"`
```
ibuildthecloud, line 338:
```suggestion
	UseCustomCABundle              *bool          `json:"useCustomCABundle" name:"use-custom-ca-bundle" usage:"Use CA bundle for admin supplied secret for all acorn control plane components"`
```

pkg/cli/testdata/info/info_test_json.txt
```diff
@@ -35,7 +35,8 @@
             "internalRegistryPrefix": null,
             "ignoreUserLabelsAndAnnotations": null,
             "workloadMemoryDefault": null,
-            "workloadMemoryMaximum": null
+            "workloadMemoryMaximum": null,
+            "useCustomCABundle": null,
```
tylerslaton, line 39: This is apart of what is breaking your tests right now. Its a little picky 😅 
```suggestion
            "useCustomCABundle": null
```

pkg/cli/testdata/info/info_test.txt
```diff
@@ -28,6 +28,7 @@ server:
     setPodSecurityEnforceProfile: null
     workloadMemoryDefault: null
     workloadMemoryMaximum: null
+    useCustomCABundle: null
```
tylerslaton, line 31:
```suggestion
    useCustomCABundle: null
    workloadMemoryDefault: null
    workloadMemoryMaximum: null
```

pkg/cli/testdata/info/info_test.txt
```diff
@@ -53,5 +54,6 @@ server:
...
```
tylerslaton, line 57:
```suggestion
    useCustomCABundle: null
    workloadMemoryDefault: null
    workloadMemoryMaximum: null
```

tylerslaton (commented):
These changes should fix the failing tests right now. Some of the CLI tests are pretty picky so its just moving stuff around and making sure the text is exactly right with your new flag.

pkg/apis/api.acorn.io/v1/types.go
```diff
@@ -335,6 +335,7 @@ type Config struct {
 	IgnoreUserLabelsAndAnnotations *bool          `json:"ignoreUserLabelsAndAnnotations" name:"ignore-user-labels-and-annotations" usage:"Don't propagate user-defined labels and annotations to dependent objects"`
 	WorkloadMemoryDefault          *int64         `json:"workloadMemoryDefault" name:"workload-memory-default" quantity:"true" usage:"Set the default memory for acorn workloads. Accepts binary suffixes (Ki, Mi, Gi, etc) and \".\" and \"_\" seperators" short:"m" default:"0"`
 	WorkloadMemoryMaximum          *int64         `json:"workloadMemoryMaximum" name:"workload-memory-maximum" quantity:"true" usage:"Set the maximum memory for acorn workloads. Accepts binary suffixes (Ki, Mi, Gi, etc) and \".\" and \"_\" seperators" default:"0"`
+	UseCustomCABundle              *bool          `json:"useCustomCABundle" name:"use-custom-ca-bundle" usage:"Use CA bundle for admin supplied secret for all acorn control plane components"`
```
tylerslaton, line 338: Can we call out what the default is here for the user?

pkg/controller/builder/builder.go
```diff
@@ -44,7 +44,7 @@ func createBuilderObjects(req router.Request, resp router.Response) (string, str
 		forNamespace = builder.Namespace
 	}
 
-	objs := imagesystem.BuilderObjects(name, system.ImagesNamespace, forNamespace, system.DefaultImage(), pubKey, privKey, builder.Status.UUID, registryDNS)
+	objs := imagesystem.BuilderObjects(name, system.ImagesNamespace, forNamespace, system.DefaultImage(), pubKey, privKey, builder.Status.UUID, registryDNS, *cfg.UseCustomCABundle)
```
tylerslaton, line 47: If `cfg.UseCustomCABundle` is `nil` when we reach this point we'll get a panic.

Example: https://go.dev/play/p/3mChCxI3yAa

pkg/install/install.go
```diff
@@ -200,7 +205,7 @@ func Install(ctx context.Context, image string, opts *Options) error {
 	s.Success()
 
 	s = opts.Progress.New(fmt.Sprintf("Installing APIServer and Controller (image %s)", image))
-	if err := applyDeployments(ctx, image, *opts.APIServerReplicas, *opts.ControllerReplicas, apply, c); err != nil {
+	if err := applyDeployments(ctx, image, *opts.APIServerReplicas, *opts.ControllerReplicas, *opts.UseCustomCABundle, apply, c); err != nil {
```
tylerslaton, line 208: Any particular reason to not use `opts.Config.UseCustomCABundle` here? Same question applies to everything below.

tylerslaton (commented):
Initial pass of code review. Looking good so far 🎸

pkg/apis/api.acorn.io/v1/types.go
StrongMonkey: It should be obvious if user didn't provide the flag. But I can add it.

pkg/controller/builder/builder.go
StrongMonkey: I think this is always initilized so that it is not nil. https://github.com/acorn-io/acorn/blob/2141a882513218dd845b1aa33ef9f9ad8928979f/pkg/config/config.go#L97

pkg/install/install.go
StrongMonkey: The main problem with this is that there is an option in install called `install -o yaml` which print out the deployment object in yaml without saving anything into the config. So in here if you reference anything in config it will be nil. So we use the static flag here. We still save the flag into the config so that builder deployment reference the config and add the volumn mount if needed.

pkg/controller/builder/builder.go
tylerslaton: Yep, that's right. Thanks for helping me to remember.

tylerslaton:
#1037

docs/docs/30-installation/02-options.md
```diff
@@ -31,6 +31,29 @@ Let's Encrypt integration is only useful if you are running a non-local Kubernet
 ## Endpoint domain names
 Acorn provides several installation options for controlling the domain name used to generate endpoints. These are outlined in detail on our [networking page](/running/networking#dns).
 
+## Custom CA bundle
+Acorn allows user to provide custom ca bundle so that users can add own private CA and have acorn trust that CA. The most common use case is for acorn to trust an internal image registry that is signed by private CA.
```
tylerslaton, line 35: nit:
```suggestion
Acorn allows a user to provide a custom certificate authority (CA) bundle so that users can add their own private CA that acorn will trust. The most common use case is for acorn to trust an internal image registry that is signed by a private CA.
```

docs/docs/30-installation/02-options.md
```diff
@@ -31,6 +31,29 @@ Let's Encrypt integration is only useful if you are running a non-local Kubernet
...
+
+To do so, you will need to add the following steps:
```
tylerslaton, line 37: nit:
```suggestion
To do so, you will need to go through the following steps:
```

docs/docs/30-installation/02-options.md
```diff
@@ -31,6 +31,29 @@ Let's Encrypt integration is only useful if you are running a non-local Kubernet
...
+1. Provide your CA certificate chain in the following secret.
+
+```bash
+kubectl -n acorn-image-system create secret generic cabundle --from-file=ca-certificates.crt=/path/to/your/ca-certificates.crt
+
+kubectl -n acorn-system create secret generic cabundle --from-file=ca-certificates.crt=/path/to/your/ca-certificates.crt
+```
+
+:::info
+You must provide **full** CA certificate chain as it will override existing CA certificates in acorn control plane.
```
tylerslaton, line 48:
```suggestion
You must provide the **full** CA certificate chain as it will override existing CA certificates in acorn control plane.
```

docs/docs/30-installation/02-options.md
```diff
@@ -31,6 +31,29 @@ Let's Encrypt integration is only useful if you are running a non-local Kubernet
...
+:::
+
+
+2. Install acorn with the following option
+
+```bash
+acorn install --use-custom-cabundle
```
tylerslaton, line 55: Just a thought around UX, should I be able to pass the name of the secret in the form of `namespace/secret`? This should allow clusters with existing CA secrets to not need to create new ones.

I don't want to block this PR with this, so I'll just put it out their for us to think about. If its something we want to do, we can do it in a follow-up outside of the release.

pkg/install/install.go
```diff
@@ -543,6 +557,41 @@ func replaceImage(image string, objs []kclient.Object) ([]kclient.Object, error)
...
+					},
+				}
+			}
+			if err := unstructured.SetNestedSlice(ustr.Object, containers, "spec", "template", "spec", "containers"); err != nil {
+				return nil, err
+			}
+
+			volumes := []interface{}{
+				map[string]any{
+					"name": "cabundle",
```
tylerslaton, line 581: nit: If we end up keeping this constant secret name that we point to, can we create a `const` for the name we're looking for?

tylerslaton (commented):
Had a few comments. Ones marked with `nit` are not blocking. I had one additional question to bring up around UX but I don't believe it is required to address in order to merge.

docs/docs/30-installation/02-options.md
StrongMonkey: Good point! @ibuildthecloud brought this up and we think at this point we might not want to make thing more complex.

See https://acorn-io.slack.com/archives/C03FU91U4SX/p1674491780902799?thread_ts=1674490264.688439&cid=C03FU91U4SX

tylerslaton: Awesome! Glad to see we already had this thought.

tylerslaton (approved):
Great work! Would be great to see the above changes implemented but they aren't blocking.

pkg/install/install.go
```diff
@@ -543,6 +557,41 @@ func replaceImage(image string, objs []kclient.Object) ([]kclient.Object, error)
 	return objs, nil
 }
 
+func replaceCabundleVolumes(objs []kclient.Object) ([]kclient.Object, error) {
```
thedadams, line 560:
```suggestion
func replaceCABundleVolumes(objs []kclient.Object) ([]kclient.Object, error) {
```

thedadams (approved):
Looks good, but I do have one question.

In the past when we have needed to replace or add a CA bundle in an image, we have had to run commands like `c_rehash` or `update-ca-certificates` (depending on the Linux flavor). Do we need to do anything like that here?

For reference to some code you may be familiar with: https://github.com/rancher/rancher/blob/9278cf152f9b98ae3c543c0f956af2ae6ab91368/package/entrypoint.sh#L44-L49

pkg/install/install.go
StrongMonkey: @thedadams Good question. In here we actually replace the whole CA bundle so there is no need to re-run `update-ca-certificates`. 

Also all the controller runs as user 1000 so we can't really touch the system cert dir, in the future we might allow user to mount additional cacerts and we can figure out a way to combine the certs with system default.

pkg/install/install.go
StrongMonkey: Changed this to use contants. 👍

pkg/install/install.go
StrongMonkey: Actually @tylerslaton this is a good point. I changed it so it can intialize the config when you run `install -o yaml` so we just use the flag in the config.

Issue #1056, opened 2023-01-18
Not able to set any acorn arguments when running apps in auto upgrade format.


sangee2004:
acorn version v0.5.0-rc2-2-g6cee142+6cee1422 

While trying to run an app in auto upgrade mode with invalid acorn arguments , parsing error is presented to the user which does not provide any useful information.
```
% acorn run index.docker.io/xxx/xxx:v#.#.# --interval 3
  ✗  ERROR:  could not parse reference: index.docker.io/sangeetha/myfirstacorn:v#.#.#
```

The same scenario is handled with better error message when not using auto upgrade mode .
```
 % acorn run index.docker.io/xx/xxx:v2.0.0 --interval 3
  ✗  ERROR:  unknown flag: --interval
```

The bigger issue here is that we are not able to set any acorn arguments when running apps in auto upgrade mode.

This run command works as expected when passing a valid acorn argument.
```
acorn run -n testlocalold mytest1:v0.0.2 --new=false 
```

When we try to run the same app in auto upgrade mode with args passed it fails with following error message:

```
%acorn run -n testlocalold mytest1:v#.#.# --new=false 
  ✗  ERROR:  could not parse reference: mytest1:v#.#.#

% acorn run -n testlocalold "mytest1:v#.#.#" --new=false 
  ✗  ERROR:  could not parse reference: mytest1:v#.#.#
```

thedadams:
This is actually not an error message problem.

The args passed after the image are "acorn args." In order to determine if the passed args are valid for an Acorn, the CLI will look at the Acorn image and parse its args.

In the case where you are providing an auto-upgrade image format, Acorn doesn't know which Acorn image to pull which causes the error. You get a different error when you pass a real image name because Acorn pulls the image and sees that the Acorn doesn't have an `--interval` option.

@cjellick Is this something we want to support?

sangee2004:
This problem happens even when valid acorn arguments are passed. I am changing the bug description to reflect this.

sangee2004:
Tested with `acorn version v0.5.1-28-g099c492+099c492f`

We are now able to set acorn arguments when launching apps in auto upgrade format.
When invalid acorn arguments are passed , user is presented with an error message.

```
% acorn run --interval=1m -n myt2 mynew:#.#.# --xx=false
  ✗  ERROR:  unknown flag: --xx
% acorn run --interval=1m -n myt2 mynew:#.#.# --new=false --xx=false
[...]

Issue #1032, opened 2023-01-13
Auto upgrade not working with * & ** special characters


samkulkarni20:
I have a package published with tag `v2.2-alpha` & tried to run it with below auto upgrade enabled commands, but it didn't work.
As you can see below, only the rules that contain `*` or `**` don't work. I do have the application image published with specific tag.
```
$ acorn run -n nginx-golang ghcr.io/samkulkarni20/nginx-golang:v#.#-* 
zsh: no matches found: ghcr.io/samkulkarni20/nginx-golang:v#.#-*
$ acorn run -n nginx-golang ghcr.io/samkulkarni20/nginx-golang:v#.#-**
zsh: no matches found: ghcr.io/samkulkarni20/nginx-golang:v#.#-**
$ acorn run -n nginx-golang ghcr.io/samkulkarni20/nginx-golang:v2.2-* 
zsh: no matches found: ghcr.io/samkulkarni20/nginx-golang:v2.2-*
```
My Acorn version:
```
$ acorn -v                                                           
acorn version v0.4.2+e23f68d4
```
Auto upgrade command with just `#` works fine
```
$acorn run -n nginx-golang ghcr.io/samkulkarni20/nginx-golang:v#.#   
nginx-golang

$ acorn apps                                                       
NAME           IMAGE                                     HEALTHY   UP-TO-DATE   CREATED   ENDPOINTS                                                              MESSAGE
nginx-golang   ghcr.io/samkulkarni20/nginx-golang:v2.1   2         2            13s ago   http://proxy-nginx-golang-1fe838eb7a71.local.on-acorn.io/ => proxy:80   OK
```
Images in my container registry:
![Screenshot from 2023-01-13 13-46-51](https://user-images.githubusercontent.com/15933023/212368142-a7c60b15-97d2-4862-b007-45e8fc98eff0.png)

thedadams:
Hi @samkulkarni20! Thank you for bringing this up.

As you can see from the message you have posted above, the error `no matches found` is actually coming from you shell `zsh`. This is because `zsh` sees the wildcard and tries to expand it. This is not what you want.

I think trying something like `acorn run -n nginx-golang "ghcr.io/samkulkarni20/nginx-golang:v#.#-*"` will work for you. Please let us know if you still have trouble.

cjellick:
closing. Feel free to reopen if this doesnt address your issue @samkulkarni20

cjellick:
[...]

PR #1098, opened 2023-01-23
Refactor and test auto-upgrade functionality

thedadams:
These changes are broken up into ~3~ 4 commits for reviewability purposes. These changes should be squashed before merging.

The first commit is where most of the changes occur. There were three ways to kick-off an `sync` previously:
1. A ticker based on the default auto upgrade config value
2. An on-demand sync call
3. After the sync was over, a goroutine would wait until the nearest calculated next time to upgrade and call sync again. This step also took into account the default auto upgrade config value.

Since 1 and 3 were doing similar things, they were combined to simplify the package-level state and make the code more testable.

Additionally, if an auto-upgrade-enabed app was deleted and recreated before the auto-upgrade sync ran, the next upgrade time for that app would still exist for the old version of the app and a user would have to wait for the next interval before the app would completely deploy. Now we detect when an app is new and sync it immediately.

The second commit is ONLY separating the `sync` method into multiple functions. The code is copied and pasted and none of it is changed with the possible exception of variable names.

The third commit adds tests to the code. Some changes were also made to the implementation for testability and mockability purposes.

A fourth commit has been added to address acorn args validation. The ImageDetails api-server endpoint now supports auto-upgrade tags and will get the latest tag for the pattern and return the ImageDetails for that image.

Issues:
https://github.com/acorn-io/acorn/issues/1032
https://github.com/acorn-io/acorn/issues/1056

### DIFF
--- a/pkg/autoupgrade/daemon.go
+++ b/pkg/autoupgrade/daemon.go
@@ -2,16 +2,11 @@ package autoupgrade
 
 import (
 	"context"
-	"fmt"
 	"strings"
-	"sync"
 	"time"
 
 	v1 "github.com/acorn-io/acorn/pkg/apis/internal.acorn.io/v1"
-	"github.com/acorn-io/acorn/pkg/autoupgrade/validate"
 	"github.com/acorn-io/acorn/pkg/config"
-	"github.com/acorn-io/acorn/pkg/images"
-	tags2 "github.com/acorn-io/acorn/pkg/tags"
 	"github.com/acorn-io/baaah/pkg/router"
 	imagename "github.com/google/go-containerregistry/pkg/name"
 	"github.com/sirupsen/logrus"
@@ -20,12 +15,7 @@ import (
 
 const defaultNoReg = "xxx-no-reg"
 
-var (
-	syncQueue       = make(chan struct{}, 1)
-	ticker          *time.Ticker
-	currentInterval string
-	m               sync.Mutex
-)
+var syncQueue = make(chan struct{}, 1)
 
 // Sync tells the daemon to trigger the image syncing logic
 func Sync() {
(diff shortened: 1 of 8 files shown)


pkg/autoupgrade/daemon.go
```diff
@@ -210,9 +186,9 @@ func (d *daemon) sync(ctx context.Context) error {
 		var tags []string
 		var pullErr error
 		if hasValidRegistry {
-			_, tags, pullErr = images.ListTags(ctx, d.client, imageKey.namespace, imageKey.image)
+			tags, pullErr = d.client.listTags(ctx, imageKey.namespace, imageKey.image)
 		}
-		localTags, err := tags2.GetTagsMatchingRepository(current, ctx, d.client, "acorn", defaultNoReg)
+		localTags, err := d.client.getTagsMatchingRepo(ctx, current, "acorn", defaultNoReg)
```
thedadams, line 191: Should `"acorn"` here be `imageKey.namespace` instead?

thedadams:
Note to reviewers: I will be making a significant change to the last commit in this PR, as well as adding an additional commit. I will comment again when this is done.

pkg/autoupgrade/daemon_test.go
```diff
@@ -0,0 +1,508 @@
...
+	kclient "github.com/acorn-io/acorn/pkg/k8sclient"
+	"github.com/acorn-io/baaah/pkg/router"
+	"github.com/acorn-io/baaah/pkg/typed"
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type mockClient struct {
```
tylerslaton, line 21: Just a thought, but maybe we could move this to the `testdata` package and rename this to be `mockDaemonClient`?

pkg/autoupgrade/daemon.go
```diff
@@ -41,239 +31,174 @@ func Sync() {
...
-	}
+func StartSync(ctx context.Context, client kclient.Client) {
+	d := newDaemon(client)
 
-	dd, err := time.ParseDuration(*cfg.AutoUpgradeInterval)
+	// Trigger one sync upon startup of the daemon
+	nextWait, err := d.sync(ctx, time.Now())
 	if err != nil {
-		return fmt.Errorf("failed to parse image check interval %v: %w", cfg.AutoUpgradeInterval, err)
+		logrus.Errorf("Encountered error syncing auto-upgrade apps: %v", err)
```
tylerslaton, line 53: Since this doesn't return an error any more, won't we start the controller with this sync never being started?

tylerslaton (commented):
Looking really good. Did a first pass at this but will come back to more sometime today or tomorrow.

pkg/controller/config/autoupgrade.go
```diff
@@ -2,21 +2,11 @@ package config
 
 import (
 	"github.com/acorn-io/acorn/pkg/autoupgrade"
-	"github.com/acorn-io/acorn/pkg/config"
 	"github.com/acorn-io/baaah/pkg/router"
 )
 
-// HandleAutoUpgradeInterval resets the ticker for auto-upgrade sync interval as it changes in the acorn config
+// HandleAutoUpgradeInterval resets the timer for auto-upgrade sync interval as it changes in the acorn config
 func HandleAutoUpgradeInterval(req router.Request, resp router.Response) error {
```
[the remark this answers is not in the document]
thedadams: They are required because this is passed as a `router.Handler`, but we can ignore them. I will update.

pkg/autoupgrade/daemon_test.go
thedadams: There isn't a `testdata` package here, and this mock is only valuable for unit tests of the `daemonClient`.

I did rename the type to `mockDaemonClient` as you suggested.

pkg/autoupgrade/daemon.go
thedadams: The timer will always be started because we will use the `config.DefaultImageCheckIntervalDefault` (which is `5m`) if there is an error getting the config or parsing the check interval from the config.

I did change this slightly because your comment made me realize that we shouldn't be returning `0` if we can get the config object. We should return the `config.DefaultImageCheckIntervalDefault` (`5m`). I updated as such.

pkg/autoupgrade/daemon_test.go
tylerslaton: Yep, totally valid conclusion here.

pkg/autoupgrade/daemon.go
tylerslaton: Ah right on. Thanks for clearing that up for me.

pkg/autoupgrade/daemon.go
```diff
@@ -41,239 +31,175 @@ func Sync() {
...
+}
+
+func newDaemon(c kclient.Client) *daemon {
+	return &daemon{
+		client:           &client{client: c},
+		appKeysPrevCheck: make(map[kclient.ObjectKey]time.Time),
+	}
 }
 
 // StartSync launches starts the daemon. It watches for new sync events coming and ensures a sync is triggered
```
cjellick, line 45:
```suggestion
// StartSync starts the daemon. It watches for new sync events coming and ensures a sync is triggered
```

pkg/autoupgrade/tags.go
```diff
@@ -1,14 +1,55 @@
...
+	if err != nil {
+		return "", false, err
+	}
+
+	newTag, err := FindLatest(imagename.DefaultTag, pattern, tags)
+	return strings.TrimPrefix(ref.Context().Tag(newTag).Name(), defaultNoReg+"/"), newTag != imagename.DefaultTag, err
+}
+
+// FindLatestTagForImageWithPattern will return the latest tag for image corresponding to the pattern.
+func FindLatestTagForImageWithPattern(ctx context.Context, c kclient.Client, namespace, image, pattern string) (string, bool, error) {
```
cjellick, line 49: why bother with both a public and private version? seems like the public is purely pass through, so why not _just_ have the public?

thedadams: I am using the `daemonClient` interface so I can mock it for unit tests. By having this exported version take a `kclient.Client`, I can keep the `daemonClient` an unexported implementation detail.

cjellick: ah, got it!

cjellick (changes requested):
Looks great! Thanks for cleaning this code up.

I noticed a random pre-existing typo in a comment. Want to suck that change in?

pkg/autoupgrade/daemon.go
thedadams: Updated.

thedadams:
I updated the comment and squashed all the commits.

pkg/autoupgrade/daemon.go
thedadams: @cjellick Before we merge, I just want to verify whether this is correct. With all the "project" work, maybe this should be `imageKey.namespace` (which is the `app.Namespace`)?

thedadams: EDIT: This isn't relevant with the recent changes, and, based on my investigation, this should be `imageKey.namespace` in the "project" paradigm.

tylerslaton (approved):
Gave this a test locally and thing are looking fine there. Great work here, code looks good 👍🏻

cjellick (approved):

cjellick:
looks good. will leave it to you to merge and assign over

Issue #1087, opened 2023-01-21
When invalid container names are passed in "-m" to `acorn run` , it gets ignored silently.


sangee2004:
acorn version v0.5.0-rc3-18-gfbe46a2+fbe46a25

In the following command ng is invalid container name and web is valid container name for the acorn file used:

` acorn run -m ng=10Mi,web=35Mi -n test6 .`

This results in memory limit for "web" container applied correctly as expected. But there is no error message to indicate that the other container name "ng" is incorrect.

cjellick:
This is an interesting one. It is a good thing to call out, but I'm not sure that it shoudl cause a validation error. Will need to get a broader opinion. @ibuildthecloud @vincent99 any thoughts?

cjellick:
Either way, non-blocking. Moving out

cjellick:
This wont be blocking either way, but want to get a broader set of opinions before we move it out of the milestone.

cjellick:
Quick discussion determined this should return a validation error

sangee2004:
acorn version v0.5.0-rc3-40-gc138b50+c138b504

Following error is presented to the user  when running app with invalid name - `acorn run -m ng=10Mi,web=35Mi -n test6 .` (in this case ng was invalid container name).
```
ERROR:  App.api.acorn.io "test6" is invalid: spec.memory: Invalid value: "ng": workload name set by user does not exist
```

PR #1101, opened 2023-01-23
handle invalid container names in "-m" to acorn run (#1087)

jsilverio22:
This PR handles when a user inputs a nonexistent container name in the `acorn run -m` flag it will now error out server side. 

Issue: #1087 

### DIFF
--- a/pkg/apis/internal.acorn.io/v1/memory.go
+++ b/pkg/apis/internal.acorn.io/v1/memory.go
@@ -52,7 +52,7 @@ var (
 	ErrInvalidAcornMemory   = errors.New("invalid memory from Acornfile")
 	ErrInvalidSetMemory     = errors.New("invalid memory set by user")
 	ErrInvalidDefaultMemory = errors.New("invalid memory default")
-	ErrInvalidWorkload      = errors.New("invalid workload name set by user")
+	ErrInvalidWorkload      = errors.New("workload name set by user does not exist")
 )
 
 func MemoryToRequirements(memSpec Memory, containerName string, container Container, specMemDefault, specMemMaximum *int64) (*corev1.ResourceRequirements, error) {
--- a/pkg/apis/internal.acorn.io/v1/memory.go
+++ b/pkg/apis/internal.acorn.io/v1/memory.go
@@ -52,6 +52,7 @@ var (
 	ErrInvalidAcornMemory   = errors.New("invalid memory from Acornfile")
 	ErrInvalidSetMemory     = errors.New("invalid memory set by user")
 	ErrInvalidDefaultMemory = errors.New("invalid memory default")
+	ErrInvalidWorkload      = errors.New("invalid workload name set by user")
 )
 
 func MemoryToRequirements(memSpec Memory, containerName string, container Container, specMemDefault, specMemMaximum *int64) (*corev1.ResourceRequirements, error) {
--- a/pkg/server/registry/apps/validator.go
+++ b/pkg/server/registry/apps/validator.go
@@ -299,10 +299,9 @@ func (s *Validator) checkPermissionsForPrivilegeEscalation(ctx context.Context, 
 
 func checkMemory(memory v1.Memory, workloads map[string]v1.Container, specMemDefault, specMemMaximum *int64) []*field.Error {
 	validationErrors := []*field.Error{}
-
 	err := validateMemoryRunFlags(memory, workloads)
 	if err != nil {
-		validationErrors = append(validationErrors, err)
+		validationErrors = append(validationErrors, err...)
 	}
 
 	for workload, container := range workloads {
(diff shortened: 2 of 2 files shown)


pkg/server/registry/apps/validator.go
```diff
@@ -319,6 +325,19 @@ func checkMemory(memory v1.Memory, workloads map[string]v1.Container, specMemDef
...
 
+func validateMemoryRunFlags(memory v1.Memory, workloads map[string]v1.Container) *field.Error {
+	for key, value := range memory {
+		if key == "" {
+			continue
+		}
+		if _, ok := workloads[key]; !ok {
+			path := field.NewPath("spec", "memory")
+			return field.Invalid(path, fmt.Sprintf("%s=%v", key, *value), v1.ErrInvalidWorkload.Error())
+		}
```
tylerslaton, line 336: nit: Simplifies the code a bit and uses the path pattern used in `checkMemory`
```suggestion
		if _, ok := workloads[key]; !ok {
			path := field.NewPath("spec", "memory", key)
			return field.Invalid(path, *value, v1.ErrInvalidWorkload.Error())
		}
```

pkg/server/registry/apps/validator.go
```diff
@@ -319,6 +325,19 @@ func checkMemory(memory v1.Memory, workloads map[string]v1.Container, specMemDef
 	return validationErrors
 }
 
+func validateMemoryRunFlags(memory v1.Memory, workloads map[string]v1.Container) *field.Error {
```
tylerslaton, line 328: Should this logic just be in `v1/memory.go` file as apart of `v1.ValidateMemory`?

tylerslaton (approved):
Two non-blocking comments.

pkg/server/registry/apps/validator.go
jsilverio22: I thought about that but it seemed like v1.validatememory handled more of the quantity logic/validation so it made sense to me to separate them out open to changing to other way if you or others have a strong opinion

pkg/server/registry/apps/validator.go
jsilverio22: any preference/thoughts?

pkg/server/registry/apps/validator.go
tylerslaton: Only real difference is that this same logic won't happen on the controller unless it exists in `v1.ValidateMemory`. Right now validation happens on both the controller and the API server.

pkg/server/registry/apps/validator.go
StrongMonkey: Should we mention in the error message that the container name doesn't exist/get defined the app? I think with the current error message, user still won't have an idea what's wrong the the specification.

thedadams: Change `*value` to `key`, as: `return field.Invalid(path, key, v1.ErrInvalidWorkload.Error())`

pkg/server/registry/apps/validator.go
jsilverio22: ahh okay in that case I will move to v1.validatememory

pkg/server/registry/apps/validator.go
tylerslaton: Yeah that's right. We want this to be as clear as possible that the user defined a workload to change that does not exist.

thedadams: Additionally, I like the idea of collecting the errors here into a slice and returning them all. Similar to how `checkMemory` does.

thedadams (changes requested):

pkg/apis/internal.acorn.io/v1/memory.go
```diff
@@ -52,6 +52,7 @@ var (
 	ErrInvalidAcornMemory   = errors.New("invalid memory from Acornfile")
 	ErrInvalidSetMemory     = errors.New("invalid memory set by user")
 	ErrInvalidDefaultMemory = errors.New("invalid memory default")
+	ErrInvalidWorkload      = errors.New("invalid workload name set by user")
```
StrongMonkey, line 55: Do we call containers as `workload` in acorn?

thedadams: Acorn containers and jobs are collectively referred to as workloads.

pkg/server/registry/apps/validator.go
tylerslaton: Discussed offline, but we'll go ahead and put this into an issue an track it as a follow-up.

pkg/server/registry/apps/validator.go
jsilverio22: the error msg is now @thedadams @StrongMonkey 
```
 ERROR:  App.api.acorn.io "winter-lake" is invalid: spec.memory: Invalid value: "ng": workload name set by user does not exist
```

jsilverio22: > Additionally, I like the idea of collecting the errors here into a slice and returning them all. Similar to how `checkMemory` does.

It does to propagated up through `checkMemory` error  slice

StrongMonkey (approved):

pkg/server/registry/apps/validator.go
jsilverio22: > Additionally, I like the idea of collecting the errors here into a slice and returning them all. Similar to how `checkMemory` does.

should be handled now for multiple errs

PR #1102, opened 2023-01-23
Add option to delete all images in acron image rm command

aniki499:
Hello,

This PR adds the ability to delete all images by passing the -a option when calling
`acorn images rm`.

I think so far you could only delete images one by one (or purge everything).
To test this simply create a couple of images and run
`acorn images rm -a -f`
which should delete all images one by one.
Hope this is useful.

Regards
Richard

### DIFF
--- a/docs/docs/100-reference/01-command-line/acorn_image_rm.md
+++ b/docs/docs/100-reference/01-command-line/acorn_image_rm.md
@@ -18,6 +18,7 @@ acorn image rm my-image
 ### Options
 
 ```
+  -a, --all     Delete all images
   -f, --force   Force Delete
   -h, --help    help for rm
 ```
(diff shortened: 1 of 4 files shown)


cjellick:
@jsilverio22 and I discussed this recently and I felt that adding this was potentially a little too destructive.

This accomplishes the same thing:
```
acorn images rm -f $(acorn images -qa)
```
while also forcing the user to jump through an extra hoop, which in this case might be desirable in order to prevent an accidental delete all. 

I don't know, what do you think @ibuildthecloud ?

aniki499:
What about maybe adding a confirmation by the user?
So whenever the command is used, it would first display a message like „Caution: This will delete all existing images. Are you sure you want to continue? (Y/N)“

That would eliminate the possibility of someone deleting everything by mistake.

ibuildthecloud:
I just don't like the UX of this.  I'd much rather see a globbing syntax.  So support * and ?.  So the user can do `acorn image rm '*'`  Obviously shell quoting is a bit of a problem, but so be it.

```
acorn i rm '*'
acorn i rm 'foo*bar'
acorn i rm 'one?character'
```

cjellick:
sorry this never went anywhere and we couldnt agree on ux. closing

Issue #505, opened 2022-08-09
Support pre-canned cluster roles


cjellick:
We should have sensible default cluster roles for use with acorn.

There's the obvious:
- acorn-admin (maybe?)
- acorn-edit
- acorn-view

But we might want roles that make sense for environments where you want to disallow building. Like, most prod environments should not allow building acorn images, only running existing ones, so we should have a role for that.

cjellick:
Darren did this:
acorn:project:admin
acorn:project:build
acorn:project:edit
acorn:project:view

sangee2004:
Tested with latest build from master Jan/17 - `acorn version v0.5.0-rc2-2-g6cee142+6cee1422`

Following cluster roles are available on acorn installs:
```
acorn:project:admin                                                    
acorn:project:build                                                    
acorn:project:edit                                                     
acorn:project:view
```

PR #1107, opened 2023-01-23
Add a note about cluster roles for RBAC

thedadams:
Related issue: https://github.com/acorn-io/acorn/issues/505


tylerslaton (approved):
Thanks for theses change @thedadams 🤝

Issue #1075, opened 2023-01-20
Should not be allowed to pull images from internal-registry-prefix setting.


sangee2004:
acorn version v0.5.0-rc3-9-g4e16d33+4e16d334

Steps to reproduce the problem:

1. Have internal-registry-prefix  set to a registry - in my case dockerhub - `docker.io/sangeetha/`
2. Build image using acorn command `acorn build -t `
3. This will result in images being pushed to  internal-registry-prefix  and will be made available as <registry>:acorn:SHA (where SHA can be got from image id - `acorn images --no-trunc -a`) 

We are not allowed to pull these images when using the format - docker.io/xxx/:acorn:SHA.
But  we are allowed to pull these images when using `index.docker.io` or `registry-1.docker.io` as registry . 

Expected Behavior:

pull should not be allowed when using `index.docker.io` or `registry-1.docker.io` as well.

cjellick:
Can you make the steps to repro this a little more explicit? I don't fully follow.

sangee2004:
acorn version v0.5.0-rc3-50-g9a379e0+9a379e0a

This issue is still reproducible:

Steps to reproduce the problem:

1. Have internal-registry-prefix set to a registry - in my case dockerhub - `docker.io/sangeetha/`
2. Build image using acorn command `acorn build -t` 
3. This will result in images being pushed to docker hub and is made available as docker.io/sangeetha/<project-name>:<SHA> (where SHA can be got from image id - `acorn images --no-trunc -a` for the image built in step2)


We are not allowed to pull these images when using the format - docker.io/xxx/<project-name>:SHA.
But we are allowed to pull these images when using `index.docker.io` or `registry-1.docker.io` as registry .

```
sangeethahariharan@sangeethas-mbp example1 % acorn pull docker.io/sangeetha/mynewtest:<hash>
  ✗  ERROR:  websocket: close 1011 (internal server error): [id image reference prefix docker.io/sangeetha/mynewtest:89337c7445dd748a1affbe4b3ff65720bffd71dc5c9b4da32bef19f10c7af7b9]
sangeethahariharan@sangeethas-mbp example1 % acorn pull index.docker.io/sangeetha/mynewtest:<hash>
 [110887065/110887065] ███████████████████████████████████████████████ 100% | 0s

sangeethahariharan@sangeethas-mbp example1 % acorn pull registry-1.docker.io/sangeetha/mynewtest:<hash>
[...]

PR #1114, opened 2023-01-24
Normalize the internal registry check

ibuildthecloud:
If the internal registry prefix is docker.io and the user puts registry-1.docker.io or index.docker.io these hosts should all be treated as the same value.
    

https://github.com/acorn-io/acorn/issues/1075


tylerslaton:
Looks like you have some lint failures.

tylerslaton (approved):

cjellick (approved):

PR #1119, opened 2023-01-24
add automated reporting for main release workflows

jsilverio22:


tylerslaton (commented):
This may start to get a little noisy so we may want to consider aggregating all of these together but for now I think this is fine given the infrequency that we release.

tylerslaton (approved):

StrongMonkey:
I think this will be a little bit noisy. Can we just do

1. Notify everything in release channel
2. Only notify failure in main and smoke test.

Curious what other people think

StrongMonkey (approved):

PR #1120, opened 2023-01-24
Set --kubeconfig to env

ibuildthecloud:
If user has passed a --kubeconfig setting set it to the environment
variable to influence all calls to restconfig.Default()


cjellick:
https://github.com/acorn-io/acorn/issues/1121

Issue #1127, opened 2023-01-25
Add an allow list for user-defined labels and annotations


thedadams:
A previous change added the `--ignore-user-labels-and-annotations` installation flag that completely ignored user-defined labels and annotations.

We need an enhancement to this feature where `--allow-user-label` and `--allow-user-annotation` flags can be passed to `acorn install` to allow certain desired labels/annotations to be propagated to the Kubernetes objects.

sangee2004:
Tested with acorn version - acorn version `v0.5.0-8-g8809ad2+8809ad2a`

With `--ignore-user-labels-and-annotations` enabled and `--allow-user-label` and `--allow-user-annotation` flags  set ,  pods, secrets and volumes allow for  labels and annotations set in `--allow-user-label` and `--allow-user-annotation` field and other labels/annotations are silently ignored.

from `acorn info`:
```
allowUserAnnotations:
    - ann1
    - ann2
    allowUserLabels:
    - test1
    - test2
    ignoreUserLabelsAndAnnotations: true
```

When app is deployed with following labels and annotations - `acorn run --label=test1/hello2=value3,test1=value1,hello1=hhh --annotation=ann1.hyu=hello2,ann1=testme.new,ann2=pp -n testnew .`

Only the following labels and annotations get set on  pods, secrets and volumes  created for the app.

```
Labels:
               test1=value1
Annotations:   ann1: testme.new
               ann2: pp
```

PR #1131, opened 2023-01-25
Support allow-list for user labels and annotations

thedadams:
A previous change made it possible to disable propagation of user-defined labels and annotations. This turned out to be too restrictive, and we need a way to allow only certain labels and annotations.

This change adds two installation flags: allow-user-labels and allow-user-annotations. If values are supplied for these flags and the ignore-user-labels-and-annotations is true, then only the allowed labels and annotations are propagated. If ignore-user-labels-and-annotations is not set or is false, then these new flags have no effect.

Issue:
https://github.com/acorn-io/acorn/issues/1127


tylerslaton:
Looks like we just need to add these new flags to the CLI info test's expected output to get CI green.

https://github.com/acorn-io/acorn/tree/main/pkg/cli/testdata/info

tylerslaton (approved):

jsilverio22 (approved):

PR #1134, opened 2023-01-26
Only notify about successful tagged releases to Acorn Users

tylerslaton:
Once this merges, we will have to add two new secrets:

1. `USERS_SLACK_BOT_RELEASE_CHANNEL`
2. `USERS_SLACK_BOT_TOKEN`

In addition we will need to invite the bot to the channel we want it to notify to. I have already created the bot in the `Acorn Users` slack so this will be easy.


jsilverio22 (approved):

cjellick:
I get that youre removing succes reports to the ci-reports channel, but what is the other thing you're doing in release.yaml? It's a modification, so where is it reporting now that you'd be dropping?

tylerslaton:
@cjellick I should change the secret names to be clearer in their intent. That being said, the change in `release.yaml ` will make it so all successfully tagged releases get announced in the public Acorn Users slack. Before this change, successful tagged releases announced in our internal slack.

jacobdonenfeld (approved):
Cool! Nice that it links directly to the release once it is available.

cjellick (approved):

PR #1137, opened 2023-01-26
Add label and annotation propagation from project

StrongMonkey:

## Context

We need the ability to propagate annotations and labels from project namespace to appNamespace. The primary goal to drive this is to allow linkerd service mesh to inject sidecar to acorn app by default. This is done by propagating annotations `linkerd.io/inject: enabled` from project to all acorn app namespaces. 

Acorn linkerd plugin will be responsible for add `linkerd.io/inject: enabled` to all acorn projects, and acorn will just propagate the labels down to app namespaces.

## Usage 

To enable propagations, run acorn install with

```bash
acorn install --annotations-propagation "linkerd.io/inject"
```

Then install linkerd plugin

```bash
acorn run ghcr.io/acorn-io/acorn-linkerd-plugin:main
```

## Changes

It addes a seperate controller for propagating annotations and labels, and also corresponding tests for it.


pkg/apis/api.acorn.io/v1/types.go
```diff
@@ -336,6 +336,8 @@ type Config struct {
 	WorkloadMemoryDefault          *int64         `json:"workloadMemoryDefault" name:"workload-memory-default" quantity:"true" usage:"Set the default memory for acorn workloads. Accepts binary suffixes (Ki, Mi, Gi, etc) and \".\" and \"_\" seperators (default 0)" short:"m"`
 	WorkloadMemoryMaximum          *int64         `json:"workloadMemoryMaximum" name:"workload-memory-maximum" quantity:"true" usage:"Set the maximum memory for acorn workloads. Accepts binary suffixes (Ki, Mi, Gi, etc) and \".\" and \"_\" seperators (default 0)"`
 	UseCustomCABundle              *bool          `json:"useCustomCABundle" name:"use-custom-ca-bundle" usage:"Use CA bundle for admin supplied secret for all acorn control plane components. Defaults to false."`
+	AnnotationsPropagation         []string       `json:"annotationsPropagation" name:"annotations-propagation" usage:"The list of keys of annotations to propagate from acorn project to app namespaces"`
```
thedadams, line 339: Is this PR doing the same thing? https://github.com/acorn-io/acorn/pull/1131

StrongMonkey: Probablt different, I assume your PR is for populating labels and annotations from appInstance to k8s resources. While this is for populating labels from project namespace to app namespaces.

pkg/controller/namespace/propagation.go
```diff
@@ -0,0 +1,56 @@
...
+		return err
+	}
+
+	cfg, err := config.Get(req.Ctx, req.Client)
+	if err != nil {
+		return err
+	}
+
+	for _, appNs := range appNamespaces.Items {
+		if len(cfg.AnnotationsPropagation) > 0 {
```
thedadams, line 29: This if-check is not necessary. Same below.

pkg/controller/namespace/propagation.go
```diff
@@ -0,0 +1,56 @@
...
+			for _, key := range cfg.AnnotationsPropagation {
+				if v, ok := projectNamespace.Annotations[key]; ok {
+					if appNs.Annotations == nil {
```
thedadams, line 32: nit: do this before the loop so it only has to happen once.

thedadams (approved):
A couple small issues. Looks good.

jsilverio22 (approved):
lgtm

thedadams (approved):

StrongMonkey:
Darren has asked some requests so I made the following modification:

1. Move the logic to AddNamespace so that the change is owned by appInstance, not a separate handler that modifies namespaces annotations and labels.
2. Move AddNamespace into a separate handler to make code clean. Also move the test to accommodate.

StrongMonkey:
End up doing a large test refactoring. Since we want to move namespace out of DeploySpec, a lot of existing tests won't work since we no longer create namespace in deploySpec.

I have removed namespace.yaml in the expected.yaml, at the same time, created new tests specifically for labeling logic(mostly around label and annotation scopes) to ensure we have the same test coverage as before.

thedadams (changes requested):

pkg/controller/appdefinition/namespace.go
```diff
@@ -82,3 +85,53 @@ func RequireNamespace(h router.Handler) router.Handler {
...
+		return err
+	}
+
+	labelMap := map[string]string{
+		labels.AcornAppName:      appInstance.Name,
+		labels.AcornAppNamespace: appInstance.Namespace,
+		labels.AcornManaged:      "true",
+	}
+
+	labelMap = labels.Merge(labelMap, labels.GatherScoped("", "", appInstance.Status.AppSpec.Labels, nil, appInstance.Spec.Labels))
```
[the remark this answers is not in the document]
StrongMonkey: I feel like we should override. I think the configuration and labeling is mostly configured by admin(admin who install with the propagation flags) so it is expected that it should override other user config. But we can bring this to standup for discussion.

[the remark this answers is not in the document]
thedadams: The user can add labels to `appInstance.Status.AppSpec.Labels` and `appInstance.Spec.Labels` when they build the Acorn. You're then adding these labels to the namespace.

I agree that the labels and annotations you are trying to propagate should be added, but I don't think these should be if `--ignore-user-labels-and-annotations` is true.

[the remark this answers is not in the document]
StrongMonkey: Oh shoot, you are right I think we should do this. I thought this is already done but the protection was removed after my refactoring.

pkg/controller/routes.go
```diff
@@ -35,6 +35,7 @@ func routes(router *router.Router, registryTransport http.RoundTripper) {
 	router.HandleFunc(&v1.AppInstance{}, appdefinition.PullAppImage(registryTransport))
 	router.HandleFunc(&v1.AppInstance{}, appdefinition.ParseAppImage)
 	router.HandleFunc(&v1.AppInstance{}, tls.ProvisionCerts) // Provision TLS certificates for port bindings with user-defined (valid) domains
+	router.HandleFunc(&v1.AppInstance{}, appdefinition.AddNamespace)
```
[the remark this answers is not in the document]
StrongMonkey: Thanks for catching this! 

It is already addressed by tests here 

https://github.com/acorn-io/acorn/blob/b00dc8caad372f184116414fcf1c1fbd3941bdb3/pkg/controller/appdefinition/testdata/deployspec/no-user-labels-namespace/input.yaml#L18 

https://github.com/acorn-io/acorn/blob/b00dc8caad372f184116414fcf1c1fbd3941bdb3/pkg/controller/appdefinition/deploy_test.go#L51

But because how we call the handler code in production and test different, the test won't be catch it. It is a lesson to learn.

pkg/controller/appdefinition/namespace.go
```diff
@@ -82,3 +85,53 @@ func RequireNamespace(h router.Handler) router.Handler {
 		return h.Handle(req, resp)
 	})
 }
+
+func AddNamespace(req router.Request, resp router.Response) error {
```
ibuildthecloud, line 89: Does this function have to stay in appdefinition package?  Can it be moved out to controller/namespace.  I'd like to minimize the amount of code that we dump in appdefinition package.

pkg/apis/api.acorn.io/v1/types.go
ibuildthecloud, line 339: Can we change this to `PropagateNamespaceLabels` and `...Annotations`.  Also the struct key and json field should be plural but the CLI arg singular, follow ClusterDomains as an example (tldr; you need "name" annotation)

cjellick: in slack, i suggested `PropagateProjectLabels`, instead of the word namespace. does that make sense @ibuildthecloud ?

pkg/controller/appdefinition/namespace.go
StrongMonkey: I can move it

pkg/apis/api.acorn.io/v1/types.go
ibuildthecloud: The reason that I didn't want "project" was that it's not actually the project labels, but labels on the namespace for the project.  Confusing.... But this points out we have an issue in that these labels should not go on the v1.Project object and the use shouldn't be able to change them.

Can we go with PropagateProjectNamespaceLabels then?

pkg/controller/routes.go
```diff
@@ -28,13 +28,14 @@ var (
...
 
-func routes(router *router.Router, registryTransport http.RoundTripper) {
+func routes(router *routeUtils.Router, registryTransport http.RoundTripper) {
 	router.OnErrorHandler = appdefinition.OnError
 
 	router.HandleFunc(&v1.AppInstance{}, appdefinition.AssignNamespace)
 	router.HandleFunc(&v1.AppInstance{}, appdefinition.PullAppImage(registryTransport))
 	router.HandleFunc(&v1.AppInstance{}, appdefinition.ParseAppImage)
 	router.HandleFunc(&v1.AppInstance{}, tls.ProvisionCerts) // Provision TLS certificates for port bindings with user-defined (valid) domains
+	router.HandleFunc(&v1.AppInstance{}, appdefinition.FilterLabelsAndAnnotationsConfig(routeUtils.HandlerFunc(namespace.AddNamespace)).Handle)
```
thedadams, line 38: nit: this makes it obvious that we are adding a middleware. Feel free to ignore if you disagree.
```suggestion
	router.Type(&v1.AppInstance{}).Middleware(appdefinition.FilterLabelsAndAnnotationsConfig).HandlerFunc(routeUtils.HandlerFunc(namespace.AddNamespace))
```

thedadams (approved):

pkg/controller/routes.go
StrongMonkey: Yeah it is fine i can change it.

Originally I wrote it in this way but change it since the test code has to be written in this way in order to test.(the test framework doesn't support middleware so it is easy to miss out).

I guess we just need to be careful to include middleware in testing code.

StrongMonkey:
Addressed all darren's comments and donnie's too.

The current flag:

```
--propagate-project-annotation strings   The list of keys of annotations to propagate from acorn project to app namespaces
--propagate-project-label strings        The list of keys of labels to propagate from acorn project to app namespaces
```

Going to merge it.

cjellick:
@StrongMonkey please open an issue for this and put it into the v0.6.0 milestone, in the QA column

StrongMonkey:
Issue: https://github.com/acorn-io/acorn/issues/1172

Issue #1061, opened 2023-01-18
Not prompted for email and accepting TOS when enabling lets-encrypt 


sangee2004:
acorn version v0.5.0-rc2-2-g6cee142+6cee1422

```
acorn install --lets-encrypt enabled
 ✗ ERROR:  letsencrypt email is required when Let's Encrypt is enabled

```

sangee2004:
acorn version v0.5.0-5-g59796db+59796dbe

email address validation fails
```
 % acorn install --lets-encrypt enabled
? You are choosing to enable Let's Encrypt for TLS certificates. To do so, you must agree to their Terms of Service: https://letsencrypt.org/documents/LE-SA-v1.3-September-21-2022.pdf
Tip: use --lets-encrypt-tos-agree to skip this prompt
Do you agree to Let's Encrypt TOS? Yes
Enter your email address for Let's Encrypt: sangeetha@acorn.io
  •  You've enabled automatic TLS certificate provisioning with Let's Encrypt. This can take a few minutes to configure.
  ✗  ERROR:  invalid email address 'sangeetha@acorn.io' provided for Let's Encrypt
```

sangee2004:
Tested with acorn version - v0.5.0-14-g20c6273+20c62732

We are prompted for email and TOS when enabling lets-encrypt and able to enable successfully.

PR #1141, opened 2023-01-27
validate email from prompt (#1061)

jsilverio22:
This PR fixes email logic using the finalConf or opts for email when validating
Issue: #1061

### DIFF
--- a/pkg/install/install.go
+++ b/pkg/install/install.go
@@ -136,10 +136,13 @@ func Install(ctx context.Context, image string, opts *Options) error {
 	}
 
 	// Validate E-Mail address provided for Let's Encrypt registration
-	if finalConfForValidation.LetsEncryptEmail != "" || *finalConfForValidation.LetsEncrypt == "enabled" {
-		email := finalConfForValidation.LetsEncryptEmail
+	email := opts.Config.LetsEncryptEmail
+	if finalConfForValidation.LetsEncryptEmail != "" {
+		email = finalConfForValidation.LetsEncryptEmail
+	}
+	if email != "" || *finalConfForValidation.LetsEncrypt == "enabled" {
 		if !validMailAddress(email) {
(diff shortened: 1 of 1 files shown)


pkg/install/install.go
```diff
@@ -136,10 +136,15 @@ func Install(ctx context.Context, image string, opts *Options) error {
...
 
 	// Validate E-Mail address provided for Let's Encrypt registration
-	if finalConfForValidation.LetsEncryptEmail != "" || *finalConfForValidation.LetsEncrypt == "enabled" {
-		email := finalConfForValidation.LetsEncryptEmail
+	email := ""
+	if finalConfForValidation.LetsEncryptEmail != "" {
+		email = finalConfForValidation.LetsEncryptEmail
+	} else {
+		email = opts.Config.LetsEncryptEmail
+	}
```
thedadams, line 144: nit:
```suggestion
	email := opts.Config.LetsEncryptEmail
	if finalConfForValidation.LetsEncryptEmail != "" {
		email = finalConfForValidation.LetsEncryptEmail
	}
```

pkg/install/install.go
```diff
@@ -136,10 +136,15 @@ func Install(ctx context.Context, image string, opts *Options) error {
...
+	if email != "" || *finalConfForValidation.LetsEncrypt == "enabled" {
```
thedadams, line 145: Is it valid to have an empty email address here?

If no, then we can remove the `email != ""` check here.

jsilverio22: I believe this is in there to handle when lets encrypt is disabled but a email is provided we still want to validate it.

thedadams: Got it. Makes sense. Thanks @jsilverio22!

thedadams (approved):

tylerslaton (approved):

PR #1157, opened 2023-01-29
fixes #1112: show error on app stop when not found

### LINKED ISSUE #1112: When invalid app name is passed to`acorn stop` , there is no error message presented to the user.
acorn version v0.5.0-rc3-40-gc138b50+c138b504

When invalid app name is passed to`acorn stop` , there is no error message presented to the user. 

```
sangeethahariharan@sangeethas-mbp example1 % acorn stop abc        
abc
sangeethahariharan@sangeethas-mbp example1 % acorn start abc
  ✗  ERROR:  starting abc: appinstances.internal.acorn.io "abc" not found
```
User should be presented with an error message similar to the one presented when attempting `acorn start` with invalid  app name.

zostay:
Signed-off-by: Sterling Hanenkamp <sterling@hanenkamp.com>

Just removes the special case for ignore the erorr if the app `IsNotFound`

### DIFF
--- a/pkg/client/app.go
+++ b/pkg/client/app.go
@@ -433,9 +433,7 @@ func (c *DefaultClient) appStop(ctx context.Context, name string) error {
 		Name:      name,
 		Namespace: c.Namespace,
 	}, app)
-	if apierrors.IsNotFound(err) {
-		return nil
-	} else if err != nil {
+	if err != nil {
 		return err
 	}
 	if app.Spec.Stop == nil || !*app.Spec.Stop {


cjellick:
Thanks for this. Will get reviews on this today.

cjellick:
@zostay we don't use/like the auto-close feature of github that comes from saying fixes #issue-number.
Once a fix is merged, we'll send the issue on to QAing.

cjellick (approved):

cjellick:
This change would bring the behavior in line with how `docker stop` behaves, which i think is the right thing

tylerslaton (approved):
Nice job @zostay! Looks good to me and I agree with @cjellick here.

jsilverio22 (approved):
looks to be working as docker stop does on my end

Issue #1092, opened 2023-01-21
No check to ensure that `workload-memory-maximum` cannot be lower than `workload-memory-default`


sangee2004:
acorn version v0.5.0-rc3-18-gfbe46a2+fbe46a25

Currently there is no check to ensure that `workload-memory-maximum` cannot be lower than `workload-memory-default`.

User is allowed to set  `workload-memory-maximum` <`workload-memory-default` and set `workload-memory-default` > `workload-memory-maximum` . acorn install should error out in this case with proper error message.

sangee2004:
Tested with acorn version v0.5.1-28-g099c492+099c492f

We are not allowed to set memory params that will result in  `workload-memory-maximum` < `workload-memory-default`

```
acorn install --workload-memory-default=290Mi
  ✗  ERROR:  invalid memory args: workload-memory-default set to 290Mi which exceeds the workload-memory-maximum of 256Mi

 acorn install --workload-memory-maximum=100Mi                                 
  ✗  ERROR:  invalid memory args: workload-memory-default set to 200Mi which exceeds the workload-memory-maximum of 100Mi

```

PR #1164, opened 2023-01-30
validate memory args before installing (#1092)

jsilverio22:
This PR addresses not checking for memory flags at install time. 
Issue: #1092 

### DIFF
--- a/pkg/install/install.go
+++ b/pkg/install/install.go
@@ -37,6 +37,7 @@ import (
 	networkingv1 "k8s.io/api/networking/v1"
 	apierror "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -151,6 +152,10 @@ func Install(ctx context.Context, image string, opts *Options) error {
 		return err
 	}
 
+	if err = validateMemoryArgs(*finalConfForValidation.WorkloadMemoryDefault, *finalConfForValidation.WorkloadMemoryMaximum); err != nil {
+		return err
+	}
+
 	opts = opts.complete()
 	if opts.OutputFormat != "" {
 		return printObject(image, opts)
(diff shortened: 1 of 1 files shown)


thedadams (approved):

pkg/install/install.go
```diff
@@ -253,6 +257,21 @@ func Install(ctx context.Context, image string, opts *Options) error {
 	return nil
 }
 
+func validateMemoryArgs(defaultMemory int64, maximumMemory int64) error {
+
+	//if default is set to infinite memory (0) max memory must also be set to (0)
```
tylerslaton, line 262: nit:
```suggestion
	// if default is set to unlimited memory (0) max memory must also be set to (0)
```

pkg/install/install.go
```diff
@@ -253,6 +257,21 @@ func Install(ctx context.Context, image string, opts *Options) error {
...
+	if defaultMemory == 0 && maximumMemory != 0 {
+		return fmt.Errorf("invalid memory args: workload-memory-default set to unlimited which exceeds the workload-memory-maximum of %v",
+			maximumMemory)
+	}
```
tylerslaton, line 266: Right now in the `run` and `Acornfile` if you specify `0` for their memory when the max is not `0`, they will instead be given the maximum memory. I'd say we should also follow that behavior here so that a user can say "here is my maximum memory and for the default I want the workloads to get as much as they can". 

Even better, in a default installation a user can specify just the `--workload-memory-maximum` flag and all workloads will default to that maximum since the `--workload-memory-default` flag is `.`. What if we instead added some output here to just let a user know what's happening?

pkg/install/install.go
```diff
@@ -253,6 +257,21 @@ func Install(ctx context.Context, image string, opts *Options) error {
...
+	//if max memory is not set to infinite default must be smaller than maximum
```
tylerslaton, line 267: nit:
```suggestion
	// if max memory is not set to unlimited default must be smaller than maximum
```

tylerslaton (changes requested):

pkg/install/install.go
jsilverio22: so if max memory != 0 and default = 0 we currently set default = max?

tylerslaton: In effect, that logic happens here:

https://github.com/acorn-io/acorn/blob/6c93a9ed787abe927b3928f7db6895832e26924d/pkg/apis/internal.acorn.io/v1/memory.go#L115-L118

tylerslaton: So if we get to that line and the amount of bytes that we would set would be equal to zero (say in the case that we're using the default) then we will get the maximum amount of memory instead.

jsilverio22: pushed up a change let me know if that seems better ux

tylerslaton: Two things here. 

1. Right now testing locally I realized that we're outputting the full byte amount to the user in the error message. Could you update this to parse it and then turn it into a suffixed string like we do elsewhere in the code base?
    ```
    ✗  ERROR:  invalid memory args: workload-memory-default set to 1048576000 which exceeds the workload-memory-maximum of 524288000
    ```

2. The warning right now is a bit misleading - there's nothing actually wrong, just a bit magical so communicating it is nice. Can you update this to use `pterm.Info.Println()` and remove the warning part? I'm envisioning this to be a similar UX to how LE communicates info [here](https://github.com/acorn-io/acorn/blob/6fa2978f51e27e98dd14dd6c98e3e838cd649640/pkg/install/install.go#L135).

pkg/install/install.go
```diff
@@ -253,6 +258,23 @@ func Install(ctx context.Context, image string, opts *Options) error {
 	return nil
 }
 
+func validateMemoryArgs(defaultMemory int64, maximumMemory int64) error {
+	// if default is set to unlimited memory (0) and max memory is not default will be set to maximum
+	if defaultMemory == 0 && maximumMemory != 0 {
+		pterm.Info.Println("workload-memory-default is being set to workload-memory-maximum if this is not intended please specify workload-memory-default to non-zero value")
```
tylerslaton, line 264: nit
```suggestion
		pterm.Info.Println("workload-memory-default is being set to workload-memory-maximum. If this is not intended please specify workload-memory-default to non-zero value")
```

tylerslaton (approved):
Nice! Great work. One non-blocking comment but feel free to merge.

pkg/install/install.go
tylerslaton: Nice, most recent push looks good.

PR #1167, opened 2023-01-30
Add controller flag to run in dev-mode

thedadams:
Currently, the only thing that dev-mode does is set the leader election
TTL to 1 hour.

### DIFF
--- a/integration/helper/controller.go
+++ b/integration/helper/controller.go
@@ -210,7 +210,7 @@ func StartController(t *testing.T) {
 	}
 
 	lock(context.Background(), k8s, func(ctx context.Context) {
-		c, err := controller.New()
+		c, err := controller.New(true)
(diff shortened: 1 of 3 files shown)


StrongMonkey:
Hmm can we keep it and make the expiration extra long? I still like to have leader election enabled in dev so that I don't accidently run controller locally against dev clusters when switching kubeconfig :)

thedadams:
@StrongMonkey We absolutely can. I was just trying to figure out a way to make it easy and explicit. i.e. not use a `DEV` environment variable.

Should we do something like `acorn controller --dev-mode` which then sets the leader election TTL to 1 hour?

cjellick (approved):

tylerslaton (approved):

jsilverio22 (approved):

PR #1170, opened 2023-01-31
Check if string array's length greater than zero before overriding

StrongMonkey:

## Context

We noticed that in dev-cluster some setting was constantly being overriden. It turns out that when merging config values, we only check whether value is not nil before overriding, and CLI will always initialize the value as an empty string array. This will cause the field being overriden as nil if `acorn install` is rerun.

Instead, we should check the length of the field and only override when its length is greater than zero.(user has explictly set the values).

### DIFF
--- a/docs/docs/30-installation/02-options.md
+++ b/docs/docs/30-installation/02-options.md
@@ -92,3 +92,9 @@ Note that in order to allow propagation of user-defined labels and annotations o
 
 ## Changing install options
 If you want to change your installation options after the initial installation, just rerun `acorn install` with the new options. This will update the existing install dynamically.
+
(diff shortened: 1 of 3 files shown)


thedadams (approved):

cjellick:
Can you add tests and an issue?

tylerslaton:
+1 to what Craig said and I'll add that it would be nice to add documentation explaining the `""` reset behavior.

StrongMonkey:
Issue: https://github.com/acorn-io/acorn/issues/1173

StrongMonkey:
@cjellick @tylerslaton docs and tests have been added. 👍

tylerslaton (approved):

Issue #1052, opened 2023-01-17
Improved storage UX via a StrorageClass resource


cjellick:
Right now, to provision storage for an app, you need to know the semi-magic string for specifying the type of storage you want. We want to improve this by introducing a "storageClass" (need a better name) type resource.

This is similar to our "computeClass" concept in that the admin will be able to configure instances of the "class" and then the end user just needs to say "launch this acorning using the super-fast-ssd storage class"

thedadams:
There are a few things to test here:
1. When Acorn is installed with default settings, it should create cluster volume classes (volume classes available to all apps running in the cluster) for each Kubernetes storage class. The default storage class will have its corresponding volume class made default. These cluster volume classes don't have any restrictions on them; they are there to ensure the current behavior is maintained. A user can update these cluster volume classes with restrictions and those restriction will be adhered to.
2. When Acorn is installed with the `--manage-volume-classes` flag, no cluster volume classes are generated, and, if they exist, they are deleted. The admin user is responsible for creating and managing cluster and project volume classes. See the [docs](https://docs.acorn.io/next/reference/admin/volumeclasses) for the configuration of volume class objects.
3. An app should not be able to run if it is trying to use a volume class that does not exist, or it has a volume size that is outside the bounds of the definition of the volume class (i.e. its size is too small or too big). Additionally, an app should not be able to run if it has a volume that has accessModes that are not allowed by the volume class.
4. If an app is already running and a volume class is updated that would make the app invalid, it will continue to run without error until the app is updated.

One note about the different objects here. There are ClusterVolumeClasses (available to the whole cluster) ProjectVolumeClasses (available to a project) and VolumeClasses. The last is how a user will interact with the VolumeClasses. The user doesn't care about whether they are using a cluster or project volume class, they just want to know what is available to them. Therefore, commands like `kubectl -n acorn get volumeclasses.api.acorn.io` and `acorn offerings volumeclasses` will return all volumeclasses available to a user regardless of whether they are cluster or project.

sangee2004:
Tested with  `acorn version v0.5.1-82-g652e22e7+652e22e7`

volumeClass feature is available and workflows outlined in the above comments work as expected. Issues found when validating this feature is tracked as standalone issues:


https://github.com/acorn-io/acorn/issues/1257
https://github.com/acorn-io/acorn/issues/1259
https://github.com/acorn-io/acorn/issues/1261
https://github.com/acorn-io/acorn/issues/1262
[...]

Issue #425, opened 2022-08-02
Detect missing storage class and report it in volume status


ibuildthecloud:

thedadams:
Now, with the implementation of volume classes, a user will not be able to run an app if a volume class doesn't exist. Additionally, the status of an app will be updated if a volume class it is using is deleted.

sangee2004:
Tested with `acorn version v0.5.1-82-g652e22e7+652e22e7`

User is not allowed to deploy an app with volume class that does not exist.

If the volume class with which the app was deployed is deleted , status of the app provides the following status as soon as the volume class gets deleted.

```
mytest1   27a2966c428b   3         3            11m ago   http://app-mytest1-afb89c24.local.on-acorn.io => app:5000   [controller: failed to update mytest1-785ae223-2ab/pgdatanew /v1, Kind=PersistentVolumeClaim for acorn-controller p3/mytest1: PersistentVolumeClaim "pgdatanew" is invalid: spec: Forbidden: spec is immutable after creation except resources.requests for bound claims
  core.PersistentVolumeClaimSpec{
          ... // 2 identical fields
          Resources:        {Requests: {s"storage": {i: {...}, s: "10G", Format: "DecimalSI"}}},
          VolumeName:       "pvc-931218ac-7fc6-4175-9f9e-1ec1b832f1d5",
-         StorageClassName: &"local-path-test4",
+         StorageClassName: &"",
          VolumeMode:       &"Filesystem",
          DataSource:       nil,
          DataSourceRef:    nil,
  }
```

Issue #415, opened 2022-08-01
Add status to CLI for pending PVCs


ibuildthecloud:
We don't show any status on volume provisioning.  If volume provisioning is happening it just looks like a stuck container.

thedadams:
The changes for this have been merged. Now, if your apps has a volume, you will see pending status for it in the CLI status.

One limitation of this is that we can only show binding errors if the storage class puts information on the PV or PVC. Many of them do not.

sangee2004:
Tested with acorn version v0.5.1-82-g652e22e7+652e22e7

Deployed an app which uses a volume with "accessMode" set to "readWriteMany" which is not supported by the default volume class  that gets picked.

```
   "pgdatanew": {
      accessModes: "readWriteMany"
   }
```

acorn run command now gives information about waiting on volume creation:
STATUS: ENDPOINTS[http://app-mytest1-9c237273.local.on-acorn.io => app:5000] HEALTHY[1/2] UPTODATE[2] [containers: app pending create; db is not scheduled to a node] [volumes: waiting for volume pgdatanew to provision and bind]

There is no acorn volumes created in this case.

We do see a pvc getting created that is stuck in "pending" state in this case with the following warnings:

```
  Warning  ProvisioningFailed    48s (x4 over 2m33s)  rancher.io/local-path_local-path-provisioner-79f67d76f8-6bdgf_3e961074-d14b-465f-881e-53d8719974b8  failed to provision volume with StorageClass "local-path-test2": Only support ReadWriteOnce access mode
```
```
% acorn apps
NAME        IMAGE          HEALTHY   UP-TO-DATE   CREATED   ENDPOINTS                                                     MESSAGE
mytest1     b23df22fec1c   1/2       2            14m ago   http://app-mytest1-9c237273.local.on-acorn.io => app:5000     [containers: app pending create; db is not scheduled to a node] [volumes: waiting for volume pgdatanew to provision and bind]
```

With this fix, do we expect acorn volume to  be created in "Pending" state with the above message from PVC ?

Also , should we expect to see the error message from PVC about "Only support ReadWriteOnce access mode" be reflected in the app messages ?

thedadams:
Spoke about this issue "offline" and determined it should be closed.

PR #1176, opened 2023-02-01
Add VolumeClasses to the Acorn API (#1052)

thedadams:
VolumeClasses are an abstraction of Storage Classes in Kubernetes. An
admin can create ProjectVolumeClasses and ClusterVolumeClasses to
indicate which Storage Classes are available to users, volume size
constraints, access mode constraints, and defaults if a user omits
information. If a user attempts to use a Volume Class that does not
exist, an error is produced.

The defaults for an app are calculated upfront. These defaults will not
change unless the generation of the app is updated. This allows running
apps to continue to run after an admin updates their Acorn installation.

Finally, there is a new API for users for VolumeClasses. This API does
not expose which VolumeClasses are Cluster or Project scoped. If a user
were to use `acorn offerings volumeclasses`, then they would be shown
all classes available to them.

Lastly, some re-factorization is done here to make the API groups in the
api-server more manageable.

An installation flag has been added that specifies that an admin would
like to manually manage the volume classes. If this flag is enabled,
then all volume classes would have to be created by an admin before any
existing storage classes can be used.

If this flag is not passed, then a controller will sync all existing
storage classes (and the ephemeral class) to ClusterVolumeClasses so
that users can use these classes. The controller is implemented such
that the storageClassName, default, and description fields are managed
by the controller and cannot be changed by an admin. The admin can
change all other fields and the controller won't overwrite them.

A final note to reviewers: these changes contain a re-factorization as mentioned above. This was included here to help with potential merge conflicts to do changes in the same area in a different PR. It serves no other purpose.

Issues:
https://github.com/acorn-io/acorn/issues/1052
https://github.com/acorn-io/acorn/issues/415
https://github.com/acorn-io/acorn/issues/425


integration/run/run_test.go
```diff
@@ -64,11 +68,94 @@ func TestVolume(t *testing.T) {
...
+			obj.Labels[labels.AcornAppNamespace] == app.Namespace &&
+			obj.Labels[labels.AcornManaged] == "true" &&
+			obj.Labels[labels.AcornVolumeName] == "my-data"
+	})
+
+	helper.WaitForObject(t, helper.Watcher(t, c), &apiv1.AppList{}, app, func(obj *apiv1.App) bool {
+		done := obj.Status.Condition(v1.AppInstanceConditionParsed).Success &&
+			obj.Status.Condition(v1.AppInstanceConditionVolumes).Success
+		if done {
+			if err = kclient.Delete(ctx, &sc); err != nil {
```
thedadams, line 148: Is this OK to do? This is the only way I could figure out to test the scenario of "what happens when an app is using a storage class that is removed."

tylerslaton:
Could just be my local system but when I try to get `acorn run -v class=` to autocomplete nothing happens. But when I do `acorn run -v ` it autocompletes and suggests different Volume Classes.

integration/run/run_test.go
tylerslaton: The only thing that I can think of being wrong here is that you could run up against the context timeout if this takes to long to spin up. That isn't an inherent problem to this code though, more just waiting in general for our codebase. 

I say this is fine.

thedadams:
@tylerslaton Thanks!

The problem was that we need to return the whole completion (`class=ephemeral` for example) for the shell to complete it. This has been fixed.

pkg/volume/volume.go
```diff
@@ -0,0 +1,53 @@
...
+	volumeClassNames = append(volumeClassNames, v1.VolumeRequestTypeEphemeral)
+	for _, sc := range volumeClasses.Items {
+		volumeClassNames = append(volumeClassNames, sc.Name)
+	}
+
+	return volumeClassNames, nil
+}
+
+func CheckVolumeClasses(ctx context.Context, c client.Client, appInstanceSpec v1.AppInstanceSpec, appSpec *v1.AppSpec) *field.Error {
+	if len(appInstanceSpec.Volumes) == 0 && len(appSpec.Volumes) == 0 {
```
jsilverio22, line 32: the appinstancespec and appspec should always be consistent, there won't be a scenario where appInstanceSpec =1 volume and appSpec = 0 right?

jsilverio22 (approved):
approved, just had a question on one piece

pkg/volume/volume.go
jsilverio22: that makes sense why it is both have to be = 0 since there are two ways to define volumes, thank you

tylerslaton (approved):
Tested it a couple of ways and everything feels right from a UX perspective. API's are good, implementation is good and has tests. Only thing I'd say is worth mentioning is to add a small note to the volume documentation around VolumeClasses [here](https://docs.acorn.io/running/volumes).

thedadams:
The new changes are up, but I need to add an install flag to this. I will update when this is completely ready.

tylerslaton:
When running locally, I notice the following output when attempting to get a VolumeClass through the CLI.

```
dacorn offerings volumeclasses local-path 
  ✗  ERROR:  an empty namespace may not be set when a resource name is provided
```

thedadams:
@tylerslaton Thanks for finding that. It has been fixed now.

pkg/controller/appdefinition/defaults.go
```diff
@@ -0,0 +1,102 @@
+package appdefinition
```
tylerslaton, line 1: Should this be its own `Defaults` package like it is in the `WorkloadClass` PR or do you prefer it this way?

pkg/controller/appdefinition/testdata/volumes/named/existing.yaml
```diff
@@ -0,0 +1,12 @@
+---
+kind: ClusterVolumeClassInstance
+apiVersion: internal.admin.acorn.io/v1
+metadata:
+  name: custom-class
+description: Just a simple test volume class
+default: true
+storageClassName: custom-class
+minSize: 1Gi
```
tylerslaton, line 9: Just realizing it down here - Should we be looking at Gigabytes instead of Gibibytes for VolumeClasses since its storage?

pkg/apis/internal.admin.acorn.io/v1/volumes.go
```diff
@@ -0,0 +1,46 @@
...
+
+	StorageClassName   string         `json:"storageClassName"`
+	Description        string         `json:"description"`
+	Default            bool           `json:"default,omitempty"`
+	AllowedAccessModes v1.AccessModes `json:"allowedAccessModes,omitempty"`
+	DefaultSize        v1.Quantity    `json:"defaultSize,omitempty"`
+	MinSize            v1.Quantity    `json:"minSize,omitempty"`
+	MaxSize            v1.Quantity    `json:"maxSize,omitempty"`
+	Inactive           bool           `json:"inactive,omitempty"`
+}
```
tylerslaton, line 22: Nit-pick with a pretty large amount of changes implied so I won't block for it. Thoughts, though?

```suggestion
type ProjectVolumeClassInstance struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`

	StorageClassName   string         `json:"storageClassName"`
	Description        string         `json:"description"`
	Default            bool           `json:"default,omitempty"`
	AllowedAccessModes v1.AccessModes `json:"allowedAccessModes,omitempty"`
	Size               Size           `json:"size,omitempty"`
	Inactive           bool           `json:"inactive,omitempty"`
}

type VolumeSize struct {
	Default  v1.Quantity    `json:"default,omitempty"`
	Min      v1.Quantity    `json:"min,omitempty"`
	Max      v1.Quantity    `json:"max,omitempty"`
}
```

pkg/volume/volume.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/acorn-io/baaah/pkg/typed"
+	"golang.org/x/exp/slices"
+	storagev1 "k8s.io/api/storage/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/kubectl/pkg/util/storage"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func SyncVolumeClasses(req router.Request, resp router.Response) error {
```
tylerslaton, line 20: Why are we syncing all StorageClasses during every reconciliation? The request should have the specific StorageClass and we can just apply that one instead of syncing all of them, right?

`sc := req.Object.(*storage.StorageClass)`

Basically something like this.

tylerslaton (commented):
Looking really good. Most of my comments are asking about skew between this PR and the WorkloadClass PR. I ran through quite a few use cases locally and all of them were looking good.

pkg/controller/appdefinition/testdata/volumes/named/existing.yaml
thedadams: Either works, as long as the validation works out.

pkg/volume/volume.go
thedadams: I am doing it this way so that the `ephemeral` class gets created as expected. If we only handled the storage class in the request, then we would never create the ephemeral one.

However, you question raises the point: what happens if there are no storage classes? Then the ephemeral class will never get created. I will fix this.

pkg/controller/appdefinition/defaults.go
```diff
@@ -0,0 +1,102 @@
...
+	v1 "github.com/acorn-io/acorn/pkg/apis/internal.acorn.io/v1"
+	adminv1 "github.com/acorn-io/acorn/pkg/apis/internal.admin.acorn.io/v1"
+	"github.com/acorn-io/acorn/pkg/condition"
+	"github.com/acorn-io/acorn/pkg/volume"
+	"github.com/acorn-io/baaah/pkg/router"
+	"github.com/acorn-io/baaah/pkg/typed"
+	kclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func CheckStatus(h router.Handler) router.Handler {
```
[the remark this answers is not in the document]
thedadams: Yep! Done.

pkg/apis/internal.admin.acorn.io/v1/volumes.go
thedadams: Done.

pkg/controller/defaults/testdata/defaults/volume-class-two-cluster-defaults/expected.yaml
```diff
@@ -0,0 +1,28 @@
...
+      container-name:
+        image: "image-name"
+        dirs:
+          "/var/tmp":
+            volume: foo
+    volumes:
+      foo: {}
+  conditions:
+    - error: true
+      message: 'cannot establish defaults because two defaults volume classes exist: test-volume-class and test-volume-class-1'
```
tylerslaton, line 25: Will this be flakey? Since the order can't be guaranteed when listing this could be flipped which I observed locally.

jsilverio22 (approved):
One question on how to output. Code structure looks fine from my knowledge, also went through and did hands on review through ` kubectl apply ` testing different cases all looks good to me

pkg/cli/volume_classes.go
```diff
@@ -0,0 +1,62 @@
...
+	client ClientFactory
+}
+
+func (a *Storage) Run(cmd *cobra.Command, args []string) error {
+	c, err := a.client.CreateDefault()
+	if err != nil {
+		return err
+	}
+
+	out := table.NewWriter(tables.VolumeClass, a.Quiet, "")
```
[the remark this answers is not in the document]
thedadams: Yes, good call out! I updated this.

tylerslaton (approved):

PR #1193, opened 2023-02-03
Refactor acorn info and add -A functionality

jacobdonenfeld:
### Checklist
- [x] All relevant issues are referenced in this PR
- [x] The title of this PR would make a good line in the Release Note's Changelog
- [x] Commits are signed-off using `git commit -s`
- [x] Automated tests added to cover the changes. If tests couldn't be added, an explanation is provided in the Verification and Testing section.

#### Proposed Changes

Refactor acorn info and support acorn info -A

Lists info for the current default project, or for all projects.

Returns a struct infoCLIResponse, {Client, []ProjectInfo}

Info may be the same for multiple project's, and could be worth while simplifying in the future, however wasn't sure how to go about this while still using a map from Project Name to Info. The only unique item in each info per project is publicKeys.keyID

#### User-Facing Change? ####
acorn info and acorn info -A will have a slightly different response and behavior.

#### Verification and Testing ####

Refactored CLI tests so info_test.go uses gomock. This tests regular info response, no info response, and info -A response.

Test file for testing client and multiclient's implementation of info as well. _integration/client/info/info_test.go_

#### Linked Issues ####
#1197


pkg/client/multi.go
```diff
@@ -408,12 +409,37 @@ func (m *MultiClient) ProjectList(ctx context.Context) ([]apiv1.Project, error)
...
+	for _, project := range projects {
+		c, err := m.Factory.ForProject(ctx, project.Name)
+		if err != nil {
+			return nil, err
+		}
+		subprojectInfo, err = c.Info(ctx)
+		if err != nil {
+			return nil, err
+		}
+		if len(subprojectInfo) == 1 {
```
StrongMonkey, line 430: It would be cleaner to just do 

```
for _, info := range subprojectInfo {
  info.name = project.name
  projectsInfo = append(projectsInfo, info)
}
```

So if a project can't list info, I fee like it is better UX to users that we just silently ignore it instead of throwing an error.

pkg/cli/info_test.go
```diff
@@ -50,6 +50,26 @@ func TestInfo(t *testing.T) {
 			wantErr: false,
 			wantOut: "./testdata/info/info_test.txt",
 		},
+		{
+			// TODO: Write a test that adequately covers -A
```
StrongMonkey, line 54: You should be able to implement this to get test running 
https://github.com/acorn-io/acorn/blob/34eb2440a787ff78b0717bf1a479c678e75658d4/pkg/cli/testdata/MockClient.go#L614

pkg/mocks/Makefile
```diff
@@ -0,0 +1,6 @@
+.PHONY: mocks
```
cjellick, line 1: Do we need this here and in the top level Makefile? Seems simpler to maintain it in a single place

jacobdonenfeld: Good call. Will move to top level

cjellick:
Just so we don't lose track of the feedback: I ran this locally and `acorn info -A` was resulting in:
```
  ✗  ERROR:  invalid project name [acorn.io/cjellick/acorn/acorn.io/cjellick/blah]: must contain zero, one or two slashes [/]
```
this needs addressed before it can be merged. **I'd also like to see if you can add a test that triggers this bug.**

Root cause of this bug is the implementation of `func (m *MultiClient) Info()` and specifically it's calling of 
```projects, err := m.ProjectList(ctx)```
That does not cause the desired effect.

A bit of debugging and inspection indicated that we probably need to use the `aggregate()` function to call Info() for each `client`, similar to how the AppList (or any of the list functions) works:
```
func (m *MultiClient) AppList(ctx context.Context) ([]apiv1.App, error) {
	return aggregate(ctx, m.Factory, func(client Client) ([]apiv1.App, error) {
		return client.AppList(ctx)
	})
}
```
...but the problem is that the aggregate function is going to prepend the project name to the name of whatever resource is returned by the client call. In this case, the object being returned is the Info object and its name IS the project name, so we need a modified version of the aggregate function that doesn't prepend the project name. I suggested the following changes but they actually don't compile because of some funkiness with generics. If we just want to add a bool parameter directly to the `aggregate()` function , rather than trying to keep its signature unchanged, I think that would be fine.
```
diff --git a/pkg/client/multi.go b/pkg/client/multi.go
index b4eaf4d1..7333a24d 100644
--- a/pkg/client/multi.go
+++ b/pkg/client/multi.go
@@ -37,6 +37,10 @@ type ObjectPointer[T any] interface {
 }

 func aggregate[T any, V ObjectPointer[T]](ctx context.Context, factory ProjectClientFactory, cb func(client Client) ([]T, error)) ([]T, error) {
+	return aggregateNameOption(ctx, true, factory, cb)
+}
+
+func aggregateNameOption[T any, V ObjectPointer[T]](ctx context.Context, appendProjectName bool, factory ProjectClientFactory, cb func(client Client) ([]T, error)) ([]T, error) {
 	var result []T
 	clients, err := factory.List(ctx)
 	if err != nil {
@@ -50,7 +54,9 @@ func aggregate[T any, V ObjectPointer[T]](ctx context.Context, factory ProjectCl
 		for _, item := range items {
 			if client.GetProject() != factory.DefaultProject() {
 				p := (V)(&item)
-				p.SetName(client.GetProject() + "/" + p.GetName())
+				if appendProjectName {
+					p.SetName(client.GetProject() + "/" + p.GetName())
+				}
 			}
 			result = append(result, item)
 		}
@@ -409,12 +415,16 @@ func (m *MultiClient) ProjectList(ctx context.Context) ([]apiv1.Project, error)
 }

 func (m *MultiClient) Info(ctx context.Context) ([]apiv1.Info, error) {
-	var projectsInfo []apiv1.Info
+	return aggregateNameOption(ctx, false, m.Factory, func(client Client) ([]apiv1.Info, error) {
+		return client.Info(ctx)
+	})
+
 	// Can this be refactored into a call to aggregate?
 	projects, err := m.ProjectList(ctx)
 	if err != nil {
 		return nil, err
 	}
+
 	var subprojectInfo []apiv1.Info
 	// For each project after listing them all, call info on each and populate projectsInfo slice
 	for _, project := range projects {
```

**Review by cjellick — DISMISSED** [review]
See my latest comment. This will probably be finished while I'm out, so @thedadams and @tylerslaton please re-review this, no need to wait for me to re-review

File: pkg/client/client.go [hunk 4]
```diff
@@ -230,7 +232,7 @@ type Client interface {
 	ProjectCreate(ctx context.Context, name string) (*apiv1.Project, error)
 	ProjectDelete(ctx context.Context, name string) (*apiv1.Project, error)
 
-	Info(ctx context.Context) (*apiv1.Info, error)
+	Info(ctx context.Context) ([]apiv1.Info, error)
```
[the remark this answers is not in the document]
cjellick: Well, the problem is that in Hub server information can vary between projects. Right now, because of how we do upgrades, the server versions happen to be consistent, but we don't want to rely on that.

pkg/config/cliconfig.go
```diff
@@ -42,6 +42,7 @@ func (a *AuthConfig) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+//go:generate $GOPATH/bin/mockgen --build_flags=--mod=mod -destination=../mocks/mock_cliconfig.go -package=mocks github.com/acorn-io/acorn/pkg/config CLIConfig
```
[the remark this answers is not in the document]
jacobdonenfeld: Removed, and moved the other to generate.go. TY

pkg/cli/info.go
```diff
@@ -27,15 +27,12 @@ type Info struct {
 	client ClientFactory
 }
 
-type ClientServerVersion struct {
+type InfoCLIResponse struct {
 	Client struct {
 		Version bversion.Version  `json:"version,omitempty"`
 		CLI     *config.CLIConfig `json:"cli,omitempty"`
 	} `json:"client,omitempty"`
-	Server  apiv1.InfoSpec `json:"server,omitempty"`
```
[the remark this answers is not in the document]
jacobdonenfeld: Server contains some public-key key/value pair unique to the project.

pkg/client/app.go
```diff
@@ -63,6 +63,7 @@ func ToApp(namespace, image string, opts *AppRunOptions) *apiv1.App {
 	}
 }
 
+// TODO: Jacob, review how to use w project
```
thedadams, line 66: Remove comment.

pkg/client/multi.go
```diff
@@ -442,12 +448,18 @@ func (m *MultiClient) ProjectList(ctx context.Context) ([]apiv1.Project, error)
...
 }
 
-func (m *MultiClient) Info(ctx context.Context) (*apiv1.Info, error) {
-	c, err := m.Factory.ForProject(ctx, m.Factory.DefaultProject())
-	if err != nil {
-		return nil, err
-	}
-	return c.Info(ctx)
+func (m *MultiClient) Info(ctx context.Context) ([]apiv1.Info, error) {
+	//var projectsInfo []apiv1.Info
```
thedadams, line 452: Remove if not necessary.

pkg/client/multi.go
```diff
@@ -50,7 +54,9 @@ func aggregate[T any, V ObjectPointer[T]](ctx context.Context, factory ProjectCl
 		for i := range items {
 			if client.GetProject() != factory.DefaultProject() {
 				p := (V)(&items[i])
-				p.SetName(client.GetProject() + "/" + p.GetName())
+				if appendProjectName {
+					p.SetName(client.GetProject() + "/" + p.GetName())
+				}
```
thedadams, line 59: nit: combine the if-statments:
```
if client.GetProject() != factory.DefaultProject() && appendProjectName {
```

integration/client/info/info_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+	ctx := helper.GetCTX(t)
+
+	cliConfig, err := config.ReadCLIConfig()
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Hack to ignore config for acorn hub and only run against local
+	//cliConfig.Auths = nil
+	//cliConfig.HubServers = []string{"acorn.io"}
```
thedadams, line 75: Remove if not necessary.

integration/helper/namespace.go
```diff
@@ -10,26 +14,76 @@ import (
...
+func namedTempCreate(t *testing.T, client client.Client, name string, project bool, generate bool) *corev1.Namespace {
+	t.Helper()
+	namespaceLabels := map[string]string{
+		"test.acorn.io/namespace": "true",
+	}
+	if project {
+		namespaceLabels[labels.AcornProject] = "true"
+	}
+	// double check to match actual project creation spec.
+	// Might be fitting to the test
```
thedadams, line 35: This looks like a note to yourself. Remove if unnecessary.

integration/helper/namespace.go
```diff
@@ -10,26 +14,76 @@ import (
...
 		},
 		Spec:   corev1.NamespaceSpec{},
 		Status: corev1.NamespaceStatus{},
 	}
+
+	if generate {
+		ns.ObjectMeta.GenerateName = name
+	} else {
+		ns.ObjectMeta.Name = name
+		ns.ObjectMeta.Namespace = name
```
thedadams, line 48: Namespaces aren't namespaced. I think the api-server will just remove this on creation.

integration/helper/namespace.go
```diff
@@ -10,26 +14,76 @@ import (
...
+	}
+
 	ctx := GetCTX(t)
 	err := client.Create(ctx, &ns)
 	if err != nil {
+		if errors.IsAlreadyExists(err) {
+			// Namespace already exists.
+			// construct blank object pointer
+			fmt.Printf("Namespace %s already exists, skipping creation.\n", name)
```
thedadams, line 57: nit: I like using `t.Log` here instead of `fmt.Printf` because `t.Log` will only display logs if the test fails or if `-test.v` flag is passed.

thedadams (approved):
Looks good! Just a few extraneous comments to be removed.

integration/helper/namespace.go
```diff
@@ -10,26 +14,76 @@ import (
...
 		Status: corev1.NamespaceStatus{},
 	}
+
+	if generate {
+		ns.ObjectMeta.GenerateName = name
+	} else {
+		ns.ObjectMeta.Name = name
+		ns.ObjectMeta.Namespace = name
+	}
+
```
tylerslaton, line 50: What if you just took in `metav1.ObjectMeta` here and applied it to the namespace when you create it? That would remove the name, project and generate parameters as well as some of this logic. It also has the benefit of being flexible should we need to add different labels or annotations.

integration/helper/namespace.go
```diff
@@ -10,26 +14,76 @@ import (
...
 	ctx := GetCTX(t)
 	err := client.Create(ctx, &ns)
 	if err != nil {
+		if errors.IsAlreadyExists(err) {
+			// Namespace already exists.
+			// construct blank object pointer
+			fmt.Printf("Namespace %s already exists, skipping creation.\n", name)
+			if generate {
+				fmt.Printf("wanted to generate a new namespace with random characters. Deeper issue.")
```
tylerslaton, line 59: Could you make this comment a bit more descriptive?

pkg/client/client.go
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/acorn-io/acorn/pkg/streams"
 	"github.com/acorn-io/acorn/pkg/system"
 	"github.com/acorn-io/baaah/pkg/restconfig"
+	_ "github.com/golang/mock/mockgen/model"
```
tylerslaton, line 16: Why is this imported here? Is it so that package's init function runs?

tylerslaton (approved):
Looks good, I agree with everything that Donnie said and had a few extra comments.

pkg/client/client.go
jacobdonenfeld: good catch, removed

integration/helper/namespace.go
jacobdonenfeld: Great point, that cleaned it up a bit. Thank you!

integration/secrets/secret_test.go
```diff
@@ -139,14 +139,14 @@ func TestEncryptionEndToEnd(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	keyBytes, err := base64.RawURLEncoding.DecodeString(info.Spec.PublicKeys[0].KeyID)
+	keyBytes, err := base64.RawURLEncoding.DecodeString(info[0].Spec.PublicKeys[0].KeyID)
```
jsilverio22, line 142: will this always have something at info[0]?

integration/secrets/secret_test.go
```diff
@@ -139,14 +139,14 @@ func TestEncryptionEndToEnd(t *testing.T) {
...
 	if err != nil {
 		t.Fatal(err)
 	}
 
 	assert.Len(t, keyBytes, 32)
 
-	encData, err := nacl.MultipleKeyEncrypt(plainTextData, []string{info.Spec.PublicKeys[0].KeyID})
+	encData, err := nacl.MultipleKeyEncrypt(plainTextData, []string{info[0].Spec.PublicKeys[0].KeyID})
```
jsilverio22, line 149: same as above

jsilverio22 (approved):
couple quick questions other than that lgtm

integration/secrets/secret_test.go
jacobdonenfeld: Either something at [0], or nothing, depending on hub implemention. Right now it always has [0]

PR #1217, opened 2023-02-09
 Don't decrypted data when writing generated secret 

ibuildthecloud:
If the user puts an encrypted value a secret data field
in the Acornfile do not decrypt the value when creating the generated
secret in the app's source namespace. This causes an undesired side
effect  in that the user can now do "acorn secret reveal" to get the
clear text of an encrypted secret. This isn't a secret issue because
the users has already been granted "reveal", but we'd rather not
reveal the clear text value for a secret that was intentionally
encrypted by the user.

### Checklist
- [x] All relevant issues are referenced in this PR
- [x] Commits follow [contributing guidance](https://github.com/acorn-io/acorn/blob/main/CONTRIBUTING.md#commits).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	cuelang.org/go v0.4.3
 	github.com/AlecAivazis/survey/v2 v2.3.6
 	github.com/acorn-io/aml v0.0.0-20220717003025-bc8cb1214693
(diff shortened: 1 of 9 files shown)


ibuildthecloud:
https://github.com/acorn-io/hub/issues/216

Issue #1216, opened 2023-02-09
Not able to exec into side car containers


sangee2004:
Acorn version - acorn version v0.5.1-49-gdc65d88+dc65d881

Steps to reproduce the problem:

Deploy apps with containers that have sidecars.
```
containers: {
  nginx1: {
    image: "nginx"
    sidecars:{
           sidecar: {
    		image: "ubuntu"
                cmd: ["bash","-c", "ls && sleep 1h"]
	}
	}
}
  ubuntu1: {
    image: "ubuntu"
    cmd: ["bash","-c", "ls && sleep 1h"]
    sidecars:{
           sidecar: {
    		image: "nginx"
	}
	}
}
}
```
Once the app is launched successfully, exec into sidecar container fails when attempting to ` acorn exec <sidecar container >` or `acorn exec <app>`

```
 % acorn containers
NAME                                     APP       IMAGE     STATE     RESTARTCOUNT   CREATED     MESSAGE
sidem.nginx1-b66cdf8d-4587p              sidem     nginx     running   0              4m15s ago   
sidem.nginx1-b66cdf8d-4587p.sidecar      sidem     ubuntu    running   0              4m15s ago   
sidem.ubuntu1-764fd666f8-66gwj           sidem     ubuntu    running   0              4m15s ago   
sidem.ubuntu1-764fd666f8-66gwj.sidecar   sidem     nginx     running   0              4m15s ago   

% acorn exec sidem.nginx1-b66cdf8d-4587p.sidecar    
 ✗  ERROR:  appinstances.internal.acorn.io "nginx1-b66cdf8d-4587p" not found

 % acorn exec sidem.ubuntu1-764fd666f8-66gwj.sidecar
  ✗  ERROR:  appinstances.internal.acorn.io "ubuntu1-764fd666f8-66gwj" not found

 % acorn exec sidem                                 
? Choose a container: sidem.nginx1-b66cdf8d-4587p.sidecar (running 5m20s ago)
  ✗  ERROR:  appinstances.internal.acorn.io "nginx1-b66cdf8d-4587p" not found
sangeethahariharan@Sangeethas-MBP sidecar % 
```

Also not able to use the "-c" option to specify the side car container name in `acorn exec` command. We still get into the primary container and not the sidecar container in this case.

cjellick:
putting this into v0.7.0 in the hopes of getting @rmodpur's fix merged

sangee2004:
[...]

PR #1199, opened 2023-02-03
add project label to namespace when used within an app (#1143)

jsilverio22:
Signed-off-by: Joshua Silverio <joshua@acorn.io>

### Checklist
- [x] All relevant issues are referenced in this PR
- [ ] The title of this PR would make a good line in the Release Note's Changelog
- [x] Commits are signed-off using `git commit -s`
- [x] Automated tests added to cover the changes. If tests couldn't be added, an explanation is provided in the Verification and Testing section.

#### Proposed Changes
_Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request_
This PR allows users to use pre-created namespaces within the project ecosystem once an app is run within that namespace
#### User-Facing Change? ####
_Does this PR introduce a user-facing change in functionality, the API or the CLI?_
Users can now see the namespace as a project after app is ran
_How will this change affect upgrades?_

#### Verification and Testing ####

_How can the changes be verified? Please provide whatever additional information necessary to help verify the proposed changes._
```
kubectl create namespace test
acorn run -j test .
acorn project
//should list test as a project and can do all things with test as a normal project (filter, use, rm, etc)
```
_Are there automated tests to cover your changes?_
yes
#### Linked Issues ####
#1143 
_Link any related issues, pull-requests, or commit hashes that are relevant to this pull request. If you are opening a PR without a corresponding issue please consider creating one first, at https://github.com/acorn-io/acorn/issues . A functional example will greatly help QA with verifying/reproducing a bug or testing new features._

Issue #1211, opened 2023-02-08
Control plane components need tolerations and node selectors


cjellick:
To better control the deploy of the acorn control plane (api-server, controller, maybe builders), we need to add resource requests (RRs) and tolerations to them.

The RRs should be configurablem so right now I'm imagining more flags on `acorn install`. Default should probably be to have no resource requests, but double check that when implementing.

The tolerations can be hard coded. We just need these components to specifically tolerate an "acorn control plane" taint.

cjellick:
This issue is now just for tolerations and node selector. ResourceRequests moved out to its own issue

sangee2004:
Tested with acorn version - `acorn version v0.6.0-141-gc3e8f286+c3e8f286`

Tested on a 3 node cluster with 1 control node , 1 etcd node and 1 worker node:

```
~# kubectl describe nodes santestw1 | grep -i taints
Taints:             taints.acorn.io/workload:NoSchedule

:~# kubectl describe nodes santeste1 | grep -i taints
Taints:             node-role.kubernetes.io/controlplane=true:NoSchedule

~# kubectl describe nodes santestc1 | grep -i taints
Taints:             node-role.kubernetes.io/control-plane=true:NoSchedule
```

Installed acorn and deployed few apps:

Control  node has api and controller pods running :
```
  acorn-system                acorn-api-c7d4bdff4-49d7b               0 (0%)        0 (0%)      0 (0%)           0 (0%)         52m
  acorn-system                acorn-controller-5c8bd5ff55-xs6qq       0 (0%)        0 (0%)      0 (0%)           0 (0%)         52m
```

Worker node has build-kit, registry and user workloads (and jobs) running:
```

  acorn-image-system          buildkitd-b5d9dc9c5-k5sdh         0 (0%)        0 (0%)      0 (0%)           0 (0%)         46m
  acorn-image-system          registry-cc85696b4-crxd7          0 (0%)        0 (0%)      0 (0%)           0 (0%)         49m
  mytest1-0c6d6ebc-cc6        myapp-67d68c6ddc-zbrvp            0 (0%)        0 (0%)      0 (0%)           0 (0%)         44m
  mytest1-0c6d6ebc-cc6        myweb-7494c9f46-br7vp             0 (0%)        0 (0%)      0 (0%)           0 (0%)         44m
[...]

PR #1230, opened 2023-02-16
Add controlplane taints and node selector to acorn system components (#1231)

StrongMonkey:
### Checklist

# Context

We need to figure Acorn system components to tolerate controlplane taints and only run it under cp nodes. 

The changes are 

1. Acorn-controll and acorn api and acorn-exec-check will be tolerating `node-role.kubernetes.io/control-plane`
2. Buildkitd, registry and user containers and jobs will be tolerating `taints.acorn.io/workload`

#1231

### DIFF
--- a/pkg/controller/scheduling/scheduling.go
+++ b/pkg/controller/scheduling/scheduling.go
@@ -3,6 +3,7 @@ package scheduling
 import (
 	v1 "github.com/acorn-io/acorn/pkg/apis/internal.acorn.io/v1"
 	"github.com/acorn-io/acorn/pkg/condition"
+	tl "github.com/acorn-io/acorn/pkg/tolerations"
 
 	adminv1 "github.com/acorn-io/acorn/pkg/apis/internal.admin.acorn.io/v1"
 	"github.com/acorn-io/acorn/pkg/config"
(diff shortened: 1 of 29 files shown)


cjellick:
Also related: https://github.com/acorn-io/acorn/issues/1211

tylerslaton (commented):
Do we care about `acorn-check-exec` job that is run on install also being on the control plane nodes? Additionally, should the image system components be there as well?

StrongMonkey:
@tylerslaton Good question, we should add toleration to acorn-check-exec so that it can run on controlplane.

cjellick:
what about the builders (buildkit pod)? where should those run?

StrongMonkey:
>what about the builders (buildkit pod)? where should those run?

Discussed offline, but buildkit pod should be running in worker nodes

pkg/controller/scheduling/scheduling.go
```diff
@@ -69,6 +70,13 @@ func addScheduling(req router.Request, appInstance *v1.AppInstance, workloads ma
 			return err
 		}
 
+		// Add default toleration to taints.acorn.io/workload. This is so that when worker nodes are tainted
+		// with taints.acorn.io/workload, user app can still tolerate.
+		tolerations = append(tolerations, corev1.Toleration{
```
tylerslaton, line 75: TIL that you can append to a nil slice!

pkg/install/apiserver.yaml
```diff
@@ -67,6 +67,9 @@ spec:
           securityContext:
             runAsUser: 1000
       serviceAccountName: acorn-system
+      tolerations:
```
tylerslaton, line 70: Shouldn't we have a `Toleration` to allow for the control-plane taint and then an `Affinity` to attract the pods to those nodes?

StrongMonkey: Great question Tyler. The caveat is that for hub every node is tainted so acorn controlplane will always go to the controlplane node. 

But for open-source, the taint is not guaranteed so it can go to any node as long as it can toleration the taints.

tylerslaton: Ah, understood. Thanks!

tylerslaton (approved):

cjellick (approved):

Issue #1232, opened 2023-02-16
Not able to specify units for volume size in acorn file. 


sangee2004:
Acorn version v0.5.1-65-ge8a480d5+e8a480d5

Steps to reproduce the problem:
Specify volume size in acorn file like
```
    "pgdatanew": {
       size: 5G
       accessModes: "readWriteOnce"
       class: "local-path-test"
    }
```

This results in volume getting created with "10G" capacity which is default size.

If the size is set as "5G" using --volume option in `acorn run` , then the volume gets created with "5G" capacity as expected.

```acorn run --volume pgdatanew,size=20G -n <name>  .```

Also there is an issue with setting numeric value for size like size=5 . Expected behavior for this is to create a volume with 5G which is not happening.   The only numbers I have seen work so far are 99 amd 11 which translate to 99G and 11G capacity for volumes that get created.

sangee2004:
Tested with `acorn version v0.5.1-72-gbd27105a+bd27105a`

Able to create volumes successfully with size < 10 G by specifying the size in acorn file as number value (defaults to G) and by specifying units like G and M.

PR #1235, opened 2023-02-16
Allow volumes with sizes less than 10G in Acornfile (#1232)

thedadams:
If a volume is defined in an Acornfile with a size less than 10G, then the size would be overwritten by the parser to be 10G. This was because it mistakenly thought there was a volume binding with size 10G. After this change, volumes defined in an Acornfile with size less than 10G will not be accidentally overwritten.

Issue: https://github.com/acorn-io/acorn/issues/1232

### Checklist

### DIFF
--- a/integration/run/testdata/volume/Acornfile
+++ b/integration/run/testdata/volume/Acornfile
@@ -3,5 +3,12 @@ containers: {
 		build: "."
 		dirs: "/test": "./files"
 		dirs: "/external": "external"
+		dirs: "/tmp": "volume://my-data"
+	}
+}
+
+volumes: {
+	"my-data": {
+		size: "1G"
 	}
 }
--- a/pkg/apis/internal.acorn.io/v1/unmarshal.go
+++ b/pkg/apis/internal.acorn.io/v1/unmarshal.go
@@ -257,13 +257,15 @@ func impliedVolumesForContainer(app *AppSpec, containerName, sideCarName string,
 					// ignore error
 					continue
 				}
-				vSize, err := resource.ParseQuantity((string)(v.Size))
-				if err != nil {
-					// ignore error
-					continue
-				}
-				if existingSize.Cmp(vSize) < 0 {
-					existing.Size = v.Size
+				if v.Size != "" {
+					vSize, err := resource.ParseQuantity((string)(v.Size))
+					if err != nil {
+						// ignore error
+						continue
+					}
+					if existingSize.Cmp(vSize) < 0 {
+						existing.Size = v.Size
+					}
 				}
 				for _, accessMode := range v.AccessModes {
 					found := false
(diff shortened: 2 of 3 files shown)


integration/run/testdata/volume/Acornfile
```diff
@@ -3,5 +3,12 @@ containers: {
...
 		dirs: "/external": "external"
+		dirs: "/tmp": "volume://my-data"
 	}
 }
+
+volumes: {
+	"my-data": {
+		size: "1G"
+	}
+}
```
jacobdonenfeld, line 14: What does this default to when not defined? Worth writing a test for when this is not defined as well?

pkg/apis/internal.acorn.io/v1/unmarshal.go
```diff
@@ -257,13 +257,15 @@ func impliedVolumesForContainer(app *AppSpec, containerName, sideCarName string,
...
-					continue
-				}
-				if existingSize.Cmp(vSize) < 0 {
-					existing.Size = v.Size
+				if v.Size != "" {
+					vSize, err := resource.ParseQuantity((string)(v.Size))
+					if err != nil {
+						// ignore error
+						continue
+					}
```
jacobdonenfeld, line 265: Worth logging? When would this happen?

pkg/apis/internal.acorn.io/v1/unmarshal.go
```diff
@@ -257,13 +257,15 @@ func impliedVolumesForContainer(app *AppSpec, containerName, sideCarName string,
...
+					if existingSize.Cmp(vSize) < 0 {
+						existing.Size = v.Size
+					}
```
jacobdonenfeld, line 268: Log or warn user when trying to shrink volume in else condition?

jacobdonenfeld (approved):
LGTM, small comments and questions. 

One thing I couldn't find when walking the code is adequate usage of MinSize/MinSizeQuantity.  Is there feedback to the user when defining or updating a volume attempting to shrink it, or when below the minimum value? Should there be a check to ensure its above the minimum size somewhere in volume.go:toPVCs? I see it referenced once.

pkg/apis/internal.acorn.io/v1/unmarshal.go
thedadams: This isn't shrinking the volume, it's growing it. Additionally this behavior isn't changing.

pkg/apis/internal.acorn.io/v1/unmarshal.go
thedadams: As far as I can tell, this is not really possible. Additionally, this isn't being changed.

integration/run/testdata/volume/Acornfile
thedadams: There are tests for this case. It defaults to 10G.

pkg/apis/internal.acorn.io/v1/unmarshal.go
jacobdonenfeld: Makes sense. Thought along the lines of:
```
if existingSize.Cmp(vSize) <= 0 {
    existing.Size = v.Size
    } else {
    log.Warn("Unable to shrink volume size")
    }
```

tylerslaton (approved):

Issue #1238, opened 2023-02-17
Release acorn CD broken because gomock not installed


jacobdonenfeld:
Missing stage setup-ci-env before running validate-ci

PR #1239, opened 2023-02-17
Missing make setup-ci-env for CD workflows breaking release (#1238)

jacobdonenfeld:
make validate-ci depends on make setup-ci-env. Make setup-ci-env now only installs golangci-lint if it is not already installed, to avoid reinstalling it or installing it twice if it was installed some other way such as brew.

One potential issue is that it doesn't keep golangci-lint up to date on a dev's system, but that isn't a huge deal.

### Checklist

Fixes: #1238

### DIFF
--- a/.github/workflows/aws-eks-test.yml
+++ b/.github/workflows/aws-eks-test.yml
@@ -25,7 +25,6 @@ jobs:
       - uses: actions/setup-go@v3
         with:
           go-version: "1.19"
-      - run: make setup-ci-env
       - run: make validate-ci
       - run: make validate
       - run: make build
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -24,7 +24,6 @@ jobs:
       - uses: debianmaster/actions-k3s@v1.0.5
         with:
           version: 'v1.23.6-k3s1'
-      - run: make setup-ci-env
       - run: make validate-ci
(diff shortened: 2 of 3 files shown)


.github/workflows/main-mac-smoke-test.yaml
```diff
@@ -42,6 +42,7 @@ jobs:
           echo "${COSIGN_KEY}" > "$GITHUB_WORKSPACE/cosign.key"
         env:
           COSIGN_KEY: ${{ secrets.COSIGN_KEY }}
+      - run: make setup-ci-env
```
tylerslaton, line 45: Should `validate-ci` just have a prerequisite of `setup-ci-env`?

tylerslaton (approved):
Non-blocking comment, feel free to merge this since its blocking main releases 😅

.github/workflows/main-mac-smoke-test.yaml
tylerslaton: Discussed offline.

jacobdonenfeld: If done with current implementation, if a dev runs it locally, it would force reinstall of golangci-lint, or install it via another way if its already installed via brew or some other way. Adding a posix check to see if the command exists and installing golangci-lint if it is not already on the path.

Issue #502, opened 2022-08-09
Support image signatures & attestations


cjellick:
This should be done using sigstore/cosign and supporting technologies.

We discussed this feature in our own internal "dogfooding" scenario. We envision this giving us the ability to promote a hub release from staging to production. Here's the less than ideal options we have now:
- To promote now, we might need to create a new tag (like going from a -rc to a non-rc image so that auto-deploy picks it up), but that means we are fundamentally doing another build and creating a new binary. That is not the ideal. The ideal is to promote an existing binary.
- We could also completely control it via gitops and not use acorn's auto-deploy feature, changing a tag value from the "old" to "new" in our flux git repo, but we really want to get to an acorn native approach that doesn't rely on a gitops tol for this.

So, our ideal scenario for our own process is to have auto-deploy working off of an attestation on existing release. And then the trigger to deploy would be someone signing the image somewhere. It could even be QA signing off on the image.

We want this to be general purpose enough to fit different org's use cases. So, an app owner should be able to define what signatures/attestations/attributes need to be on an image to qualify it for promotion.

Once we get the core functionality down, we need to also consider other things like:
- auditing
- rollback 
- What happens if you revoke a sign-off?

cjellick:
Beyond just generically signable, we want attestations and/or annotations to indicate arbitrary checks or approvals the image has gone through

cjellick:
In dogfooding scenario:
- creating a new tag is not good because its a new binary
- updating gitops is still making an external change
- ideal scenario for our own process is auto-deploy working off of an attestation on existing binary

cjellick:
@thedadams just a heads up - i've asked thorsten to start looking at this.

PR #1240, opened 2023-02-17
Require annotated image signatures to promote Acorn images to protected projects (#502)

iwilltry42:
## About

TODO

## To-Do

- [x] Add basic API resource für AcornImageAllowRules including internal API counterpart
- [ ] Add rule enforcement in controller (cosign verification)
- [ ] Add RBAC: read-only for users, r/w for admins/operators
- [ ] Ensure project level behavior
- [ ] Add CLI options to manage (at least get/list/delete) the resources
- [ ] Add CLI option to sign Acorn images (keyless, via OIDC)
- [ ] ... TBD ...

### References

- #502

### Checklist
- [ x] The title of this PR would make a good line in Acorn's Release Note's Changelog


pkg/imageallowrules/imageallowrules.go
```diff
@@ -0,0 +1,118 @@
...
+	return fmt.Sprintf("image %s is not allowed by rule %s", e.Image, e.Rule)
+}
+
+func (e *ErrImageNotAllowed) Is(target error) bool {
+	_, ok := target.(*ErrImageNotAllowed)
+	return ok
+}
+
+// CheckImageAllowed checks if the image is allowed by the ImageAllowRules in the same namespace as the AppInstance
+// @return string denoting the reason for *not* allowing the image (empty string if image is allowed)
```
g-linville, line 31: Nit: this function doesn't return a string anymore

g-linville:
@iwilltry42 I'm having trouble getting this to work in my local installation. Can you give me a quick example of how I can do this just using acorn and cosign? This is what I tried:

1. Build a simple acorn image and push it to Docker Hub
2. Use cosign to sign the acorn image on Docker Hub (i.e. `cosign sign --key <key> <image>@<digest>`)
3. Create a ProjectImageAllowRules for my Acorn project which only allows images signed by cosign key.
4. Attempt to run the image (`acorn run <image>@<digest>`), but it fails because it parses a bit weirdly (`@sha256@sha256` appears in the error message, which tells me I shouldn't be using the digest in the `acorn run` command).
5. Attempt to run the image with the tag (`acorn run <image>:<tag>`), but it fails and the error message saying that it doesn't match the rule shows a different digest than the one I see on Docker Hub.

iwilltry42:
@g-linville I updated the PR description with a How-To

pkg/imageallowrules/imageallowrules.go
iwilltry42: Yup, removed that :)

g-linville:
What is the intended behavior when imageAllowRules are changed when Acorn apps are already running in the cluster? I deployed an app and then created allowRules that would disallow the app. Now I can't stop it with `acorn stop`:

```
❯ acorn stop nginx-unallowed
  ✗  ERROR:  stopping nginx-unallowed: App.api.acorn.io "nginx-unallowed" is invalid: spec.image: Invalid value: "docker.io/grantlinville/nginx:simple-acorn": image docker.io/grantlinville/nginx@sha256:bbaef1fea727526ff4c422dc24fbc29cb069aa10e64b1709112db3f59ea5ab24 is not allowed by rule acorn/testrules-project
```

iwilltry42:
@g-linville good point!
Actually this shouldn't have any effect on already running apps at this point (we aligned on not supporting runtime-effects like rollbacks etc. right now).

I've reverted the introduced middlewares in the router that checks whether an image is allowed.

However, I feel like it should still deny **starting** an app with a disallowed image.
I.e. IMO `acorn run` and `acorn start` should return an error and also `acorn update` should return an error if it changes the image to a disallowed image. I'll work that out.

tylerslaton:
@iwilltry42 I notice in the error messaging to the user it could be hard to understand *why* their image is not allowed. Take your example for instance. In it we need to specify a certain tag when signing the image but the error message just says what rule isn't allowing it. Is it important that we communicate how a user can amend this fix or do we feel that reading the rule is sufficient for debugging?

pkg/cosign/cosign_test.go
```diff
@@ -0,0 +1,234 @@
+package cosign
+
+import (
+	"context"
+	_ "embed"
+	"fmt"
+	"io/ioutil"
```
tylerslaton, line 7: Just like the bot says, this is causing CI validation to fail.

iwilltry42: Yup, didn't check that, fixed.

iwilltry42:
@tylerslaton how much detail would you like to see in the error message?
I.e. failure in `anyOf` or `allOf` rule would be easy to expose, but e.g. which rule of the annotations didn't work would require us to break up the Kubernetes label selector logic that we're using for this.

tylerslaton:
I see what you're saying here. Here's what I was hoping for if I was using something using this feature:

```
$ acorn run <image>                                         
  ✗  ERROR:  App.api.acorn.io "floral-silence" is invalid: spec.image: Invalid value: "<image>": image <image> is not allowed by /testrules-cluster.0.annotation
```

or

```
$ acorn run <image>                                         
  ✗  ERROR:  App.api.acorn.io "floral-silence" is invalid: spec.image: Invalid value: "<image>": image <image> is not allowed by /testrules-cluster.0.anyOf
```

This tells me what rule is not allowing my image and what part of the rule I need to address. Looking at the code I notice that this would require bubbling up and parsing errors in the api server. What are your thoughts though; is that worth effort? I mainly want to avoid a situation where a user is unable to determine what rule is failing in a large`*ImageAllowRules` object.

iwilltry42:
Good pointers  @tylerslaton , thanks!

Here we go:

- Failure on `allOf` can be fairly specific:

```
✗  ERROR:  App.api.acorn.io "still-water" is invalid: spec.image: Invalid value: "my.registry.local/public/testsig:signed-cluster": disallowed by imageAllowRules: image my.registry.local/public/testsig:signed-cluster is not allowed by rule /testrules-cluster.signatures.0.allOf.0 (annotations unmatched)
```

- Failure on `anyOf` is hard to output more specific, since we'd have to output the issue for every checked key/signer for it to bring any value. Potentially one could output it failed on all signers by key or if there was one where the signature matched but the annotation didn't, e.g. appending `(signatures unmatched)` if check failed for all signers by key verification or `(annotations unmatched)` if all signers were verified by key, but the annotations didn't match (for all of them) or `(signatures/annotations unmatched)` for mixed cases. I'm not sure that would help a lot though.

```
✗  ERROR:  App.api.acorn.io "young-pond" is invalid: spec.image: Invalid value: "my.registry.local/public/testsig:signed-cluster": disallowed by imageAllowRules: image my.registry.localpublic/testsig:signed-cluster is not allowed by rule /testrules-cluster.signatures.0.anyOf
```

WDYT?

iwilltry42:
... rebased once again and fixed validation error ...

pkg/roles/roles.go
```diff
@@ -144,6 +145,8 @@ var (
 					"clustervolumeclasses",
 					"projectcomputeclasses",
 					"clustercomputeclasses",
+					"imageallowrules",
+					"clusterimageallowrules",
```
thedadams, line 149: Do `projectimageallowrules` need to be added here too?

pkg/server/registry/apigroups/acorn/imageallowrules/validator.go
```diff
@@ -0,0 +1,56 @@
...
+	aiar := obj.(*apiv1.ClusterImageAllowRules)
+	result = append(result, validateSignatureRules(ctx, aiar.Signatures)...)
+	return
+}
+
+func (s *ClusterValidator) ValidateUpdate(ctx context.Context, obj, old runtime.Object) field.ErrorList {
+	return s.Validate(ctx, obj)
+}
+
+func (s *ClusterValidator) ValidateCreate(ctx context.Context, obj, old runtime.Object) field.ErrorList {
```
thedadams, line 24: Is this necessary? The interface that `WithValidateCreate` accepts has a `Validate` method.

pkg/server/registry/apigroups/acorn/imageallowrules/validator.go
```diff
@@ -0,0 +1,56 @@
...
+	aiar := obj.(*apiv1.ProjectImageAllowRules)
+	result = append(result, validateSignatureRules(ctx, aiar.Signatures)...)
+	return
+}
+
+func (s *ProjectValidator) ValidateUpdate(ctx context.Context, obj, old runtime.Object) field.ErrorList {
+	return s.Validate(ctx, obj)
+}
+
+func (s *ProjectValidator) ValidateCreate(ctx context.Context, obj, old runtime.Object) field.ErrorList {
```
thedadams, line 40: Same question here.

pkg/server/registry/apigroups/acorn/imageallowrules/strategy.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	imageAllowRules := apiv1.ImageAllowRulesList{Items: make(
+		[]apiv1.ImageAllowRules,
+		0,
+		len(clusterImageAllowRules.Items)+len(projectImageAllowRules.Items))}
+
+	for _, cic := range clusterImageAllowRules.Items {
+		imageAllowRules.Items = append(imageAllowRules.Items, apiv1.ImageAllowRules{
+			ObjectMeta: v1.ObjectMeta{Name: cic.Name, CreationTimestamp: cic.CreationTimestamp},
```
thedadams, line 52: How are cluster and project objects with the same name handled? Does the project object take precedence? Would it be possible for a user to select a cluster image allow rule when a project image allow rule with the same name exists?

All of these questions lead to: should we not list a cluster allow rule if a project allow rule with the same name exists?

pkg/apis/internal.acorn.io/v1/imageallowrules.go
```diff
@@ -0,0 +1,83 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type ClusterImageAllowRulesInstanceList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []ClusterImageAllowRulesInstance `json:"items"`
+}
+
+func (list *ClusterImageAllowRulesInstanceList) AsProjectImageAllowRulesInstanceList() *ProjectImageAllowRulesInstanceList {
+	var items []ProjectImageAllowRulesInstance
```
thedadams, line 74: nit:
```suggestion
	items := make([]ProjectImageAllowRulesInstance, 0, len(list.Items))
```

pkg/apis/internal.acorn.io/v1/imageallowrules.go
```diff
@@ -0,0 +1,83 @@
...
+	for _, item := range list.Items {
+		items = append(items, ProjectImageAllowRulesInstance(item))
+	}
+	return &ProjectImageAllowRulesInstanceList{
+		TypeMeta: list.TypeMeta,
+		ListMeta: list.ListMeta,
+		Items:    items,
+	}
```
thedadams, line 82: There is some weirdness with including `TypeMeta` and `ListMeta` here with respect to resource version, generation, and kinds.

Would returning the slice be enough? If not, then maybe leave out the `TypeMeta` and `ListMeta`?

pkg/autoupgrade/tags.go
```diff
@@ -41,8 +41,27 @@ func findLatestTagForImageWithPattern(ctx context.Context, c daemonClient, names
...
+		if err := c.checkImageAllowed(ctx, namespace, img); err != nil {
+			// remove the tag from the list and try again
+			tags = slices.Filter(nil, tags, func(tag string) bool { return tag != nTag })
+		} else {
+			// found a valid tag that is allowed by all rules, so use it
+			newTag = nTag
+			newImage = img
+			break
+		}
+	}
```
thedadams, line 62: Can this logic be incorporated into `FindLatest` so that the tag matching happens as few times a possible?

pkg/controller/appdefinition/checkimageallowed.go
```diff
@@ -0,0 +1,65 @@
...
+	v1 "github.com/acorn-io/acorn/pkg/apis/internal.acorn.io/v1"
+	"github.com/acorn-io/acorn/pkg/condition"
+	"github.com/acorn-io/acorn/pkg/imageallowrules"
+	"github.com/acorn-io/baaah/pkg/router"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/sirupsen/logrus"
+)
+
+// ImageAllowed is a middleware that checks if the image is allowed by the image allow rules.
+func ImageAllowed(h router.Handler) router.Handler {
```
thedadams, line 17: This function looks like it is not used anywhere. Can it be removed?

pkg/controller/appdefinition/checkimageallowed.go
```diff
@@ -0,0 +1,65 @@
...
+				} else {
+					cond.Error(fmt.Errorf("no image specified"))
+				}
+				return nil
+			}
+		}
+
+		if err := imageallowrules.CheckImageAllowed(req.Ctx, req.Client, appInstance.Namespace, targetImage, remote.WithTransport(transport)); err != nil {
+			if errors.Is(err, &imageallowrules.ErrImageNotAllowed{}) {
+				logrus.Warning(err)
```
thedadams, line 52: It seems that if we get an `imageallowrules.ErrImageNotAllowed` then our logic has worked as expected. Meaning that logging a warning here is possibly not warranted?

pkg/controller/appdefinition/checkimageallowed.go
```diff
@@ -0,0 +1,65 @@
...
+		if targetImage == "" {
+			if appInstance.Status.AppImage.Name != "" {
+				targetImage = appInstance.Status.AppImage.Name
+			} else {
+				if unknownReason != "" {
+					cond.Unknown(unknownReason)
+				} else {
+					cond.Error(fmt.Errorf("no image specified"))
+				}
+				return nil
```
thedadams, line 46: What situation would lead to this return? Do we want to be a message on the condition in this case?

pkg/server/registry/apigroups/acorn/imageallowrules/validator.go
```diff
@@ -0,0 +1,56 @@
...
+	return s.Validate(ctx, obj)
+}
+
+func validateSignatureRules(ctx context.Context, sigRules internalv1.ImageAllowRuleSignatures) (result field.ErrorList) {
+	if sigRules.Rules == nil || len(sigRules.Rules) == 0 {
```
thedadams, line 45: This can be simplified:
```suggestion
	if len(sigRules.Rules) == 0 {
```

pkg/server/registry/apigroups/acorn/imageallowrules/validator.go
```diff
@@ -0,0 +1,56 @@
...
+		result = append(result, field.Invalid(field.NewPath("signatures"), sigRules, "must not be empty"))
```
thedadams, line 46: nit: return here so that the `else` can be removed?

pkg/server/registry/apigroups/acorn/imageallowrules/validator.go
```diff
@@ -0,0 +1,56 @@
...
+	} else {
+		for i, rule := range sigRules.Rules {
+			if (rule.SignedBy.AnyOf == nil || len(rule.SignedBy.AnyOf) == 0) && (rule.SignedBy.AllOf == nil || len(rule.SignedBy.AllOf) == 0) {
```
thedadams, line 49: Similarly simplified:
```suggestion
			if len(rule.SignedBy.AnyOf) == 0 && len(rule.SignedBy.AllOf) == 0 {
```

thedadams (commented):
I reviewed everything except `cosign.go`. I am running out of steam, but I wanted to get most of this reviewed.

pkg/roles/roles.go
iwilltry42: Absolutely!

pkg/apis/internal.acorn.io/v1/imageallowrules.go
iwilltry42: I see no problem returning only the slice here.

pkg/server/registry/apigroups/acorn/imageallowrules/strategy.go
iwilltry42: They're additive, so you cannot "replace" or negate a cluster rule with a project role, you can only tighten the rules, not loosen them.

> Would it be possible for a user to select a cluster image allow rule when a project image allow rule with the same name exists?

What do you mean by "selecting a rule" here?

pkg/server/registry/apigroups/acorn/imageallowrules/validator.go
iwilltry42: Absolutely not necessary, you're right!

pkg/controller/appdefinition/checkimageallowed.go
iwilltry42: Yeah, it's more useful for Acorn admins that get complaints from users.
But actually I improved the user-facing error messages recently, making this irrelevant as it adds no extra information.

pkg/controller/appdefinition/checkimageallowed.go
iwilltry42: Since `appInstance.Status.AppImage.Name == ""` means that the image was not pulled/applied to the app yet, we would only end up here, if in addition to that the determination logic cannot find a suitable image, which is e.g. when the autoupgrade sync was never triggered before (see https://github.com/acorn-io/acorn/blob/main/pkg/controller/appdefinition/pullappimage.go#L77)

It's pretty unlikely though, except for something went really wrong.
I think though that we could never hit the `else` clause here.

pkg/apis/api.acorn.io/v1/imageallowrules.go
```diff
@@ -0,0 +1,34 @@
+package v1
+
+import (
+	v1 "github.com/acorn-io/acorn/pkg/apis/internal.acorn.io/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type ProjectImageAllowRules v1.ProjectImageAllowRulesInstance
```
tylerslaton, line 10: Two thoughts here.

1. Should this be an `admin` API? Creating `ImageAllowRules` seems to be an administrative adjacent feature similar to `ComputeClasses` and `VolumeClasses`.
2. I know that we're getting rid of the concept of `Project*Instance` resources. As such, should we consider removing this type or is there some reason this type wouldn't be affected?

pkg/apis/api.acorn.io/v1/scheme.go
```diff
@@ -57,6 +57,12 @@ func AddToSchemeWithGV(scheme *runtime.Scheme, schemeGroupVersion schema.GroupVe
 		&AcornImageBuildList{},
 		&ComputeClass{},
 		&ComputeClassList{},
+		&ImageAllowRules{},
+		&ImageAllowRulesList{},
+		&ProjectImageAllowRules{},
+		&ProjectImageAllowRulesList{},
+		&ClusterImageAllowRules{},
+		&ClusterImageAllowRulesList{},
```
tylerslaton, line 65: I'm slightly confused by these types. When we did this pattern for `ComputeClasses` and `VolumeClasses` it made sense as the `ImageAllowRules` (in this example) served as an aggregate user facing type for the `Project*` and `Cluster*` admin  facing types. I notice that there is no way for a user to query the currently available `ImageAllowRules` via the CLI so could you help understand the need for the `ImageAllowRules` type?

tylerslaton (commented):
Still reviewing but did have two main comments I wanted to make sure we talked through.

tylerslaton:
Also, looks like something is funky with the Mocks on this PR which is causing the test suite to fail.

pkg/server/registry/apigroups/acorn/imageallowrules/strategy.go
thedadams: What I am asking here is about cluster rules and project rules with the same name. It seems that if there is a cluster rule and project rule with the same name, then all of them are being returned here. So, for example, if a user lists all the image allow rules, they may see duplicated names/configurations.

There was a similar issue with compute/volume classes. We made the choice that if a project object has the same name as a cluster object, then the project object would be returned/used and not the cluster object.

Should the same rule apply here?

pkg/controller/appdefinition/checkimageallowed.go
thedadams: OK. That makes sense. Thank you for the explanation.

Do you think a message should be added to the condition just in case? Or is that a waste?

pkg/apis/api.acorn.io/v1/imageallowrules.go
iwilltry42: 1. I thought so too, but @ibuildthecloud mentioned in the beginning of our discussions, that it should be in the "normal" API but with restricted (get/list) access for non-admin users.

2. Yeah, I saw that, but wasn't sure about the reasoning. Do we have some docs on the decision somewhere? No hard feelings from my side.

pkg/apis/api.acorn.io/v1/scheme.go
iwilltry42: Yes, `ImageAllowRules` is an aggregate, and listing it as a combined list of cluster and project rules, it also works for me :thinking:

pkg/autoupgrade/tags.go
iwilltry42: Unfortunately I couldn't find any cleaner or more performant way of doing this.
Copied two options from Slack:

>a) Check for every tag -> that's not really performant, but at least linear
b) Check only for latest tag -> In that case, we'd need to keep a sorted list, to cycle through it in reverse or we will end up with the same issue
I see that (b) can achieve way better performance in all non-worst cases, but would also force us to go from "keep note of the latest tag" to "keep a sorted list" in the current logic.

We aligned on keeping it as it is for now.

iwilltry42:
**Update** as per the latest alignment, I'm going to remove the cluster-level implementation and rename all Project* resources to just ImageAllowRules.

pkg/apis/api.acorn.io/v1/scheme.go
iwilltry42: As per latest alignment, we won't have Acorn CLI functions for this right now.
Also, the distinction between Cluster* and Project* will be removed now.

pkg/server/registry/apigroups/acorn/imageallowrules/strategy.go
iwilltry42: As per latest alignment, we will only continue with project-level resources for now.

iwilltry42:
> Also, looks like something is funky with the Mocks on this PR which is causing the test suite to fail.

I wish I knew what's going on here :thinking:

pkg/imageallowrules/imageallowrules.go
```diff
@@ -0,0 +1,125 @@
...
+						return fmt.Errorf("failed to verify signature: %w", err)
+					}
+				}
+			}
+
+			// anyOf: only one signature must pass verification
+			if rule.SignedBy.AnyOf != nil {
+				anyOfOK := false
+				for anyOfRuleIndex, signer := range rule.SignedBy.AnyOf {
+					logrus.Debugf("Checking image %s against anyOf rule #%d in %s/%s", image, anyOfRuleIndex, ImageAllowRules.Namespace, ImageAllowRules.Name)
```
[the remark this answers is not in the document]
iwilltry42: I cleaned up the logging and error handling a bit here to make it more clear. Hope that addresses this :+1:

tylerslaton (approved):
Nice! Everything looks good to me now, nothing blocking.

g-linville (approved):

pkg/cosign/cosign.go
```diff
@@ -0,0 +1,316 @@
...
+		if err := json.Unmarshal(pld, &sci); err != nil {
+			return nil, fmt.Errorf("error decoding the payload: %w", err)
+		}
+
+		sigPayloads = append(sigPayloads, sci)
+	}
+	return sigPayloads, nil
+}
+
+var ALGORITHMS = map[string]crypto.Hash{
```
thedadams, line 293: Should this be lowercase so it doesn't get exported?

cjellick:
@iwilltry42 i am sure @sangee2004 will use your above walk through when she goes to test this feature. its a little out of date (the CRD is misnamed). Can you update it?

pkg/imageallowrules/imageallowrules.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func (e *ErrImageNotAllowed) Error() string {
+	return fmt.Sprintf("image %s is not allowed by rule %s.%s.%s", e.Image, e.Rule, e.SubruleType, e.SubrulePath)
+}
+
+func (e *ErrImageNotAllowed) Is(target error) bool {
+	_, ok := target.(*ErrImageNotAllowed)
+	return ok
+}
```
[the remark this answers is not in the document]
iwilltry42: This is actually implementing that `errors.Is` interface function for the custom `ErrImageNotAllowed`

pkg/imageallowrules/imageallowrules.go
```diff
@@ -0,0 +1,135 @@
...
+
+		// > Signatures
+		notAllowedErr.SubruleType = "signatures"
+		for ruleIndex, rule := range ImageAllowRules.Signatures.Rules {
+
+			verifyOpts.AnnotationRules = rule.Annotations
+			notAllowedErr.SubrulePath = fmt.Sprintf("%d", ruleIndex)
+
+			// allOf: all signatures must pass verification
+			if rule.SignedBy.AllOf != nil {
```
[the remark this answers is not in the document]
iwilltry42:
```suggestion
			if len(rule.SignedBy.AllOf) != 0 {
```

pkg/cosign/cosign.go
iwilltry42:
```suggestion
var algorithms = map[string]crypto.Hash{
```

pkg/cosign/cosign.go
```diff
@@ -0,0 +1,316 @@
...
+	"":       crypto.SHA256,
+	"sha256": crypto.SHA256,
+	"sha512": crypto.SHA512,
+}
+
+func LoadKey(ctx context.Context, keyRef string, algorithm string) (verifier signature.Verifier, err error) {
+	if strings.HasPrefix(strings.TrimSpace(keyRef), "-----BEGIN PUBLIC KEY-----") {
+		// no scheme, inline PEM
+		verifier, err = decodePEM([]byte(keyRef), ALGORITHMS[algorithm])
```
iwilltry42, line 302:
```suggestion
		verifier, err = decodePEM([]byte(keyRef), algorithms[algorithm])
```

cjellick (approved):
Ran some basic use cases for normal `acorn run` and auto-upgrade scenarios. Looks good!

As someone else already pointed out, the error messages are a little clunky, but maybe we can improve on that over time

pkg/server/registry/apigroups/acorn/imageallowrules/strategy.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	imageAllowRules := list.(*apiv1.ImageAllowRulesList)
+	for _, imageAllowRulesItem := range imageAllowRules.Items {
+		if imageAllowRulesItem.Name == name {
+			return &imageAllowRulesItem, nil
+		}
+	}
```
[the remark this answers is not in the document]
iwilltry42: Actually the whole strategy was unused, so I removed it.

pkg/imageallowrules/imageallowrules.go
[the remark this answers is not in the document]
thedadams: I thought that `errors.Is` would work without this. Not a big deal at this point.

thedadams (approved):

sangee2004:
Tested  with acorn version `v0.6.0-98-g871782ee+871782ee`

Support for image signatures and attestations are now available to users by creating `ImageAllowRule` at project level.

Bugs found when testing this feature are tracked as separate issues:

https://github.com/acorn-io/acorn/issues/1409
https://github.com/acorn-io/acorn/issues/1410
https://github.com/acorn-io/acorn/issues/1411
https://github.com/acorn-io/acorn/issues/1412
https://github.com/acorn-io/acorn/issues/1414

PR #1246, opened 2023-02-21
fix info's integration tests breaking on eks (#1244)

### LINKED ISSUE #1244: EKS tests broken due to checking a specific number of namespaces
broken by #1193

jacobdonenfeld:
Refactoring integration tests to use generateName when creating namespaces, as well as verifying namespace names, instead of the length of info's response. This allows it to be more flexible on what cluster it is running on, and allowing other namespaces to preexist.

### Checklist

Closes #1244

### DIFF
--- a/integration/helper/assertions.go
+++ b/integration/helper/assertions.go
@@ -1,0 +1,28 @@
+package helper
+
+import (
+	"testing"
+)
+
+// Subset returns true if the first array is completely
+// contained in the second array. There must be at least
+// the same number of duplicate values in second as there
+// are in first.
+func Subset[V, W any, Z comparable](t *testing.T, first []V, second []W, firstLookup func(V) Z, secondLookup func(W) Z) bool {
+	t.Helper()
+
+	set := make(map[Z]int)
+	for _, value := range second {
+		set[secondLookup(value)] += 1
+	}
+
+	for _, value := range first {
+		parsedValue := firstLookup(value)
+		if set[parsedValue] < 1 {
+			return false
+		}
+		set[parsedValue]--
+	}
+
+	return true
+}
--- a/integration/helper/namespace.go
+++ b/integration/helper/namespace.go
@@ -24,7 +24,24 @@ func TempNamespace(t *testing.T, client client.Client) *corev1.Namespace {
 		Spec:   corev1.NamespaceSpec{},
 		Status: corev1.NamespaceStatus{},
 	}
-	return namedTempCreate(t, client, ns)
+	return tempCreateHelper(t, client, ns)
+}
+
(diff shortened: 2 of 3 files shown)


tylerslaton (approved):

integration/helper/assertions.go
```diff
@@ -0,0 +1,34 @@
...
+		if count, found := set[parsedValue]; !found {
+			return false
+		} else if count < 1 {
+			// Provide a useful failure message
+			assert.Contains(t, set, parsedValue)
+			return false
+		} else {
+			set[parsedValue] = count - 1
+		}
+	}
```
thedadams, line 31: If `parsedValue` is in `set`, but `count < 1`, then I think `assert.Contains` will pass. Also, this can be simplified. Note that reading a key that doesn't exist in a map is perfectly valid and will return the type's zero value.
```suggestion
		if set[parsedValue] < 1 {
			return false
		} 
		set[parsedValue]--
```

integration/helper/namespace.go
```diff
@@ -40,17 +57,19 @@ func NamedTempProject(t *testing.T, client client.Client, name string) *corev1.N
...
 	}
-	return namedTempCreate(t, client, ns)
+	return tempCreateHelper(t, client, ns)
 }
 
-func namedTempCreate(t *testing.T, client client.Client, namespaceObject corev1.Namespace) *corev1.Namespace {
+func tempCreateHelper(t *testing.T, client client.Client, namespaceObject corev1.Namespace) *corev1.Namespace {
 	t.Helper()
-	namespaceName := namespaceObject.GetNamespace()
+	namespaceName := namespaceObject.GetNamespace() // Will be "" for generateName
```
thedadams, line 65: Do you mean `namespaceObject.GetName()` here? I think `namespaceObject.GetNamespace()` is always `""`.

integration/helper/namespace.go
```diff
@@ -73,8 +92,11 @@ func namedTempCreate(t *testing.T, client client.Client, namespaceObject corev1.
 		t.Fatal(err)
 		return nil
 	}
+
 	t.Cleanup(func() {
-		namespaceDeleting := namespaceName
+		// At this point, if using generateName
+		// .getNamespace() and .Namespace aren't populated, just Name
```
thedadams, line 98: I think this comment should be updated as well.

thedadams (changes requested):

integration/helper/assertions.go
jacobdonenfeld: Great point, moved the assertion outside of the function. TY

thedadams (approved):

Issue #358, opened 2022-07-28
List containers by App


iwilltry42:
- [ ] `acorn containers -h` shows `acorn container [flags] [APP_NAME...]`, where passing in an actual app name would result in
  ```bash
  $ acorn containers little-water
  Error: containers.api.acorn.io "little-water" not found
  ```
  - I guess it should be `acorn container [flags] [CONTAINER_NAME...]`
- [ ] It would still be nice to have a way to filter the `acorn containers` list output by app name, e.g. via a flag `acorn containers --app APP-NAME` or by having some `--filter` options, similar to the docker filters

sangee2004:
Tested with `acorn version v0.5.1-82-g652e22e7+652e22e7`

Able to list containers by apps and container names now.

```
%acorn app test1
NAME      IMAGE          HEALTHY   UP-TO-DATE   CREATED   ENDPOINTS                                                         MESSAGE
test1     b1ef5ce44c21   1         1            19m ago   http://frontend-test1-80c87462.local.on-acorn.io => frontend:80   OK

% acorn containers test1
NAME                                      APP       IMAGE                                                                     STATE     RESTARTCOUNT   CREATED   MESSAGE
test1.frontend-8584688cbc-9wdt8           test1     sha256:ee451425b88b30205998e298b96470bacb7a9eaa75d9124f759d9cf314af8ba8   running   0              10m ago   
test1.frontend-8584688cbc-9wdt8.sidecar   test1     nginx                                                                     running   0              10m ago   

% acorn containers test1.frontend-8584688cbc-9wdt8 
NAME                              APP       IMAGE                                                                     STATE     RESTARTCOUNT   CREATED   MESSAGE
test1.frontend-8584688cbc-9wdt8   test1     sha256:ee451425b88b30205998e298b96470bacb7a9eaa75d9124f759d9cf314af8ba8   running   0              10m ago   

```

We are not able to list sidecar containers by their names which is tracked in https://github.com/acorn-io/acorn/issues/1269

PR #1252, opened 2023-02-23
Draft: List all containers in app with 'acorn containers <app name>' (#358)

g-linville:
re: #358

This change allows users to easily list all containers in a given app by running `acorn containers <app name>`.

Here is an example using the example acorn from the Getting Started guide on our docs site:

```
$ acorn containers flask
NAME                           APP       IMAGE                                                                     STATE     RESTARTCOUNT   CREATED   MESSAGE
flask.app-6fc4dff597-mpt5s     flask     sha256:ed041817d15ba2dfb043c5a101ea9d8e17141a272f804c41452de549d356bdcf   running   0              87m ago   
flask.cache-7b9c8d94f8-fpcv5   flask     redis:alpine                                                              running   0              87m ago   
flask.db-6c444c5654-x5fcc      flask     postgres:alpine                                                           running   0              87m ago 
```

If we want to be able to kill all containers in an app just by using the app name, that should be a fairly easy change to make as well, but I didn't implement that yet since the issue was specifically just about listing containers.

It looks like the current automated tests can't really distinguish between `acorn container <container name>` and `acorn container <app name>` because the mock client uses identical built-in names for apps and for containers ("`found`"). This also is something I could change if we want, but I didn't want to mess with it right now because I don't know the full impact it would have on our other tests.

### Checklist
- [ ] The title of this PR ends with a link to the main issue being address in paranthesis, like: `This is a title (#1216)`. [Here's an example](https://github.com/acorn-io/acorn/pull/1199)

### DIFF
--- a/pkg/cli/containers.go
+++ b/pkg/cli/containers.go
@@ -3,6 +3,7 @@ package cli
 import (
 	cli "github.com/acorn-io/acorn/pkg/cli/builder"
 	"github.com/acorn-io/acorn/pkg/cli/builder/table"
+	"github.com/acorn-io/acorn/pkg/client"
 	"github.com/acorn-io/acorn/pkg/tables"
 	"github.com/spf13/cobra"
 
@@ -39,11 +40,23 @@ func (a *Container) Run(cmd *cobra.Command, args []string) error {
 	out := table.NewWriter(tables.Container, a.Quiet, a.Output)
 
 	if len(args) == 1 {
-		app, err := c.ContainerReplicaGet(cmd.Context(), args[0])
+		app, err := c.AppGet(cmd.Context(), args[0])
 		if err != nil {
-			return err
+			// see if it's the name of a container instead
+			container, err := c.ContainerReplicaGet(cmd.Context(), args[0])
+			if err != nil {
+				return err
+			}
+			out.Write(container)
+		} else {
+			containers, err := c.ContainerReplicaList(cmd.Context(), &client.ContainerReplicaListOptions{App: app.Name})
+			if err != nil {
+				return err
+			}
+			for _, container := range containers {
+				out.Write(container)
+			}
 		}
-		out.Write(app)
 		return out.Err()
 	}
 


pkg/cli/containers_test.go
```diff
@@ -38,36 +87,40 @@ func TestContainer(t *testing.T) {
...
 			},
 			args: args{
-				args:   []string{"--", "found.container"},
-				client: &testdata.MockClient{},
+				args:   []string{"--", "found.container1"},
+				client: mClient,
 			},
 			wantErr: false,
-			wantOut: "NAME              APP       IMAGE     STATE     RESTARTCOUNT   CREATED    MESSAGE\nfound.container                                 0              292y ago   \n",
+			wantOut: "NAME               APP       IMAGE     STATE     RESTARTCOUNT   CREATED    MESSAGE\nfound.container1                                 0              292y ago   \n",
```
thedadams, line 123: I realize you didn't do this originally, but there is a problem here. These tests are going to start failing every year because `292y` will increase to `293y`, and so on.

While you are here changing this code, could you add a `CreationTimestamp` to the `ObjectMeta` objects here? Something like `time.Now().AddDate(-10, 0, 0)` should make it consistent for the foreseeable future.

thedadams (changes requested):

pkg/cli/containers.go
```diff
@@ -39,11 +40,23 @@ func (a *Container) Run(cmd *cobra.Command, args []string) error {
...
+				return err
+			}
+			out.Write(container)
+		} else {
+			containers, err := c.ContainerReplicaList(cmd.Context(), &client.ContainerReplicaListOptions{App: app.Name})
+			if err != nil {
+				return err
+			}
+			for _, container := range containers {
+				out.Write(container)
```
thedadams, line 57: Sorry, one more comment. Should we make this output similar to the output below in that stopped containers are only printed if the `-a` flag is passed?

g-linville: I'm not 100% sure. It looks like the case for that below is when the command run is simply `acorn containers` at which point it is pretty much the same as just `acorn ps`. I can make it behave that way if that is what would be best.

cjellick: Yeah, let's make `-a` consistently the same
- `acorn containers` - all running containers
- `acorn containers -a` - all running and stopped containers
- `acorn containers my-app` - all running containers belonging to my-app
- `acorn containers -a my-app` - all running and stopped containers belong to my-app

So (probably obvious), you'll need to add the `a.All || container.Status.Columns.State != "stopped" ` check

pkg/cli/containers.go
```diff
@@ -39,11 +40,25 @@ func (a *Container) Run(cmd *cobra.Command, args []string) error {
...
+		} else {
+			containers, err := c.ContainerReplicaList(cmd.Context(), &client.ContainerReplicaListOptions{App: app.Name})
+			if err != nil {
+				return err
+			}
+			for _, container := range containers {
+				if a.All || container.Status.Columns.State != "stopped" {
+					out.Write(container)
+				}
+			}
```
thedadams, line 60: nit: I would prefer this be DRY-ed up.

thedadams (approved):

pkg/cli/containers.go
g-linville: I restructured this whole function as a switch statement and created a helper function to do the standard printing. 👍

pkg/cli/containers.go
```diff
@@ -38,29 +39,49 @@ func (a *Container) Run(cmd *cobra.Command, args []string) error {
 
 	out := table.NewWriter(tables.Container, a.Quiet, a.Output)
 
-	if len(args) == 1 {
-		app, err := c.ContainerReplicaGet(cmd.Context(), args[0])
+	switch len(args) {
+	case 0:
+		if err := printContainerReplicas(cmd.Context(), c, nil, a.All, &out); err != nil {
```
cjellick, line 44:
```suggestion
		// No app or container name supplied, list all containers
		if err := printContainerReplicas(cmd.Context(), c, nil, a.All, &out); err != nil {
```

pkg/cli/containers.go
```diff
@@ -38,29 +39,49 @@ func (a *Container) Run(cmd *cobra.Command, args []string) error {
...
+			return err
+		}
+	case 1:
+		app, err := c.AppGet(cmd.Context(), args[0])
```
cjellick, line 48:
```suggestion
		// One app or container name supplied
		app, err := c.AppGet(cmd.Context(), args[0])
```

pkg/cli/containers.go
```diff
@@ -38,29 +39,49 @@ func (a *Container) Run(cmd *cobra.Command, args []string) error {
...
+			if err != nil {
+				return err
+			}
+			out.Write(container)
+		} else {
+			if err := printContainerReplicas(cmd.Context(), c, &client.ContainerReplicaListOptions{App: app.Name}, a.All, &out); err != nil {
+				return err
+			}
+		}
+	default:
```
cjellick, line 61:
```suggestion
	default:
		// More than one name supplied, iterate through containers and list any that match
```

cjellick (changes requested):
Some clarifying comments. WDYT?

cjellick (approved):

Issue #1262, opened 2023-02-24
No validation check done for  DEFAULT-SIZE to be within the SIZE-RANGE when editing ProjectVolumeClass.


sangee2004:
acorn version v0.5.1-79-g84d2894c+84d2894c

Steps to reproduce the problem:
We are NOT allowed to create ProjectVolumeClass where DEFAULT-SIZE is less than the SIZE-RANGE .
But when user modifies an existing ProjectVolumeClass , he is allowed to set a DEFAULT-SIZE that is outside the SIZE-RANGE .
User should  be presented with error message.

This the behavior seen when creating/editing ClusterVolumeClass with invalid default size. In this case user is presented with error message-`invalid: size.default: Invalid value: "4G": default size should be at least min size`

thedadams:
Additionally, the `storageClassName` field should be immutable.

sangee2004:
Tested with acorn version v0.6.0-rc1+f4c514dd

Validation check are now done for  DEFAULT-SIZE to be within the SIZE-RANGE when editing ProjectVolumeClass/ClusterVolumeClass.
Also we are not allowed to change storageClassName editing ProjectVolumeClass/ClusterVolumeClass..

```
The ProjectVolumeClass "pvclass3" is invalid: size.default: Invalid value: "9G": default size should be at least min size        
The ProjectVolumeClass "pvclass3" is invalid: size.default: Invalid value: "20G": default size should be at most max size
The ProjectVolumeClass "pvclass3" is invalid: storageClassName: Invalid value: "local-path-test4": storageClassName cannot 
```

PR #1275, opened 2023-02-27
Strict validation on volume class updates, including making the storage class name immutable (#1262)

thedadams:
### Checklist

Issue: https://github.com/acorn-io/acorn/issues/1262

### DIFF
--- a/integration/client/volumes/volumeclasses_test.go
+++ b/integration/client/volumes/volumeclasses_test.go
@@ -451,3 +451,200 @@ func TestCreateProjectDefaultWithExistingClusterDefault(t *testing.T) {
 
 	t.Fatal("project default not found")
 }
+
+func TestProjectVolumeClassUpdateValidation(t *testing.T) {
+	helper.StartController(t)
+
+	ctx := helper.GetCTX(t)
+	kclient := helper.MustReturn(kclient.Default)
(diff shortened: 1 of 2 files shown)


integration/client/volumes/volumeclasses_test.go
```diff
@@ -451,3 +451,200 @@ func TestCreateProjectDefaultWithExistingClusterDefault(t *testing.T) {
 
 	t.Fatal("project default not found")
 }
+
+func TestProjectVolumeClassUpdateValidation(t *testing.T) {
```
tylerslaton, line 455: nit:

Could/should we combine `TestProjectVolumeClassUpdateValidation` and `TestClusterComputeClassUpdateValidation` so the same test cases are run on each type?

tylerslaton (approved):

integration/client/volumes/volumeclasses_test.go
thedadams: No, we don't want the same test cases run on both types because there will already be a default ClusterVolumeClass when Acorn is installed. That means that the test won't be able to create a default ClusterVolumeClass. That is why they are separate.

jsilverio22 (approved):
straightforward changes to validation lgtm

PR #1281, opened 2023-02-28
Remove linkerd containers from acorn logs output

g-linville:
This modifies the `acorn logs` log to filter out logs from containers called `linkerd-proxy` or `linkerd-init`. ~~These logs can optionally be included by passing the `--include-proxy-containers` flag.~~

This can be tested by running the logging integration test in a cluster with linkerd and the acorn-linkerd-plugin installed.

### Checklist

### DIFF
--- a/docs/docs/100-reference/01-command-line/acorn_logs.md
+++ b/docs/docs/100-reference/01-command-line/acorn_logs.md
@@ -12,11 +12,10 @@ acorn logs [flags] [APP_NAME|CONTAINER_NAME]
 ### Options
 
 ```
-  -f, --follow                     Follow log output
-  -h, --help                       help for logs
-      --include-proxy-containers   Also display logs from Linkerd proxy containers
-  -s, --since string               Show logs since timestamp (e.g. 42m for 42 minutes)
-  -n, --tail int                   Number of lines in log output
+  -f, --follow         Follow log output
+  -h, --help           help for logs
+  -s, --since string   Show logs since timestamp (e.g. 42m for 42 minutes)
+  -n, --tail int       Number of lines in log output
 ```
 
 ### Options inherited from parent commands
--- a/docs/docs/100-reference/01-command-line/acorn_logs.md
+++ b/docs/docs/100-reference/01-command-line/acorn_logs.md
@@ -12,10 +12,11 @@ acorn logs [flags] [APP_NAME|CONTAINER_NAME]
 ### Options
 
 ```
-  -f, --follow         Follow log output
-  -h, --help           help for logs
-  -s, --since string   Show logs since timestamp (e.g. 42m for 42 minutes)
-  -n, --tail int       Number of lines in log output
+  -f, --follow                     Follow log output
+  -h, --help                       help for logs
+      --include-proxy-containers   Also display logs from Linkerd proxy containers
+  -s, --since string               Show logs since timestamp (e.g. 42m for 42 minutes)
+  -n, --tail int                   Number of lines in log output
 ```
 
 ### Options inherited from parent commands
--- a/integration/log/log_test.go
+++ b/integration/log/log_test.go
@@ -2,15 +2,15 @@ package log
 
 import (
 	"context"
-	"sort"
-	"strings"
-	"testing"
-
 	"github.com/acorn-io/acorn/integration/helper"
 	apiv1 "github.com/acorn-io/acorn/pkg/apis/api.acorn.io/v1"
+	v1 "github.com/acorn-io/acorn/pkg/apis/internal.acorn.io/v1"
 	"github.com/acorn-io/acorn/pkg/client"
 	"github.com/sirupsen/logrus"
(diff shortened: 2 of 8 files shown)


g-linville:
Still need to extend the tests to account for this new parameter. It's pretty late rn so I will work on that tomorrow.

docs/docs/100-reference/01-command-line/acorn_logs.md
```diff
@@ -12,10 +12,11 @@ acorn logs [flags] [APP_NAME|CONTAINER_NAME]
 ### Options
 
 ```
-  -f, --follow         Follow log output
-  -h, --help           help for logs
-  -s, --since string   Show logs since timestamp (e.g. 42m for 42 minutes)
-  -n, --tail int       Number of lines in log output
+  -f, --follow                     Follow log output
+  -h, --help                       help for logs
+      --include-proxy-containers   Also display logs from Linkerd proxy containers
```
ibuildthecloud, line 17: This shouldn't be an option.  If someone want's to "break through" the abstraction acorn they can just use kubectl.  We don't want to expose the internal details of the platform to the user.

pkg/log/log.go
```diff
@@ -384,7 +385,9 @@ func matchesPod(pod *corev1.Pod, options *Options) bool {
 }
 
 func matchesContainer(pod *corev1.Pod, container corev1.Container, options *Options) bool {
-	if options == nil || options.ContainerReplica == "" {
+	if options != nil && !options.IncludeProxyContainers && (container.Name == "linkerd-proxy" || container.Name == "linkerd-init") {
```
ibuildthecloud, line 388: Instead of hard coding these names we should check the containers against the "container spec". On each pod there is an annotation that has the v1.Container struct encoded in it.  The valid containers are, first, the value of `pod.Label[labels.AcornContainerName]` and then each key name in the Sidecars fields from the v1.Container.  An example of decoding this struct can be seen here.

https://github.com/acorn-io/acorn/blob/d12aeb70639466bad3396d7ba8b1e9dffaa05233/pkg/server/registry/apigroups/acorn/containers/translator.go#L95

docs/docs/100-reference/01-command-line/acorn_logs.md
g-linville: Removed. Through from playing around, I did figure out it is still technically possible for them to get the linkerd logs through acorn by running something like `acorn logs <app name>.<container name>.linkerd-proxy`. I could look for this case and explicitly block it if we want, but I don't know that that is worth doing.

ibuildthecloud: That's probably fine.

g-linville:
The best way to test this is to run the logging integration test in a cluster that has the acorn-linkerd-plugin installed. Due to a weird timing issue, I had to make a modification to this integration test in order for it to actually work with linkerd sidecars in place, which is explained in a comment in the code.

integration/log/log_test.go
```diff
@@ -33,7 +33,20 @@ func TestLog(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	app, err := c.AppRun(context.Background(), image.ID, nil)
+	app, err := c.AppRun(context.Background(), image.ID, &client.AppRunOptions{
```
StrongMonkey, line 36: Does this test actually verify something or you just want to enable linkerd annotation so you can manually verify the log when running in a cluster with linkerd installed?

pkg/log/log.go
```diff
@@ -384,16 +388,45 @@ func matchesPod(pod *corev1.Pod, options *Options) bool {
...
 
 func matchesContainer(pod *corev1.Pod, container corev1.Container, options *Options) bool {
-	if options == nil || options.ContainerReplica == "" {
-		return true
+	if options != nil && options.ContainerReplica != "" {
+		parts := strings.SplitN(options.ContainerReplica, ".", 3)
+		if len(parts) == 3 {
+			return pod.Name == parts[1] && container.Name == parts[2]
+		} else if len(parts) == 2 {
+			if pod.Labels[applabels.AcornContainerName] != "" {
```
StrongMonkey, line 396: Good catch with additional job filtering! Mind adding a small test to catch this behavior?

integration/log/log_test.go
g-linville: Yeah, I add the linkerd annotation to the app itself in this case because of the race condition that happens if I don't. The app will start running before the acorn-controller propagates the injection label from the project namespace to the app namespace, and as a result, the pods will start up without the linkerd sidecars. It's a bit weird so I left a comment in here to explain why I did it.

pkg/log/log.go
g-linville: @StrongMonkey I went ahead and added several tests for this `matchesContainer` function since its logic is a bit complex. 👍

StrongMonkey (approved):

Issue #1309, opened 2023-03-02
TestUsingComputeClasses flakes and fails


jacobdonenfeld:
```
--- FAIL: TestUsingComputeClasses (57.50s)
    --- FAIL: TestUsingComputeClasses/clustercomputeclass_unrestricted_default_gets_maximum (5.65s)
        run_test.go:1040: appinstances.internal.acorn.io "autumn-water" already exists
```

PR #1312, opened 2023-03-03
Ensure ComputeClasses and Apps created for ComputeClass tests are deleted before next test

tylerslaton:
We've been noticing some test flakes that occur as a result of some recent ComputeClass integration tests. This is because an app is created that already exists in the TestUsingComputeClasses test. In response, I am now ensuring that they are deleted between every test run.

Also, cleaning up any apps that are created between test case runs

#1309 

### Checklist
- [X] The title of this PR ends with a link to the main issue being address in paranthesis, like: `This is a title (#1216)`. [Here's an example](https://github.com/acorn-io/acorn/pull/1199)
- [X] All relevant issues are referenced in the PR description. *NOTE: don't use [GitHub keyworkds](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) that auto-close issues*

### DIFF
--- a/integration/run/run_test.go
+++ b/integration/run/run_test.go
@@ -18,13 +18,14 @@ import (
 	"github.com/acorn-io/acorn/pkg/labels"
 	"github.com/acorn-io/acorn/pkg/run"
 	"github.com/acorn-io/acorn/pkg/tolerations"
+	"github.com/acorn-io/baaah/pkg/router"
 	"github.com/stretchr/testify/assert"
 	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	crClient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 func TestVolume(t *testing.T) {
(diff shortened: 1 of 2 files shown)


integration/run/run_test.go
```diff
@@ -1036,16 +1037,28 @@ func TestUsingComputeClasses(t *testing.T) {
...
 					if err := kclient.Create(ctx, computeClass); err != nil {
 						t.Fatal(err)
 					}
+
+					// Clean-up and gurantee the computeclass doesn't exist after this test run
 					t.Cleanup(func() {
 						if err = kclient.Delete(context.Background(), computeClass); err != nil && !apierrors.IsNotFound(err) {
 							t.Fatal(err)
 						}
+						err := helper.EnsureDoesNotExist(ctx, func() (crClient.Object, error) {
```
thedadams, line 1054: I don't understand why this is necessary. If we delete it and that delete is successful, why do we have to ensure it doesn't exist?

tylerslaton: I'm trying to find a good source for this that backs me up here, but if I remember correctly deletes are asynchronous.  and thus deletes through the client are asynchronous. So the call to delete passes but it sticks around as things like network calls, finalizers, etc. finish up. The suite is failing because either apps or ComputeClasses are not getting deleted in time between test cases and collisions occur.

tylerslaton: It could also be that the resource exists in the cache still even though it has been deleted.

jacobdonenfeld (approved):
lgtm

integration/run/run_test.go
tylerslaton: For context - @thedadams and I talked about this offline. We tried getting around using the `EnsureDoesNotExist` function using a `PropogationPolicy` of `foreground` to no avail in the delete to no avail

thedadams (approved):

integration/run/run_test.go
tylerslaton: Talked offline further - this approach is fine given the circumstances.

PR #1314, opened 2023-03-03
Add region to project spec

thedadams:


jacobdonenfeld (approved):

pkg/cli/project_create.go
```diff
@@ -23,11 +23,17 @@ acorn project create acorn.io/username/new-project
...
+	// been registered for this flag. Not returning the error since neither of these is likely occur.
+	if err := cmd.RegisterFlagCompletionFunc("region", newCompletion(c.ClientFactory, regionsCompletion).complete); err != nil {
+		cmd.Printf("Error registering completion function for --region flag: %v\n", err)
+	}
 	return cmd
 }
 
 type ProjectCreate struct {
 	client ClientFactory
+	Region string `usage:"Region to create project in" short:"r"`
```
ibuildthecloud, line 36: Lets hold off on the short form and just have `--region` right now.  It's easier to add a short form later, but not remove it.

pkg/apis/internal.acorn.io/v1/appinstance.go
```diff
@@ -185,6 +190,7 @@ type AppInstanceStatus struct {
 type Defaults struct {
 	Volumes map[string]VolumeDefault `json:"volumes,omitempty"`
 	Memory  map[string]*int64        `json:"memory,omitempty"`
+	Region  string                   `json:"region,omitempty"`
```
StrongMonkey, line 193: Do we need to default Region concept in app too? Or it can just be `status.Region` and it will be calculated through spec.region, and fallback to project.default.region when spec.region is empty?

thedadams: This is the calculation of the default region based on the project default. The idea here is that by storing the default region on the app status, the region won't change for the app if an admin changes the default region for the project.

We don't want the app to be changed unless something in the app changes, not because something external (like the project default region) changed.

jsilverio22 (approved):
code lgtm from my understanding -- was able to pull down pr and see offerings regions / add & create w/ projects etc

StrongMonkey (approved):
LGTM

pkg/apis/internal.admin.acorn.io/v1/types.go
```diff
@@ -50,3 +50,41 @@ type ComputeClassMemory struct {
...
+
+type RegionInstance struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+
+	Spec   RegionInstanceSpec   `json:"spec,omitempty"`
+	Status RegionInstanceStatus `json:"status,omitempty"`
+}
+
+func (in *RegionInstance) ForRegion(region string) bool {
```
tylerslaton, line 64: I'm having a hard time understanding the need for this function. Could you help me understand?

integration/helper/namespace.go
```diff
@@ -19,6 +20,11 @@ func TempNamespace(t *testing.T, client client.Client) *corev1.Namespace {
 			GenerateName: "acorn-test-",
 			Labels: map[string]string{
 				"test.acorn.io/namespace": "true",
+				labels.AcornProject:       "true",
+			},
+			Annotations: map[string]string{
+				labels.AcornProjectDefaultRegion:    "local",
+				labels.AcornProjectSupportedRegions: "local",
```
tylerslaton, line 27: What if we want to test a Project with multiple supported regions - is that possible to do here or are we restricted to local?

pkg/client/project.go
```diff
@@ -28,12 +28,18 @@ func (c *DefaultClient) ProjectList(ctx context.Context) ([]apiv1.Project, error
...
 		ObjectMeta: metav1.ObjectMeta{
 			Name: name,
 		},
 	}
+	if region != "" {
+		project.Spec = apiv1.ProjectSpec{
+			DefaultRegion:    region,
+			SupportedRegions: []string{region},
+		}
+	}
```
tylerslaton, line 42: nit: This does not introduce any issues currently so I'm fine with it going in as is. That being said - what are your thoughts on this? This doesn't completely overwrite the `project.Spec` in the case that it ever has values that are non-region related in the future. 
```suggestion
	if region != "" {
		project.Spec.DefaultRegion = region
		project.Spec.SupportedRegions = []string{region}
	}
```

tylerslaton (commented):
Couple of moderately small comments. Looks good overall!

pkg/apis/internal.acorn.io/v1/appinstance.go
```diff
@@ -44,6 +44,10 @@ type AppInstance struct {
 	Status AppInstanceStatus `json:"status,omitempty"`
 }
 
+func (in *AppInstance) ForRegion(region string) bool {
```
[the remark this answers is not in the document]
tylerslaton: Resolving since this function doesn't exist anymore :)

pkg/apis/api.acorn.io/v1/types.go
```diff
@@ -365,18 +366,29 @@ type InfoList struct {
...
+	Placement     string `json:"placement,omitempty"`
+	Namespace     string `json:"namespace,omitempty"`
+	DefaultRegion string `json:"defaultRegion,omitempty"`
 }
 
 func (in *Project) NamespaceScoped() bool {
 	return false
 }
 
+func (in *Project) ForRegion(region string) bool {
```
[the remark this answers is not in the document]
thedadams: I am fine changing the names of these methods, but they have to have the same name because an interface is used in Hub to detect when objects are "regioned"

If it's OK with you, we can merge this and adjust the name in the next set of changes.

pkg/apis/internal.admin.acorn.io/v1/types.go
thedadams: It's used in Hub to determine that this is a region-scoped object and should be selectively synced.

integration/helper/namespace.go
thedadams: The local region is the only supported region in OSS.

tylerslaton:
When running locally I notice it is hard to tell what region an application is running in. Can/should we update the table for apps to display the region they're in?

pkg/client/project.go
thedadams: Yes, that's better.

pkg/cli/project_create.go
thedadams: I removed it from the run command and not here. I will remove here too. Thanks for catching it.

pkg/apis/internal.admin.acorn.io/v1/types.go
tylerslaton: Ah that does make sense. Thank you!

pkg/apis/api.acorn.io/v1/types.go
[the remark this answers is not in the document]
tylerslaton: Sounds good to me!

tylerslaton (approved):
Thanks for the changes! Looks good to me 👍🏻

PR #1317, opened 2023-03-06
Fix sidecar container exec (#1216)

rmodpur:
### Checklist

### Description
Suppose there is an app named `app` with one container named `web` that has a sidecar named `sidecar`

Before the fix these worked:
a. `acorn exec -c web app` (Note that app is not entered twice)
b. `acorn exec app.web-5c979bdb7b-vzq7j` (the extra characters come from the actual pod running in Kubernetes)

With the fix the below mentioned variations work as well:
c. `acorn exec app.web-5c979bdb7b-vzq7j.web`
d. `acorn exec -c web.sidecar app`
e. `acorn exec app.web-5c979bdb7b-vzq7j.sidecar`

The -c flag accepts the container name as defined in the Acornfile and not the Kubernetes pod name.

For #1216

### DIFF
--- a/pkg/server/registry/apigroups/acorn/containers/translator.go
+++ b/pkg/server/registry/apigroups/acorn/containers/translator.go
@@ -28,6 +28,7 @@ type Translator struct {
 	client kclient.Client
 }
 
+// name can be <app-name>+"."+<pod-name> or <app-name>+"."+<pod-name>+"."+<container-name>
 func (t *Translator) FromPublicName(ctx context.Context, namespace, name string) (string, string, error) {
 	for {
 		prefix, suffix, ok := strings.Cut(name, ".")
@@ -42,9 +43,13 @@ func (t *Translator) FromPublicName(ctx context.Context, namespace, name string)
 
 		name = suffix
 		namespace = app.Status.Namespace
+		// stop when name becomes <pod-name> or <pod-name>+"."+<container-name>
+		if strings.Count(name, ".") < 2 {
+			break
+		}
 	}
-
-	return namespace, name, nil
+	// return <pod-name> with <container-name> stripped if any
+	return namespace, strings.Split(name, ".")[0], nil
 }
 
 func (t *Translator) ListOpts(ctx context.Context, namespace string, opts storage.ListOptions) (string, storage.ListOptions, error) {
--- a/pkg/server/registry/apigroups/acorn/containers/translator_test.go
+++ b/pkg/server/registry/apigroups/acorn/containers/translator_test.go
@@ -1,0 +1,105 @@
+package containers
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	v1 "github.com/acorn-io/acorn/pkg/apis/internal.acorn.io/v1"
+	"github.com/acorn-io/acorn/pkg/scheme"
+	"github.com/acorn-io/baaah/pkg/router/tester"
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestFromPublicName(t *testing.T) {
+	app := &v1.AppInstance{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "app",
+			Namespace: "appNs",
+		},
+		Status: v1.AppInstanceStatus{
+			Namespace: "podNs",
+		},
+	}
+
+	tests := []struct {
+		containerPublicName      string
+		containerPublicNamespace string
+
+		expectedPodName      string
+		expectedPodNamespace string
+		expectedErr          error
+	}{
+		{
+			containerPublicName:      "app.pod",
+			containerPublicNamespace: app.Namespace,
+
+			expectedPodName:      "pod",
+			expectedPodNamespace: app.Status.Namespace,
+			expectedErr:          nil,
+		},
+		{
+			containerPublicName:      "app.pod.container",
+			containerPublicNamespace: app.Namespace,
+
+			expectedPodName:      "pod",
+			expectedPodNamespace: app.Status.Namespace,
+			expectedErr:          nil,
+		},
+
+		{
+			containerPublicName:      "nonExistingApp.pod.container",
+			containerPublicNamespace: app.Namespace,
+
+			expectedPodName:      "nonExistingApp.pod.container",
+			expectedPodNamespace: app.Namespace,
+			expectedErr:          fmt.Errorf("\"nonExistingApp\" not found"),
+		},
+		{
+			containerPublicName:      "app.pod.container",
+			containerPublicNamespace: "nonExistingNamespace",
+
+			expectedPodName:      "app.pod.container",
+			expectedPodNamespace: "nonExistingNamespace",
+			expectedErr:          fmt.Errorf("\"app\" not found"),
+		},
+
+		{
+			containerPublicName:      "app", // incorrect format
+			containerPublicNamespace: app.Namespace,
+
+			expectedPodName:      "app",
+			expectedPodNamespace: app.Namespace,
+			expectedErr:          nil,
+		},
+		{
+			containerPublicName:      "", // incorrect format
+			containerPublicNamespace: app.Namespace,
+
+			expectedPodName:      "",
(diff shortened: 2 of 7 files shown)


rmodpur:
@cjellick could you please review this ?

pkg/server/registry/apigroups/acorn/containers/translator.go
```diff
@@ -30,8 +30,9 @@ type Translator struct {
 
 func (t *Translator) FromPublicName(ctx context.Context, namespace, name string) (string, string, error) {
 	for {
-		prefix, suffix, ok := strings.Cut(name, ".")
-		if !ok {
+		prefix, suffix, _ := strings.Cut(name, ".")
+		// stop when name = <pod name> + "." + <container name>
```
jacobdonenfeld, line 34: Hesitant on this function, especially without seeing any unit tests covering the different ways the for loop can execute. Other then this, lgtm. Nice to see tests written and all passing!

rmodpur: I have added the unit tests

pkg/server/registry/apigroups/acorn/containers/translator.go
```diff
@@ -30,8 +30,9 @@ type Translator struct {
...
+		if strings.Count(suffix, ".") == 0 {
 			break
 		}
 		app := &v1.AppInstance{}
```
jacobdonenfeld, line 38: NIT: can refactor the for loop to 
```	// stop when name = <pod name> + "." + <container name>
	for strings.Count(name, ".") > 1 {
		prefix, suffix, _ := strings.Cut(name, ".")
		app := &v1.AppInstance{}
		err := t.client.Get(ctx, router.Key(namespace, prefix), app)
		if err != nil {
			return namespace, name, err
		}

		name = suffix
		namespace = app.Status.Namespace
	}
```

jacobdonenfeld (approved):
LGTM, one small nit. Thanks for adding tests! :)

pkg/server/registry/apigroups/acorn/containers/translator.go
rmodpur: Done
Thanks for the review !

tylerslaton (commented):
Tested these changes out locally and they seem to work great! 

Only thing I want to highlight is that `acorn exec -c` doesn't seem to work nicely with sidecars right now so I have to always manually select the sidecar from the interactive prompt.

rmodpur:
> Tested these changes out locally and they seem to work great!
> 
> Only thing I want to highlight is that `acorn exec -c` doesn't seem to work nicely with sidecars right now so I have to always manually select the sidecar from the interactive prompt.

What are you using as container name for `-c` ? the entire container public name or just the last part of it?
I tried `-c` with entire name(`<app-name>.<pod-name>.<container-name>`) it worked for both main & sidecar container.

tylerslaton:
Interesting, maybe there is some environment skew on my end? I'm seeing the following:

<details>

<summary>This works like I'd expect</summary>

  ```bash
  $ acorn containers bold-leaf            
  NAME                                     APP         IMAGE                                          STATE     RESTARTCOUNT   CREATED   MESSAGE
  bold-leaf.app-65dffdbd9c-9f92g.app       bold-leaf   public.ecr.aws/docker/library/nginx:latest     running   0              61m ago   
  bold-leaf.app-65dffdbd9c-9f92g.sidecar   bold-leaf   public.ecr.aws/docker/library/busybox:latest   running   0              61m ago    
  $ acorn exec bold-leaf.app-65dffdbd9c-9f92g.sidecar
  / #
  ```

</details>

<details>

<summary>This fails even though it seems that it shouldn't. Am I doing something wrong here?</summary>

```bash
# same containers seen in the first drop down
$ acorn exec -c bold-leaf.app-65dffdbd9c-9f92g.sidecar bold-leaf
  ✗  ERROR:  failed to find any containers for app bold-leaf
```

</details>

<details> 

<summary>As an aside, it would a nice UX if this worked for sidecars since it works for regular containers.</summary>

```bash
$ acorn exec -c app.sidecar bold-leaf
✗  ERROR:  failed to find any containers for app bold-leaf
$ acorn exec -c app bold-leaf
root@app:/#
```

</details>

rmodpur:
Your 1st & 2nd attempts should work, 3rd one should not work at all(for both sidecars & main container) with these changes
There were some changes in api server code as well, could you check docker image used for api server ?

tylerslaton (approved):
Looks like it must have been environment skew for me because the 2nd attempt started working. This looks good to me now, thank you for the contribution.

thedadams:
@rmodpur Are you saying that `acorn exec -c app bold-leaf` will no longer work with these changes?

rmodpur:
> @rmodpur Are you saying that `acorn exec -c app bold-leaf` should no longer work with these changes?

Yes, the assumption is the end user will only interact with acorn objects using it's public name which for the container is it's full name

thedadams:
That command works currently and should continue to work after these changes. Additionally, `acorn exec -c app.sidecar bold-leaf` should work, too, but we won't block merging this for this additional enhancement.

For now, I will dismiss the approvals so this change doesn't get merged accidentally until it is ensured that this command continues to work.

rmodpur:
Let's assume there are two containers for an app named `app`
1. `app.pod.main`
2. `app.pod.sidecar`

What are the variants that should work ?
a. `acorn exec app.pod.sidecar` (works now)
b. `acorn exec -c app.pod.sidecar app` (works now)
c. `acorn exec -c sidecar app` (doesn't work)

I understand that this PR may not be right place for it but I do have one suggestion, I think `-c` flag is redundant because the way exec is defined `exec [flags] APP_NAME|CONTAINER_NAME CMD` an user should be able to exec using these ways
1. `acorn exec` (will be asked for app first & then container)
2. `acorn exec app` (will be asked for a choice of container if more than one)
3.  `acorn exec app.pod.sidecar`

thedadams:
I think there is some confusion here on the what I am talking about. Let's suppose I have an app named `app` with one container named `web` that has a sidecar named `sidecar`.

Currently, in `main`, the following work:
a. `acorn exec -c web app` (Note that `app` is not entered twice)
b. `acorn exec app.web-5c979bdb7b-vzq7j` (the extra characters come from the actual pod running in Kubernetes)

These should continue to work with your changes.

The following do not work in `main` currently:
a. `acorn exec -c web-5c979bdb7b-vzq7j app`
b. `acorn exec app.web`
c. `acorn exec app.web-5c979bdb7b-vzq7j.web`
d. `acorn exec -c web.sidecar app`
e. `acorn exec app.web-5c979bdb7b-vzq7j.sidecar`

As I understand it, this change was to ensure e. works (and hopefully d. too).

I understand what you are saying about the `-c` flag feeling redundant. However, it is there for automated (CI/GitOps) interactions with Acorn where being given a selection to chose from doesn't work. Additionally, the `-c` flag accepts the container name as defined in the Acornfile and not the Kubernetes pod name. Ideally, the Kubernetes-ness is abstracted from the user as much as possible.

rmodpur:
@thedadams could you please review

Apart from the currently working ones you mentioned above these work with the fix
c. `acorn exec app.web-5c979bdb7b-vzq7j.web`
d. `acorn exec -c web.sidecar app`
e. `acorn exec app.web-5c979bdb7b-vzq7j.sidecar`

pkg/server/registry/apigroups/acorn/containers/translator_test.go
```diff
@@ -0,0 +1,105 @@
...
+			expectedErr:          nil,
+		},
+		{
+			containerPublicName:      "", // incorrect format
+			containerPublicNamespace: app.Namespace,
+
+			expectedPodName:      "",
+			expectedPodNamespace: app.Namespace,
+			expectedErr:          nil,
+		},
```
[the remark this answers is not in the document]
rmodpur: container's `FromPublicName` method is never called with `container.sidecar` pattern, in this case the cli code for `exec` lists all the containers for the `app` & matches  `container.sidecar` with `app.container-5c979bdb7b-vzq7j.sidecar` for example.

tylerslaton (approved):

thedadams (approved):

PR #1321, opened 2023-03-08
Cache golang setup and change fetch depth to speed up CI

jacobdonenfeld:
Total time saved: ~1 minute

Changed checkout depth from 0 to 1. 0 means to checkout all history for all branches and tags. 1 means to just checkout the latest commit and compress everything else. This seemed to speed up the checkout job by ~1 second. 

Updated setup_go action from v3 to v4, which came out last week. This enables caching by default, however it seemed like caching was already preset. There seems to be some other optimizations in the update but nothing obvious. 

### Checklist

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -17,8 +17,8 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
-          fetch-depth: 0
-      - uses: actions/setup-go@v3
+          fetch-depth: 1
+      - uses: actions/setup-go@v4
         with:
           go-version: "1.19"
       - uses: debianmaster/actions-k3s@v1.0.5
--- a/.github/workflows/aws-eks-test.yml
+++ b/.github/workflows/aws-eks-test.yml
@@ -22,7 +22,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
-      - uses: actions/setup-go@v3
+      - uses: actions/setup-go@v4
         with:
           go-version: "1.19"
       - run: make validate-code
--- a/.github/workflows/main-mac-smoke-test.yaml
+++ b/.github/workflows/main-mac-smoke-test.yaml
@@ -18,7 +18,7 @@ jobs:
         with:
           fetch-depth: 0
       - name: Set up Go
-        uses: actions/setup-go@v3
+        uses: actions/setup-go@v4
         with:
           go-version: "1.19"
       - name: Set up Docker Buildx
--- a/.github/workflows/main-release.yaml
+++ b/.github/workflows/main-release.yaml
@@ -22,7 +22,7 @@ jobs:
         with:
           fetch-depth: 0
       - name: Set up Go
-        uses: actions/setup-go@v3
+        uses: actions/setup-go@v4
         with:
           go-version: "1.19"
       - name: Set up QEMU
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -18,7 +18,7 @@ jobs:
         with:
           fetch-depth: 0
       - name: Set up Go
-        uses: actions/setup-go@v3
+        uses: actions/setup-go@v4
         with:
           go-version: "1.19"
       - name: Set up Docker Buildx
--- a/.github/workflows/validate-docs.yaml
+++ b/.github/workflows/validate-docs.yaml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
-      - uses: actions/setup-go@v3
+      - uses: actions/setup-go@v4
         with:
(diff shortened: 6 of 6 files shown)


cjellick:
Can you give a better more descriptive commit message body to this? Refer to the guidelines referenced in the "Commits follow [contributing guidance](https://github.com/acorn-io/acorn/blob/main/CONTRIBUTING.md#commits)" checkbox

.github/workflows/test.yaml
```diff
@@ -21,13 +21,14 @@ jobs:
       - uses: actions/setup-go@v3
         with:
           go-version: "1.19"
+          cache: true
```
cjellick, line 24: How about we upgrade to v4 of the plugin instead? It appears to enable cache by default

https://github.com/actions/setup-go#v4

.github/workflows/test.yaml
```diff
@@ -21,13 +21,14 @@ jobs:
...
       - uses: debianmaster/actions-k3s@v1.0.5
         with:
           version: 'v1.23.6-k3s1'
       - run: make validate-ci
       - run: make validate
       - run: make build
       - run: docker buildx install
-      - run: make setup-ci-image
+      - run: make setup-ci-image -j2
```
cjellick, line 32: curious why you added -j to just these two specific steps. And why you chose 2

.github/workflows/test.yaml
jacobdonenfeld: Oh wow, v4 came out 2 hours ago. That's convenient! Yes agreed

jacobdonenfeld: Turns out they added it in the readme, but haven't actually cut a release and published it. Couldn't resolve v4

.github/workflows/test.yaml
jacobdonenfeld: Removed from this step, turns out it wasn't necessary. Ran with -j2 later on because there are two makefile dependencies in make test, so it would run each of those in a separate thread.

tylerslaton (approved):

jsilverio22 (approved):

cjellick:
What's the actual speedup improvement here?

My understanding (could be wrong) is that go will test each package in parallel by default (up to the number of CPUs). This is really spread out across several go help pages though. The best easily found "evidence" is the conversations in the stack overflow: https://stackoverflow.com/questions/24375966/does-go-test-run-unit-tests-concurrently

I also compared the run of this PR vs one that doesn't have this change and they both about the same amount of time:
- https://github.com/acorn-io/acorn/actions/runs/4376662332
- https://github.com/acorn-io/acorn/actions/runs/4387143029

So, just wondering what we're gaining.

cjellick:
> Can you give a better more descriptive commit message body to this? Refer to the guidelines referenced in the "Commits follow [contributing guidance](https://github.com/acorn-io/acorn/blob/main/CONTRIBUTING.md#commits)" checkbox

I realize I'm being picky here, but now your commit title is overly long. Title should not execeed 50 chars per https://github.com/acorn-io/acorn/blob/main/CONTRIBUTING.md#commits:
> 2. Limit the subject line to 50 characters

It causes weird wrapping issues in various git log viewers (including github)

jsilverio22:
> What's the actual speedup improvement here?
> 
> My understanding (could be wrong) is that go will test each package in parallel by default (up to the number of CPUs). This is really spread out across several go help pages though. The best easily found "evidence" is the conversations in the stack overflow: https://stackoverflow.com/questions/24375966/does-go-test-run-unit-tests-concurrently
> 
> I also compared the run of this PR vs one that doesn't have this change and they both about the same amount of time:
> 
> * https://github.com/acorn-io/acorn/actions/runs/4376662332
> * https://github.com/acorn-io/acorn/actions/runs/4387143029
> 
> So, just wondering what we're gaining.

was playing around with these flags locally and am not seeing much difference either changing num of cpus adding `t.Parallel()`

cjellick (changes requested):
left a few comments i'd like to hear back on before merging

tylerslaton:
For whatever reason [v4](https://github.com/actions/setup-go/releases/tag/v4.0.0) of the `setup-go` isn't showing up 🤔

jacobdonenfeld:
> ~For whatever reason v4 of the `setup-go` isn't showing up 🤔~ This was because of a Github outage.
> 
> https://github.com/actions/setup-go/releases/tag/v4.0.0

They added the release to their readme but didn't tag it at the time I looked into it. https://github.com/actions/setup-go/issues/353

Its out now and works!

jacobdonenfeld:
> My understanding (could be wrong) is that go will test each package in parallel by default (up to the number of CPUs)

I also see this written as the case, however, it seems like the integration tests would be failing if they were being ran in parallel, as they live in multiple packages. (One package for images, one for credentials, etc.)

Will need to play around with this a little bit more and figure out what is going on.

Issue #1323, opened 2023-03-09
Sometimes building/running an image in a new project fails with `websocket: bad handshake`


thedadams:
Retrying the build/run command immediately after this failure usually succeeds, indicating that the build server is not actually ready. This will only happen in Acorn if `--publish-builders` installation flag is passed.

sangee2004:
Tested with `acorn version v0.6.0-12-g5e4b75d0+5e4b75d0`
Have  acron installed with `--publish-builders` flag enabled.

Creating project 
Deploy app.
Repeat the above step multiple times.
App deployment succeeds and no websocket errors are seen . But there is a delay of ~30 secs every time for build process to start when apps are deployed - https://github.com/acorn-io/acorn/issues/1332.

PR #1324, opened 2023-03-09
Change to HTTP readiness probe for build-server (#1323)

thedadams:
The TCP probe tends to have issues, especially when running with linkerd. This change switches the probe to an HTTP GET probe, which is already enabled in the build-server.

A special note about this change: since this is changing the deployment spec for the build-server, all currently running build-servers will be restarted once this fix is pushed out to users.

If redeploying all currently running build-servers is not acceptable, then altering these changes to take affect only when the `BuilderInstance` generation changes would be possible, however, would be more involved. I pushed it this way because of the simplicity, but am happy to implement the more involved logic.

### Checklist

Issue: https://github.com/acorn-io/acorn/issues/1323

### DIFF
--- a/pkg/imagesystem/buildertemplate.go
+++ b/pkg/imagesystem/buildertemplate.go
@@ -25,9 +25,9 @@ func BuilderObjects(name, namespace, forNamespace, buildKitImage, pub, privKey, 
 				{
 					Name:     system.BuildKitName,
 					Protocol: corev1.ProtocolTCP,
-					Port:     int32(system.BuildkitPort),
+					Port:     system.BuildkitPort,
 					TargetPort: intstr.IntOrString{
-						IntVal: int32(system.BuildkitPort),
+						IntVal: system.BuildkitPort,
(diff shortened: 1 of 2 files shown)


StrongMonkey:
I think restarting buildkit service is fine for now. Have we make sure that once we switch to this, simply watching deployment.status.replicas > 0 would be sufficient to wait for buildkit to be up?

thedadams:
@StrongMonkey Even with this change, we still need to do the `PingBuilder` call. It is possible that Acorn recognizes that the pod is ready and tries to move on before Kubernetes adds the pod to the service load balancer. I saw a few instances where the first GET call in `PingBuilder` failed with a timeout error, but the second one passed.

The difference between the readiness probe and `PingBuilder` is that `PingBuilder` uses the endpoint and the readiness probe uses the pod IP.

ibuildthecloud (approved):

cjellick (approved):

Issue #1335, opened 2023-03-13
Validate the default project when acorn commands are run


jsilverio22:
Context: from @iwilltry42 
```
So I got super confused today.
I have the following setup:
My homelab K3s cluster -> Installed some development version of Acorn and created the testsignature project which I used
A local k3d cluster -> Installed another development version of Acorn
What happened:
Switched from homelab to k3d cluster using kubectx
Uninstalled acorn and re-installed it
acorn pulled some image -> got an error on that
Inspected the registry and figured, that it actually pulled the image into the acorn/testsignature repository :eyes:
Did a hundred checks that indeed I was using the correct cluster and didn't create the same project there, re-created the cluster, re-installed Acorn, re-build acorn locally, etc.
Lastly, I switched to the homelab cluster, deleted the testsignature cluster, switched back et voilà, the image pulled into acorn/acorn
Now I finally found out, that Acorn keeps track of the currently used project in ~/.config/acorn/config.yaml, but doesn't seem to check if that project is available in the currently connected cluster 
Now I'm wondering if we should keep track of the selected project per kube context or check if the selected project exists in the new cluster?
Note, that acorn projects in the k3d cluster gave me an empty response, so from the CLI I couldn't have known what's going on.
```

sangee2004:
Able to reproduce the problem with following steps:
1. Created project p1 and made it default project. 
2. Uninstalled acorn. 
3. Also deleted project namespace -p1.
4. Install acorn.
5. acorn pull sangeetha/myfirstacorn:v0.0.1
 [55407181/55444378] █████████████████████████████████████████████  100% | 9m10s
  ✗  ERROR:  images.api.acorn.io "ee8a243e3e7a613254925f0927ab6832a15a5e233c1f3c03ac4e26e08dacedd1" not found
sangeethahariharan@Sangeethas-MBP hub % acorn images
REPOSITORY   TAG       IMAGE-ID


Tested with acorn version - `v0.6.0-21-gb1b07672+b1b07672`

When following the same steps as above , user is presented with error message that project is not found:

In this case default project was set to p2 before acron was uninstalled.
Also the namespace p2 was deleted before acorn was installed.

```
% acorn projects 
NAME                                     DEFAULT
acorn                                    
hub.stg.acornlabs.com/hub-e2e/acorn      
hub.stg.acornlabs.com/hub-e2e/admin      
hub.stg.acornlabs.com/hub-e2e/readonly   

[...]

PR #1336, opened 2023-03-13
add check to project default to ensure it exists (#1335)

jsilverio22:
### Checklist
#1335

### DIFF
--- a/pkg/cli/acorn.go
+++ b/pkg/cli/acorn.go
@@ -120,24 +120,30 @@ func (a *Acorn) PersistentPre(cmd *cobra.Command, args []string) error {
 			return err
 		}
 	}
-
+	cfg, err := config.ReadCLIConfig()
+	if err != nil {
+		return err
+	}
+	clientFactory := CommandClientFactory{
+		cmd:   cmd,
+		acorn: a,
+	}
+	//check default project to ensure it exists in the current context
+	//except for project cmds to allow project use <valid project> to succeed
+	if a.Project == "" && !isProjectCmd(cmd) {
+		a.Project = cfg.CurrentProject
+	}
 	if a.Project != "" {
-		clientFactory := CommandClientFactory{
-			cmd:   cmd,
-			acorn: a,
-		}
-		cfg, err := config.ReadCLIConfig()
-		if err != nil {
-			return err
-		}
-
 		err = project.Exists(cmd.Context(), clientFactory.Options().WithCLIConfig(cfg), a.Project)
 		if err != nil {
 			return err
 		}
 	}
-
 	return nil
+}
+
+func isProjectCmd(cmd *cobra.Command) bool {
+	return (cmd.Parent() != nil && cmd.Parent().Name() == "project") || (cmd.Name() == "project")
 }
 
 func (a *Acorn) Run(cmd *cobra.Command, args []string) error {


pkg/cli/acorn.go
```diff
@@ -120,21 +120,21 @@ func (a *Acorn) PersistentPre(cmd *cobra.Command, args []string) error {
...
-		}
+	cfg, err := config.ReadCLIConfig()
+	if err != nil {
+		return err
+	}
+	clientFactory := CommandClientFactory{
+		cmd:   cmd,
+		acorn: a,
+	}
+	//check default project to ensure it exists in the current context
```
jacobdonenfeld, line 131: NIT: move this comment above the project.Exists() check

pkg/cli/acorn.go
```diff
@@ -120,21 +120,21 @@ func (a *Acorn) PersistentPre(cmd *cobra.Command, args []string) error {
...
+	if a.Project == "" {
+		a.Project = cfg.CurrentProject
+	}
```
jacobdonenfeld, line 134: Seems like this PR is adding tolerance for a.Project == "". When would this happen?

jacobdonenfeld (commented):
So previous behavior when default project wasn't set was to succeed if acorn.Project == "".  I see now that if it's set to "", set acorn.Project = cfg.CurrentProject.

What if currentProject == ""? When would currentProject be set and acorn.Project not be?

pkg/cli/acorn.go
jsilverio22: so a.Project is only set when -j/--project is used in a cmd 

from the disscusion in standup @iwilltry42 was running into the issue where if you have a project default set and swap kubeconfigs you will not be alerted that the previously set default does not exist in that new context

this change will not check if no project is specified in the -j/--project flag for a cmd to check the currently set default project to make sure it exists

tylerslaton (approved):
Think we should add a test case to capture this?

jsilverio22:
> Think we should add a test case to capture this?

yeah I'm trying to think of if it can be tested and if so what test makes the most sense. mainly because currently the way to get it to repro is create a project then swap kubecontexts so the project dne in that context so thinking of a way to replicate that through our different testing Envs

tylerslaton:
I'd say this is definitely non-essential so don't spend too much time if we need this soon. That being said...

I think we can do this with a CLI test that specifies an environment variable `KUBECONFIG` iside of the `testdata` directory and test with that. Looking at `baaaah` [it respects this](https://github.com/acorn-io/baaah/blob/282f5388cca5aa36b347f26ef3fc918f45100334/pkg/restconfig/restconfig.go#LL40). That way we could at least test the failure path (which we want in this case). I'm unsure if we'll run into any tricky-ness verification that the KUIBECONFIG is connected to a real cluster or not though.

pkg/cli/acorn.go
jacobdonenfeld: Awesome, and then it would hit that error condition. Makes sense! Seems like a test could be written in a github workflow most easily, but would be funky to have a single test written in there. Maybe a new job inside of test.yml for this?

jacobdonenfeld (approved):

iwilltry42:
Tested:

```bash
$ acorn projects                          
NAME               DEFAULT
acorn              *
nosignaturerules   

$ acorn project use nosignaturerules

$ acorn projects                    
NAME               DEFAULT
acorn              
nosignaturerules   *

# Switching Context
$ kubectx k3d-acorn                       
Switched to context "k3d-acorn".

$ acorn projects
NAME      DEFAULT
acorn     

$ acorn pull ghcr.io/acorn-io/library/hello-world:latest
  ✗  ERROR:  projects.api.acorn.io "nosignaturerules" not found
```

While this prevents the original issue, it may still be weird for some clueless user, right?
I.e. in general, the user has no chance of knowing that `nosignaturerules` is the currently selected project (after switching to the second cluster). In this specific case, there's also no other choice than using the `acorn` project in the second cluster. 

I honestly don't know a perfect solution for this, so I think this is good enough as it covers the basics and should push every user in the right direction :+1:

iwilltry42 (approved):

iwilltry42:
@jsilverio22 this conflicts with `acorn install` unfortunately :confused:

jsilverio22:
> @jsilverio22 this conflicts with `acorn install` unfortunately 😕

oh how so?

iwilltry42:
Same scenario as in my previous test case:

```bash
# current project: foo

$ kubectx cluster-without-acorn

$ acorn install          
  ✗  ERROR:  the server could not find the requested resource (get projects.api.acorn.io foo)
```

jsilverio22:
> Same scenario as in my previous test case:
> 
> ```shell
> # current project: foo
> 
> $ kubectx cluster-without-acorn
> 
> $ acorn install          
>   ✗  ERROR:  the server could not find the requested resource (get projects.api.acorn.io foo)
> ```

ahh and if it's not installed can't change the project should we check for install command and bypass the default check logic if it is as well?

iwilltry42:
I assume that there may be multiple commands that are "project agnostic" and won't need (or like with install "should not have") this check? :thinking:

Issue #1185, opened 2023-02-02
`acorn images -A` lists images only from the first project in `acorn projects`.


sangee2004:
Server version - acorn version v0.5.1-5-g28a7873+28a78730

Steps to reproduce the problem:
Have few projects created with images in each of the projects. 
p1 - image1
p2 - image2

`acorn images -a` lists images only from the first project in `acorn projects`.
in the above case , only `image1` will be listed.

```
% acorn images -a -A
REPOSITORY   TAG       IMAGE-ID
<none>       <none>    d5bae66abc18
<none>       <none>    2c7ba01b4cc9

 % acorn projects
NAME           DEFAULT   DESCRIPTION
mynewproject             
acorn          *         
mytest2                  
test                     

% acorn images -a -j mynewproject 
REPOSITORY   TAG       IMAGE-ID
<none>       <none>    d5bae66abc18
<none>       <none>    2c7ba01b4cc9
 % acorn images -a -j acorn       
REPOSITORY                              TAG       IMAGE-ID
<none>                                  <none>    d5bae66abc18
<none>                                  <none>    371cd2fcba06
<none>                                  <none>    17f2d0dfba2e
<none>                                  <none>    ac725dc8ea46
<none>                                  <none>    b32144f8faa1
index.docker.io/sangeetha/mytestacorn   v0.0.2    d02d3c8d9a22
```

Also , In the output for ` acorn images -A ` , we should be able to show which project the images belong to  like `<project-name>/<image-id>`

cjellick:
Based on out-of-band convo, this will probably get addressed holistically as part of the PR for project::resource. I'm going to move this issue back to "in progress" and it can be moved forward when that PR lands.

cjellick:
moving out to v0.8.0

cjellick:
We got rid of -A as a flag for `acorn images`. Its now just a command specific flag (i dont know all the commands that support it, but for `acorn ps` does)

PR #1341, opened 2023-03-14
fix acorn images -A  to list all images

jacobdonenfeld:
### Checklist
- [ ] The title of this PR ends with a link to the main issue being address in paranthesis, like: `This is a title (#1216)`. [Here's an example](https://github.com/acorn-io/acorn/pull/1199)

Addresses  #1185

Acorn images -A now lists all images


cjellick:
this isnt passing the lint check @jacobdonenfeld

jsilverio22 (approved):

jsilverio22 (commented):
seems to be failing tests still

pkg/cli/images_test.go
```diff
@@ -165,6 +165,40 @@ func TestImage(t *testing.T) {
 			wantErr: false,
 			wantOut: "found-image1234567\nfound-image-no-tag\nfound-image-two-tags1234567\nfound-image-two-tags1234567\n",
 		},
+		{
+			// TODO: doesn't adiquetly test yet.
```
cjellick, line 169:
```suggestion
			// TODO: doesn't adequately test yet.
```

cjellick:
@sangee2004 mentioned in the original issue:

> Also , In the output for acorn images -A , we should be able to show which project the images belong to like <project-name>/<image-id>

So we need to add in some project info to the output when -A is added, to disambiguate images. See how when I list all images, I can't tell what project they belong to:
```
$ acorn images -A
REPOSITORY   TAG       IMAGE-ID
local1       latest    e7e61e7913c2
asdf         latest    d3d69cf97d84
asd2         latest    dc958d0bed20
zip/bar      1         dc958d0bed20
```

For `acorn ps`, we handle that by prepending the project:
```
acorn $ acorn ps -A
NAME                           IMAGE                                  HEALTHY   UP-TO-DATE   CREATED    ENDPOINTS                                                                                MESSAGE
acorn.io/cjellick/prod1/asdf   ghcr.io/acorn-io/library/hello-world   1         1            128m ago   https://webapp-asdf-04228a1a.a-cjellick-prod1-8b8d8b57-steve1.on-acorn.io => webapp:80   OK
```
See how that differs from the same output with no -A:
```
acorn $ acorn ps
NAME      IMAGE                                  HEALTHY   UP-TO-DATE   CREATED    ENDPOINTS                                                                                MESSAGE
asdf      ghcr.io/acorn-io/library/hello-world   1         1            129m ago   https://webapp-asdf-04228a1a.a-cjellick-prod1-8b8d8b57-steve1.on-acorn.io => webapp:80   OK
```

The only weird thing for images is that prepending the project will look odd if the image already has a remote registry. For example, consider this image list:
```
$ acorn images -A
REPOSITORY                             TAG       IMAGE-ID
index.docker.io/cjellick/simple-demo   1         1e9545cc6f3c
local1                                 latest    e7e61e7913c2
```
I don't think this will look great:
```
REPOSITORY                                                         TAG       IMAGE-ID
acorn.io/cjellick/prod1/index.docker.io/cjellick/simple-demo       1         1e9545cc6f3c
acorn/local1                                                      latest    e7e61e7913c2
```
So, let's add an extra PROJECT column when -A is added:
```
PROJECT                           REPOSITORY                                                         TAG       IMAGE-ID
acorn.io/cjellick/prod1       index.docker.io/cjellick/simple-demo       1         1e9545cc6f3c
acorn                            local1                                                      latest    e7e61e7913c2
```
Forgive the bad spacing above, just doing it quickly

cjellick (changes requested):
see my comment

pkg/apis/internal.acorn.io/v1/imageinstance.go
```diff
@@ -34,9 +34,10 @@ type ImageInstance struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
 
-	Repo   string   `json:"repo,omitempty"`
-	Digest string   `json:"digest,omitempty"`
-	Tags   []string `json:"tags,omitempty"`
+	Project string   `json:"project,omitempty"`
```
jacobdonenfeld, line 37: Open question. Is it fine to be adding another field to Image and ImageInstance? Translation between pub/priv works fine, and it is omitempty for json parsing.

pkg/tables/tables.go
```diff
@@ -56,7 +56,8 @@ var (
 
 	// Used for acorn image related printing
 	ImageAcorn = [][]string{
-		{"Repository", "{{if eq .Repository \"\"}}<none>{{else}}{{.Repository}}{{end}}"},
+		// 		{"Repo", "{{if and (not .Repo) (not .Project) }}<none>{{ else if and (not .Project) .Repo }}{{ .Repo }}{{ else if and (not .Repo) .Project }}{{ .Project }}{{ else }}{{ .Project }}::{{ .Repo }}{{ end }}"},
+		{"Repo", "{{if (not .Repo)}}<none>{{else if .Project }}{{.Project}}::{{end}}{{ .Repo }}"},
```
jacobdonenfeld, line 60: Could have moved this logic into the images.go cli code, but figured it made more sense to keep it logically separated and compile together here.

pkg/apis/api.acorn.io/v1/types.go
```diff
@@ -96,9 +96,10 @@ type Image struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
 
-	Repo   string   `json:"repo,omitempty"`
-	Digest string   `json:"digest,omitempty"`
-	Tags   []string `json:"tags,omitempty"`
+	Project string   `json:"project,omitempty"`
```
cjellick, line 99: This is probably a no-go. I don't think any other types expose Project as a field like this.

pkg/apis/internal.acorn.io/v1/imageinstance.go
cjellick: yeah, as mentioned above, putting project on the object is probably problematic in general

Makefile
```diff
@@ -7,9 +7,6 @@ tidy:
 generate:
 	go generate
 
-mocks:
```
thedadams, line 10: Why is this being removed?

integration/client/images/images_test.go
```diff
@@ -410,3 +413,84 @@ func TestImageBadTag(t *testing.T) {
 	assert.Equal(t, "repository can only contain the characters `abcdefghijklmnopqrstuvwxyz0123456789_-./`: foo@@", err.Error())
 
 }
+
+func TestImageListBuildTwoImagesAssertMultiClientLists(t *testing.T) {
```
thedadams, line 417: Is there something that is being tested here that requires this be an integration test? Can this be a unit test instead?

pkg/cli/images.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"strings"
 
+	"github.com/acorn-io/acorn/pkg/labels"
+
```
thedadams, line 8: nit: remove line

pkg/cli/images.go
```diff
@@ -127,52 +138,57 @@ func (a *Image) Run(cmd *cobra.Command, args []string) error {
...
-				Container: imageContainer.Container,
-				Repo:      imageContainer.Repo,
-				Digest:    imageContainer.Digest,
-				ImageID:   imageContainer.ImageID,
+				Container:  imageContainer.Container,
+				Repository: imageContainer.Repository,
+				Digest:     imageContainer.Digest,
+				ImageID:    imageContainer.ImageID,
+			}
+			if allProjects { // Same project for all container images
```
thedadams, line 187: nit: is this comment correct?

pkg/client/image_test.go
```diff
@@ -0,0 +1,323 @@
...
+	ns := corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: config.projectName,
+			Labels: map[string]string{
+				"test.acorn.io/namespace": "true",
+				labels.AcornProject:       "true",
+			},
+		},
+		Spec:   corev1.NamespaceSpec{},
+		Status: corev1.NamespaceStatus{},
```
thedadams, line 41: nit: not needed.

pkg/client/image_test.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+
+	imageListObj := config.images
+	testingScheme := scheme2.Scheme
+	err := scheme2.AddToScheme(testingScheme)
+	if err != nil {
+		return client.DefaultClient{}, err
+	}
```
thedadams, line 49: Not needed. This is done when you import the package.

pkg/client/image_test.go
```diff
@@ -0,0 +1,323 @@
...
+	testcontrollerclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+type defaultClientImagesConfig struct {
+	images      v12.ImageList
+	projectName string
+	namespace   string
+}
+
+func createMockedDefaultClientImageLister(t *testing.T, config defaultClientImagesConfig) (client.DefaultClient, error) {
```
thedadams, line 30: Now that I look at this, I'm not sure this is needed. Why doesn't the `gomock` package work here?

pkg/client/multi.go
```diff
@@ -5,6 +5,8 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/acorn-io/acorn/pkg/labels"
+
```
thedadams, line 9: nit: remove line.

Makefile
```diff
@@ -7,9 +7,6 @@ tidy:
...
-	go run github.com/golang/mock/mockgen --build_flags=--mod=mod -destination=./pkg/mocks/mock_client.go -package=mocks github.com/acorn-io/acorn/pkg/client Client,ProjectClientFactory
-
```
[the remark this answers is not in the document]
jacobdonenfeld: our mocks run from go generate. Craig added a second command to generate mocks for another library as well. This is redundant and not up to date.

pkg/client/image_test.go
jacobdonenfeld: It does work. Didn't want to mock a k8's library when k8 provided their own fake library for it. Figured I'd run into random versioning issues or something else.

I see we are now mocking the server k8s library, but I feel using k8's fake library is more robust. I could be completely wrong.

pkg/client/image_test.go
jacobdonenfeld: You're right. Thanks!!

pkg/cli/images.go
jacobdonenfeld: I believe so. If we grab an image, and call `imgDetails := ImageDetails(image)`, following `imgDetails.AppImage.ImageData.(Containers|Jobs) `should be a part of the same project as the original image.

pkg/client/image_test.go
cjellick: We've discussed this in a few different places. Let's not introduce k8s's fake library if we can avoid it.

It seems fine and would address the use case, but I want to limit the number of different ways we mock things. @ibuildthecloud pointed out at that baaah has a test client that can accomplish this and it is what is used behind the scenes for all our yaml-based tests.

Here is an example usage:
https://github.com/acorn-io/acorn/blob/f6ce48cd568bc3623e3e343f62e0811a83672945/pkg/ref/expr_test.go#L216-L219

pkg/ref/expr_test.go
```diff
@@ -38,7 +38,8 @@ func TestResolve(t *testing.T) {
 			ObjectMeta: metav1.ObjectMeta{
 				Name: "app-namespace",
 				Labels: map[string]string{
-					labels.AcornAppName:      "app-name",
+					labels.AcornAppName: "app-name",
+					// Does the below not need to match "app-namespace"?
```
thedadams, line 42: The answer to this question is no. The namespace is called `app-namespace` and it corresponds to the `project-name`, which is the namespace where the `App` object lives.

Let's remove this comment.

pkg/cli/images.go
```diff
@@ -127,52 +137,57 @@ func (a *Image) Run(cmd *cobra.Command, args []string) error {
...
 }
 
 type imageContainer struct {
-	Container string
-	Repo      string
-	Tags      []string
-	Digest    string
-	ImageID   string
+	Container  string
+	Repository string
```
thedadams, line 149: Why does this need to change?

pkg/ref/expr_test.go
jacobdonenfeld: Perfect, thank you for the info! Removed

pkg/cli/images.go
jacobdonenfeld: Naming consistency with imagePrint and imageContainerPrint. Could change it back if desired.

pkg/client/image_test.go
jacobdonenfeld: Can't currently baaah's client to work as kclient.withWatch because it doesn't implement Watch, so can't use that as the base client underneath defaultclient.

Is it worthwhile for me to modify baaah to be able to use it, or to use what is currently written?

pkg/client/multi.go
```diff
@@ -372,11 +382,17 @@ func (m *MultiClient) ImageTag(ctx context.Context, image, tag string) error {
...
 func (m *MultiClient) ImageDetails(ctx context.Context, imageName string, opts *ImageDetailsOptions) (result *ImageDetails, err error) {
-	c, err := m.Factory.ForProject(ctx, m.Factory.DefaultProject())
+	// Image may exist on any project within MultiClient
+	clients, err := m.Factory.List(ctx)
 	if err != nil {
 		return nil, err
 	}
-	return c.ImageDetails(ctx, imageName, opts)
+	for _, client := range clients {
+		if details, err := client.ImageDetails(ctx, imageName, opts); err == nil {
```
cjellick, line 391: This is causing the panic I shared in slack. In my case, that ImageDetails call returns an error, but we arent returning the error, we just fall out of the loop eventually and when we get to line 395, the `err` returned there is not the same err defined on this line

jacobdonenfeld: fixed, ty!

thedadams:
The only thing I can think of is that the following two commands don't work yet:
```
 acorn -A images thedadams/my-tag:latest
```
```
 acorn -A images 0f9225144e1c
```
Spoke with @jacobdonenfeld "offline" and he agreed it would be good to get this working in this change.

jacobdonenfeld:
Paused merging this in until parsing project::resource is completed.
Will also aim to implement what Donnie said above.

Issue #456, opened 2022-08-03
Feature request: Support network policies


olljanat:
As Acorn can be used to deploy multiple containers as bundle it it would be nice to have also option to generate network policy which isolates application from others.

ibuildthecloud:
Yes, this absolutely will come.  The design of acorn is such that all network communication paths are known and have some logical boundary.  We will either be generating network policies and/or service mesh config (istio/linkerd) to enforce this.  This is key future feature for the security posture of Acorn.  One of the great things about Acorn is we know so much about the application behavior that we can create good implicit rules.

ibuildthecloud:
Researching this a bit, network policies would be extremely simple.  LInkerd looks straight forward too and Istio is rocket science like normal.  So we will start with NetworkPolicies and go from there.

olljanat:
In case of one app with multiple containers it should be very simple yes. In case of multiple apps you will need also figure out some nice way to refer another app in Acornfile bur perhaps even that can be just name reference.

g-linville:
## Design Document

### Requirements

- Traffic between different projects is blocked by default
- Traffic to published ports is allowed from all sources

### UX and API changes

- New flag for `acorn install` that disables the creation of NetworkPolicies (something like `--disable-network-policies`).
  - This might also result in a slight API change for the struct that handles these installation settings. I haven't looked at this part of the code yet.

### Visualization

![image](https://user-images.githubusercontent.com/53102776/222276250-c4c567c5-5481-48c5-ae44-07c107ad5256.png)

### Additional considerations

- Backwards compatibility: if we turn on NetworkPolicies by default (and I think we should), this has the potential to break some users' existing configurations, if they were relying on applications in separate projects to be able to talk to each other on non-published ports. We would definitely need to warn users about this.
- Tests: I should be able to write tests for the changes I have already made in this draft PR. I haven't written tests for a new handler yet but I'll reference the other ones and I imagine it will be relatively straightforward. I can also add an integration test to truly deploy two apps in separate projects and try to have one ping or curl the other, if that is something we think is worthwhile.
- Time estimate: I think this should be a fairly easy feature to add that we can easily land in the next release.

cjellick:
> New flag for acorn install that disables the creation of NetworkPolicies (something like --disable-network-policies).

I am wondering if we shou;d we have it disabled by default and you need to explicitly enable it?

> Traffic between different projects is blocked by default

[...]

PR #1352, opened 2023-03-17
Redesign NetworkPolicies (#456)

g-linville:
re: #456

This redesigns the approach to NetworkPolicies. Previously I tried to examine the app spec and create every needed policy based on this. This required me to know exactly how the publishing logic was going to work (which Ingress resources and LoadBalancer Services it was going to create, for example) and then create NetPols based on that.

Instead, I am now creating the main app-wide policy based on the AppInstance, and additional policies for Ingress resources and LoadBalancer Services created for published ports. In this way, the NetworkPolicy handlers are just "responding" to the publish logic that is handled by a different part of the controller. I don't need to predict which ingresses and services will be created, I just look at them and then create the policy to allow access to the port that they are exposing.

In addition to `--network-policies` and `--ingress-controller-namespace` from the first time around, this adds another installation option, `--allow-traffic-from-namespace` (repeatable), which allows the admin to specify namespaces that should be allowed to send traffic to all Acorn apps. This is useful for monitoring for example.

### Checklist


pkg/controller/appdefinition/networkpolicy.go
```diff
@@ -1,25 +1,235 @@
...
+
+	for svcName, ports := range svcNameToPorts {
+		// get the Service from k8s
+		svc := corev1.Service{}
+		err = req.Get(&svc, ingress.Namespace, svcName)
+		if err != nil {
+			// is this okay to do? sometimes the Service doesn't exist yet when we need to get it
+			// so errors might occasionally appear in the logs but things eventually settle and work
+			return err
+		}
```
g-linville, line 88: Wanted to leave a quick comment on this for reviewers - is this the right approach here? The service will exist eventually, so I need this request to be re-queued, but I'm not sure if there's another way to do it besides just returning an error.

tylerslaton: Instead of returning an error, you could set `resp.RetryAfter()` and return `nil` here.

g-linville: Thanks! Didn't know that was a thing but that is perfect for this situation.

tylerslaton (approved):

docs/docs/100-reference/01-command-line/acorn_install.md
```diff
@@ -21,25 +21,29 @@ acorn install
...
       --controller-replicas int                acorn-controller deployment replica count
+      --disable-network-policies               Disable the creation Kubernetes NetworkPolicies which block cross-project network traffic (default false)
   -h, --help                                   help for install
       --http-endpoint-pattern string           Go template for formatting application http endpoints. Valid variables to use are: App, Container, Namespace, Hash and ClusterDomain. (default pattern is {{hashConcat 8 .Container .App .Namespace | truncate}}.{{.ClusterDomain}})
       --ignore-user-labels-and-annotations     Don't propagate user-defined labels and annotations to dependent objects
       --image string                           Override the default image used for the deployment
       --ingress-class-name string              The ingress class name to assign to all created ingress resources (default '')
+      --ingress-controller-namespace string    The namespace where the ingress controller runs - used to secure published HTTP ports with NetworkPolicies.
```
cjellick, line 38: did we agree on this explicitly in a design meeting? I dont recall. Just want to make sure we agreed it was the right approach

cjellick:
I am not sure about these install flags:

> In addition to --disable-network-policies, --pod-cidr, and --ingress-controller-namespace from the first time around, this adds another installation option, --allow-traffic-from-namespace (repeatable), which allows the admin to specify namespaces that should be allowed to send traffic to all Acorn apps. This is useful for monitoring for example.

I want to ask the rest of the team on monday if they all seem reasonable

docs/docs/100-reference/01-command-line/acorn_install.md
g-linville: After talking through it I think we came to the conclusion that we did not want to require the admin to provide this information but wanted to make it optional, which is why I created the install parameter.

g-linville:
At standup today we decided to have a `--network-policies` flag instead of `--disable-network-policies` in order to avoid double negatives. I also learned that the pod CIDR(s) is/are determinable based on the node specs, so I removed `--pod-cidr` as an installation option.

cjellick:
@g-linville once this is merged, let's make sure we get hub setup correctly with the new parameters.

Keep in mind that dev hub would get updated within an hour (probably less) and staging would get updated at night.

We want hub to use the most-secure version of the feature as possible, so I think that means we will want to lock down the published ports->ingress flow and that we'll need to whitelist any other namespaces that need to talk to app namespaces (like youve point out, that's most likely monitoring).

I'm sure another eng or ops person can point you at the right places to put in PR to set the flags properly for hub.

docs/docs/100-reference/01-command-line/acorn_install.md
```diff
@@ -21,6 +21,7 @@ acorn install
 ```
       --acorn-dns string                       enabled|disabled|auto. If enabled, containers created by Acorn will get public FQDNs. Auto functions as disabled if a custom clusterDomain has been supplied (default auto)
       --acorn-dns-endpoint string              The URL to access the Acorn DNS service
+      --allow-traffic-from-namespace strings   Namespaces that can bypass NetworkPolicies and send network traffic to all Acorn apps
```
cjellick, line 24: "bypass" sounds off. like theyre circumventing net policies, but its just part of the policy to allow the traffic. How about:

> Namespaces that are allowed to send network traffic to all Acorn apps

docs/docs/30-installation/02-options.md
```diff
@@ -94,6 +94,13 @@ The default installation of Acorn will automatically create and sync any storage
 
 If an admin would rather manually manage the volume classes and not have these generated ones, then the `--manage-volume-classes` installation flag is available. The generated volume classes are not generated if this flag is used, and are deleted when the flag is set on an existing Acorn installation. If the flag is again switched off with `--manage-volume-classes=false`, then the volume classes will be generated again.
 
+## Kubernetes NetworkPolicies
+By default, Acorn will automatically create and manage Kubernetes [NetworkPolicies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) to isolate Acorn projects on the network level. This behavior can be disabled by passing `--network-policies=false` to `acorn install`, and can later be re-enabled by passing `--network-policies=true`.
+
+Containers and jobs that publish HTTP ports will be able to receive traffic from anywhere (inside and outside the cluster) to that port. To lock this down better (and allow traffic coming from the outside, but not other Acorn projects on the inside of the cluster), specify the `--ingress-controller-namespace` parameter during installation.
```
cjellick, line 100: I'm going to attempt to reword this. I am not sure if it is better or worse 😅 . 
```suggestion
By default, Acorn workloads that publish ports will be allowed to receive traffic from internal sources (other pods in the cluster) and external sources (through the cluster's ingress). To secure this further, you can require all traffic to Acorn workloads flow through your ingress by specifying the `--ingress-controller-namespace` parameter during installation.
```

integration/run/testdata/networkpolicy/publish.Acornfile
```diff
@@ -0,0 +1,6 @@
+containers: {
+	bar: {
+		image: "public.ecr.aws/docker/library/nginx:latest"
```
cjellick, line 3: can you udpate this to align with https://github.com/acorn-io/acorn/pull/1390

cjellick:
Can you rebase and also run `goimports` against your changes? I noticed an out-of-place import and that led me down the path of adding it to our CI

docs/docs/100-reference/01-command-line/acorn_install.md
g-linville: Good call

docs/docs/30-installation/02-options.md
g-linville: Yeah this is definitely better than what I had before. Only thing I'm adding back in is the HTTP distinction, since TCP/UDP ports will be more secure by default since we can determine the pod IP range based on node specs.

pkg/controller/appdefinition/networkpolicy.go
```diff
@@ -93,9 +94,9 @@ func NetworkPolicyForIngress(req router.Request, resp router.Response) error {
 		svc := corev1.Service{}
 		err = req.Get(&svc, ingress.Namespace, svcName)
 		if err != nil {
-			// is this okay to do? sometimes the Service doesn't exist yet when we need to get it
-			// so errors might occasionally appear in the logs but things eventually settle and work
-			return err
+			// service doesn't exist yet, so retry in 3 seconds
+			resp.RetryAfter(3 * time.Second)
```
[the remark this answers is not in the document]
cjellick: @tylerslaton whats your comment mean?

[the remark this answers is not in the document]
cjellick: @g-linville can you add an info level log statement here? can say basically the same thing as the comment

pkg/controller/appdefinition/networkpolicy.go
```diff
@@ -1,25 +1,257 @@
...
+						Protocol: &[]corev1.Protocol{corev1.ProtocolTCP}[0],
+						Port:     &targetPort,
+					})
+					netPolName = name.SafeConcatName(netPolName, strconv.Itoa(int(targetPort.IntVal)))
+				}
+			}
+		}
+
+		if len(netPolPorts) == 0 {
+			logrus.Warnf("found no matching ports between Ingress %s and Service %s in Namespace %s", ingress.Name, svcName, ingress.Namespace)
```
cjellick, line 135: Is this really a warn level statement? That feels a bit severe

cjellick (changes requested):
two minor comments around logging. fix those and i'll lgtm

pkg/controller/appdefinition/networkpolicy.go
g-linville: Yeah I wasn't really sure. Theoretically this should never happen - if it does, there's a logic error somewhere else in the Acorn controller. So I wasn't sure whether it would be appropriate for this handler to return an error in this case. But I can make it return an error if that would be better.

pkg/controller/appdefinition/networkpolicy.go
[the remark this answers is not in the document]
g-linville: Thanks Darren - I'll go change it to that. Baaah is very magical

cjellick (approved):

Issue #1227, opened 2023-02-14
Include alias for volumes similar to secret alias.


sangee2004:
acorn version v0.5.1-51-gea0d145+ea0d1458

Steps to reproduce the problem:
Deploy app using Acorn file which has following volume and secret entries:
```
volumes: {
  if !args.dev {
    "pgdata": {
      accessModes: "readWriteOnce"
    }
  }
}

secrets: {
  "quickstart-pg-pass": {
      type: "token"
  }
```

After the app is deployed , the volume created for this app does not have an alias field (like secret has) . There is no easy way to query for volumes associated with the app using its alias name that we would expect to be app-name.volume name like in the case of secrets.

```
sangeethahariharan@Sangeethas-MBP acorn-test-app % acorn secrets
ALIAS                          NAME                       TYPE      KEYS      CREATED
myapp1.quickstart-pg-pass      quickstart-pg-pass-nzqs2   token     [token]   11m ago

sangeethahariharan@Sangeethas-MBP acorn-test-app % acorn volumes
NAME                                       APP-NAME   BOUND-VOLUME   CAPACITY   STATUS    ACCESS-MODES   CREATED
pvc-dff0d439-ef52-4c9e-a49d-d9311f83d46f                             10Gi                                11m ago
sa
```

sangee2004:
Tested with acorn version - `acorn version v0.6.0-101-g613b23ca+613b23ca`

When apps are deployed with volumes , we now have "alias" fields as expected:

```
% acorn volumes                          
ALIAS                         NAME                                       APP-NAME         BOUND-VOLUME    CAPACITY   VOLUME-CLASS   STATUS     ACCESS-MODES   CREATED
testvolbindnew.my-data-bind   pvc-0ff3cb3a-812c-4b49-a2e5-e5d65f24a4b9   testvolbindnew   my-data-bind    10G                       bound      RWO            37m ago

```

PR #1359, opened 2023-03-20
add volume aliases (#1227)

jacobdonenfeld:
Acorn volumes has a new category: alias. This is `<appName>.<volumeName>`

Refactored old tests which suggested aliases already worked due to how the mockclient was set up, to more accurately match what the client is replying with.

Modified the volume translator to either accept the k8 volume name (IE: pvc-xxxx), or of the alias structure  `<appName>.<volumeName>`

addresses #1227

Also fixed an off-by-one error for parsing secret aliases

### Checklist

### DIFF
--- a/pkg/cli/rm_helper.go
+++ b/pkg/cli/rm_helper.go
@@ -37,14 +37,17 @@ func addRmObject(rmObjects *RmObjects, obj string) {
 func getSecretsToRemove(arg string, client client.Client, cmd *cobra.Command) ([]string, error) {
 	var result []string
 	secrets, err := client.SecretList(cmd.Context())
-	apps, _ := client.AppList(cmd.Context())
+	if err != nil {
+		return nil, err
+	}
 
+	apps, err := client.AppList(cmd.Context())
 	if err != nil {
 		return nil, err
 	}
 
 	for _, secret := range secrets {
-		aliasList := aliases(&secret, apps)
+		aliasList := secretAliases(&secret, apps)
 		if len(aliasList) != 0 {
 			secretName := strings.Split(aliasList[0], ".")
 			if len(secretName) != 0 && arg == secretName[0] {
@@ -62,10 +65,9 @@ func getVolumesToDelete(arg string, client client.Client, cmd *cobra.Command) ([
 	}
 
 	for _, volume := range volumes {
-		if arg == volume.Status.AppName {
+		if arg == volume.Status.AppName { // if the volume is a part of the app
 			result = append(result, volume.Name)
 		}
-
 	}
 	return result, nil
 }
--- a/pkg/cli/testdata/MockClient.go
+++ b/pkg/cli/testdata/MockClient.go
@@ -3,6 +3,7 @@ package testdata
 import (
 	"context"
 	"fmt"
+	"github.com/acorn-io/acorn/pkg/labels"
 	"net"
 
 	apiv1 "github.com/acorn-io/acorn/pkg/apis/api.acorn.io/v1"
@@ -451,10 +452,14 @@ func (m *MockClient) VolumeList(ctx context.Context) ([]apiv1.Volume, error) {
 		return m.Volumes, nil
 	}
 	return []apiv1.Volume{{
-		TypeMeta:   metav1.TypeMeta{},
-		ObjectMeta: metav1.ObjectMeta{Name: "found.volume"},
-		Spec:       apiv1.VolumeSpec{},
-		Status:     apiv1.VolumeStatus{AppName: "found", VolumeName: "found.volume"},
+		TypeMeta: metav1.TypeMeta{},
+		ObjectMeta: metav1.ObjectMeta{Name: "volume",
+			Labels: map[string]string{
+				labels.AcornVolumeName: "vol",
+				labels.AcornAppName:    "found",
+			}},
+		Spec:   apiv1.VolumeSpec{},
+		Status: apiv1.VolumeStatus{AppName: "found", VolumeName: "vol"},
 	}}, nil
 }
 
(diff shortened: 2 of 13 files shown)


pkg/cli/rm_helper.go
```diff
@@ -37,14 +37,20 @@ func addRmObject(rmObjects *RmObjects, obj string) {
...
 	var result []string
 	secrets, err := client.SecretList(cmd.Context())
-	apps, _ := client.AppList(cmd.Context())
+	if err != nil {
+		return nil, err
+	}
+	apps, err := client.AppList(cmd.Context())
+	if err != nil {
+		return nil, err
+	}
```
thedadams, line 46: This is duplicated below.

pkg/cli/volume_rm.go
thedadams, line 1: It looks like the changes in this file can be reverted.

pkg/cli/volumes.go
```diff
@@ -34,7 +40,12 @@ func (a *Volume) Run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	apps, _ := c.AppList(cmd.Context())
```
thedadams, line 43: Let's not ignore this error.

pkg/server/registry/apigroups/acorn/volumes/translator.go
```diff
@@ -22,6 +22,33 @@ type Translator struct {
 }
 
 func (t *Translator) FromPublicName(ctx context.Context, namespace, name string) (string, string, error) {
+
+	i := strings.LastIndex(name, ".")
```
thedadams, line 26: Would there be situations where there would be more than one `.`? If no, can we use `strings.Cut` as similarly used in other places? Just for consistency.

pkg/server/registry/apigroups/acorn/volumes/translator.go
```diff
@@ -22,6 +22,33 @@ type Translator struct {
...
+			labels.AcornAppName:    prefix,
+			labels.AcornVolumeName: volumeName,
+		}),
+	})
+	if err != nil {
+		return "", "", err
+	}
+
+	if len(volumes.Items) == 1 {
+		return "", volumes.Items[0].Name, nil
```
thedadams, line 49: It looks like the first string returned from this function is always empty. Can we remove this return value?

thedadams (changes requested):

pkg/server/registry/apigroups/acorn/volumes/translator.go
```diff
@@ -22,6 +22,33 @@ type Translator struct {
 }
 
 func (t *Translator) FromPublicName(ctx context.Context, namespace, name string) (string, string, error) {
+
+	i := strings.LastIndex(name, ".")
+	// If there is not a period, parse it not as an alias
+	if i == -1 || i+1 > len(name) {
```
g-linville, line 28: I don't understand the point of the `i+1 > len(name)` part of this. If the `.` is the last character in the string, then `i+1 == len(name)`. I don't think it can possibly be greater than the length of the name, unless I'm mistaken about something.

pkg/server/registry/apigroups/acorn/volumes/translator.go
g-linville: Or if the goal is to just see whether the name contains a `.` at all, we could use `strings.Contains`.

jacobdonenfeld: I copied the same approach Darren used in secrets/translator.go:

```
func (t *Translator) FromPublicName(ctx context.Context, namespace, name string) (string, string, error) {
	i := strings.LastIndex(name, ".")
	if i == -1 || i+1 > len(name) {
		return namespace, name, nil
	}

	prefix := name[:i]
	secretName := name[i+1:]
```

@ibuildthecloud is there some logic I'm missing here. For both this comment and @g-linville's comment below:

pkg/server/registry/apigroups/acorn/volumes/translator.go
jacobdonenfeld: Talked with Darren about it. Supposed to check the last character. Bug in indexing. Fixing this

jacobdonenfeld:
> @jacobdonenfeld would it be feasible to refactor the test you touched to use a `gomock` client rather than our own homegrown one? Now that we have gomock in place, id like to incrementally get ride of our custom mock client

Refactored :)

pkg/server/registry/apigroups/acorn/volumes/translator.go
jacobdonenfeld: Could potentially, but didn't want to dig into the debugging and refactoring of this. Doesn't change how it behaved before or add any new fields so I left it as is for now.

cjellick:
@jacobdonenfeld can you rebase this PR? I see an issue with import organization that I would have expected the goimports linter check to catch. So I'm curious why it didnt

pkg/cli/testdata/MockClient.go
```diff
@@ -3,6 +3,7 @@ package testdata
 import (
 	"context"
 	"fmt"
+	"github.com/acorn-io/acorn/pkg/labels"
```
cjellick, line 6: hm. goimports lint check should have caught this. can you rebase this pr to see if that will cause the linter to catch it?

pkg/cli/testdata/MockClient.go
```diff
@@ -142,6 +143,7 @@ func (m *MockClient) AppGet(ctx context.Context, name string) (*apiv1.App, error
 	case "dne":
 		return nil, fmt.Errorf("error: app %s does not exist", name)
 	case "found":
+
```
cjellick, line 146: this is the only case that has a newline between the case and the logic. can you remove it for consistency?

cjellick (changes requested):
Two small nitpicks @jacobdonenfeld

pkg/cli/volumes.go
```diff
@@ -62,3 +69,13 @@ func (a *Volume) Run(cmd *cobra.Command, args []string) error {
 
 	return out.Err()
 }
+
+// volumeAliases matches the correct app name to the given volume
+func volumeAlias(volume *apiv1.Volume) string {
+
```
cjellick, line 75: this is going to trigger a lint failure because of donnie's whitespace linter

cjellick:
```suggestion
```

pkg/cli/volumes_test.go
```diff
@@ -17,6 +22,56 @@ import (
...
+				case "dne":
+					return nil, nil
+				case "volume":
+					return &apiv1.Volume{}, nil
+				case "found.vol":
+					return &apiv1.Volume{}, nil
+				}
+				return nil, nil
+			}).AnyTimes()
+
```
cjellick, line 72: this will trigger the lint issue as well
```suggestion
```

pkg/server/registry/apigroups/acorn/volumes/translator.go
```diff
@@ -22,6 +22,33 @@ type Translator struct {
 }
 
 func (t *Translator) FromPublicName(ctx context.Context, namespace, name string) (string, string, error) {
+
```
cjellick, line 25: future lint issue
```suggestion
```

cjellick:
I pre-emptively addressed a couple whitespace issues that would have later failed because of the new linter we introduced.

cjellick (approved):

Issue #1370, opened 2023-03-22
local.on-acorn.io breaks cert-manager integration


cjellick:
In our own infra, we have the following setup:

First, acorn-dns and let's encrypt integration are both disabled. We provision certs using cert-manager instead. It watches ingress with proper annotations and creates certs accordingly.

Next, we create an app using the --publish syntax to specify the FQDN we want to use, like:
```
acorn run --publish myapp.jellick.world:webapp ghcr.io/acorn-io/library/hello-world
```

The problem is that acorn will always also inject a local.on-acorn.io endpoint for that app. So, the ingress that is created has the user-specified domain AND the local.on-acorn.io domain. Cert-manager then tries to generate certs for both, but the local one ALWAYS fails and it blocks the user-specified one from completing. So, our apps are unable to get certs.

sangee2004:
Tested on internal Infra with the fix:

Launch app with the following publish option to external names :

```
acorn run --publish myapp.test.world:webapp ghcr.io/acorn-io/library/hello-world
```

Once the app is deployed, we are able to get endpoint for  external name.


In the  Infra setup with no fix , launching the app in the similar way will not result in any endpoint for external name being made available.

PR #1371, opened 2023-03-22
Dont use local.on-acorn.io when acorn-dns disabled (#1370)

cjellick:
This is a change of behavior to how we handle the acorn-dns install flag being set to "disabled". Previously, we would still inject a local.on-acorn.io endpoint if the user hadn't supplied a custom clusterDomain. With this change, if acorn-dns is disabled, we won't inject the local domain. This could potentially result in an app having no endpoints, but that's now the expected behavior.

This fixes an issue we hit in our Hub clusters where we wanted to control our app endpoints using the --publish syntax and then use cert-manager to provision certs for the generated ingress, but because the generated ingress would always have a local.on-acorn.io endpoint (that would never correctly resolve in the ACME HTTP01 challenge), the cert for the endpoint specified via --publish would never get a cert.

Addresses #1370

### Checklist


generate.go
```diff
@@ -4,5 +4,6 @@
 //go:generate go run github.com/acorn-io/baaah/cmd/deepcopy ./pkg/apis/admin.acorn.io/v1/
 //go:generate go run k8s.io/kube-openapi/cmd/openapi-gen -i github.com/acorn-io/acorn/pkg/apis/internal.acorn.io/v1,github.com/acorn-io/acorn/pkg/apis/api.acorn.io/v1,github.com/acorn-io/acorn/pkg/apis/internal.admin.acorn.io/v1,github.com/acorn-io/acorn/pkg/apis/admin.acorn.io/v1,k8s.io/apimachinery/pkg/apis/meta/v1,k8s.io/apimachinery/pkg/runtime,k8s.io/apimachinery/pkg/version,k8s.io/apimachinery/pkg/api/resource,k8s.io/api/core/v1,k8s.io/api/rbac/v1 -p ./pkg/openapi/generated -h tools/header.txt
 //go:generate go run github.com/golang/mock/mockgen --build_flags=--mod=mod -destination=./pkg/mocks/mock_client.go -package=mocks github.com/acorn-io/acorn/pkg/client Client,ProjectClientFactory
+//go:generate go run github.com/golang/mock/mockgen --build_flags=--mod=mod -destination=./pkg/mocks/k8s/mock.go -package=mocks sigs.k8s.io/controller-runtime/pkg/client Reader
```
jacobdonenfeld, line 7: Another option I used when mocking the controller was to use the k8's fake package for controller-runtime: sigs.k8s.io/controller-runtime/pkg/client/fake. This was a more finicky to figure out how to use our CRD's, but I assumed the fake library would be the most hardened way to do that.

Good example at acorn/pkg/client/info_test.go

I also like this approach though and don't see anything wrong with it

Used this in acorn/pkg/client/info_test.go

cjellick: yeah, I started there as well. Went with this mock for a few reasons:
1. see how hard/easy it would be to do
2. I didn't really want to depend on the fake client's behavior if a standard mock would do. Seemed like having a single approach to managing mocks made sense.

thedadams (approved):

generate.go
jacobdonenfeld: Yeah, this is way more straightforward than using fake.

jacobdonenfeld (approved):

Issue #1388, opened 2023-03-25
Not able to deploy apps in new project because of error relating to no default regions being present. 


sangee2004:
Acorn version v0.6.0-77-g5c64a93e+5c64a93e

Steps to reproduce the problem:
Create a new project.
Deploy app in that project.
App deployment fails with following errors:
```
  ✗  ERROR:  App.api.acorn.io "mytest2" is invalid: spec.region: Invalid value: "": no region can be determined because project default region is not set
```
```
 %acorn project create testnew 
testnew
% acorn project use testnew
% acorn run -n mytest2 .      
[+] Building 3.5s (5/5) FINISHED                                                                                                                                                                     
 => [internal] load build definition from Dockerfile                                                                                                                                            0.0s
 => => transferring dockerfile: 48B                                                                                                                                                             0.0s
 => [internal] load .dockerignore                                                                                                                                                               0.0s
 => => transferring context: 35B                                                                                                                                                                0.0s
 => [internal] load metadata for docker.io/library/nginx:latest                                                                                                                                 1.2s
 => CACHED [1/1] FROM docker.io/library/nginx@sha256:f4e3b6489888647ce1834b601c6c06b9f8c03dee6e097e13ed3e28c01ea3ac8c                                                                           0.0s
 => => resolve docker.io/library/nginx@sha256:f4e3b6489888647ce1834b601c6c06b9f8c03dee6e097e13ed3e28c01ea3ac8c                                                                                  0.0s
 => exporting to image                                                                                                                                                                          2.3s
 => => exporting layers                                                                                                                                                                         0.0s
 => => exporting manifest sha256:0ff71b46d22886f535ff153809798ba0fea5343f0f93c1d174f4e2222445c3be                                                                                               0.0s
 => => exporting config sha256:71e271c6570043c75162588a9534cf04d9c3803a8e24175422b88509000ac2c3                                                                                                 0.0s
 => => pushing layers                                                                                                                                                                           2.3s
 => => pushing manifest for 127.0.0.1:5000/acorn/testnew:latest@sha256:0ff71b46d22886f535ff153809798ba0fea5343f0f93c1d174f4e2222445c3be                                                         0.0s
[+] Building 0.1s (5/5) FINISHED                                                                                                                                                                     
 => [internal] load build definition from Dockerfile                                                                                                                                            0.0s
 => => transferring dockerfile: 58B                                                                                                                                                             0.0s
 => [internal] load .dockerignore                                                                                                                                                               0.0s
[...]

PR #1389, opened 2023-03-25
Ensure calculated default set on project (#1388)

thedadams:
### Checklist

Issue: https://github.com/acorn-io/acorn/issues/1388


integration/helper/conflict.go
```diff
@@ -0,0 +1,30 @@
...
+		updater()
+
+		if err := c.Update(ctx, obj); !apierrors.IsConflict(err) {
+			return err
+		}
+
+		// Give things a chance to settle down.
+		time.Sleep(time.Second)
+	}
+}
```
g-linville, line 30: It looks like the only way this loop terminates is if an error occurs. I might be misunderstanding something, but should there be a check to break from the loop if the update succeeded?

thedadams: IsConflict returns false if `err == nil`.

g-linville: Oh that makes sense. For some reason my brain was looking at that return and thought that it was another error case and not a success case.

g-linville (approved):

thedadams:
I updated this PR to do two things:
1. Handle existing projects on upgrade so that the region logic works
2. Refactor some tests that were integration tests that should have been unit tests.

pkg/server/registry/apigroups/acorn/projects/validator_test.go
```diff
@@ -0,0 +1,148 @@
...
+					SupportedRegions: []string{"my-region"},
+				},
+			},
+		},
+		{
+			name: "Update project to have default region and non-existent supported regions",
+			newProject: apiv1.Project{
+				Spec: apiv1.ProjectSpec{
+					DefaultRegion:    "my-region",
+					SupportedRegions: []string{"my-region", "dne-region"},
```
tylerslaton, line 115: This shouldn't fail?

tylerslaton (approved):
Looks good to me, non-blocking comments we can address if we feel they're necessary.

pkg/server/registry/apigroups/acorn/apps/validator.go
```diff
@@ -135,11 +135,14 @@ func (s *Validator) validateRegion(ctx context.Context, app *apiv1.App) error {
 	}
 
 	if app.Spec.Region == "" {
-		if project.Spec.DefaultRegion == "" && project.Status.DefaultRegion == "" {
-			return fmt.Errorf("no region can be determined because project default region is not set")
+		app.Status.Defaults.Region = project.Spec.DefaultRegion
```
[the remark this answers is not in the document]
thedadams: We have to have something to indicate the default region of an app so that it gets synced properly.

Answer to your side question: yes. If the project default changes, then the app default will change. However, this should be handled correctly. That switch should be rare, however. I don't want apps to be switching regions constantly and possible cause issues.

pkg/server/registry/apigroups/acorn/projects/validator_test.go
tylerslaton: Makes sense, thanks.

pkg/server/registry/apigroups/acorn/projects/validator_test.go
```diff
@@ -0,0 +1,148 @@
...
+			name:      "Update project to have default region, no supported regions",
+			wantError: true,
+			newProject: apiv1.Project{
+				Spec: apiv1.ProjectSpec{
+					DefaultRegion: "my-region",
+				},
+			},
+		},
+		{
+			name: "Update project to have default region and supported region",
```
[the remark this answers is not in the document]
thedadams: I am not sure what you are asking here. My interpretation of your question is handled by this test case.

g-linville (approved):

pkg/server/registry/apigroups/acorn/apps/validator.go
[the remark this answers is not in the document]
thedadams: Technically speaking, an app's region is immutable. If a user supplies a region when they run an app, then they cannot change it.

The default is not immutable. I did this to be consistent with other scenarios. An admin changes a default on a project and eventually all the things that use that default are changed, too.

I'd be happy to change this to be immutable as well if we can come up with reasonable behavior.

PR #1408, opened 2023-03-29
NetworkPolicies: Handle case where an Ingress routes to an ExternalName Service (#456)

g-linville:
re: #456

Using a router and links, it is possible to expose a port in a service in a different namespace than the namespace in which the corresponding Ingress lives. Here is an example:

```cue
containers: "two": {
  image: "nginx:latest"
  ports: "80/http"
  files: {
    "etc/nginx/conf.d/default.conf": """
    server {
      listen 80;
      server_name localhost;
      location /two {
        return 200;
      }
    }
    """
  }
}
```

Run the above Acornfile: `acorn run --name two .`

```cue
services: {
  one: {
    container: "nginx-one"
    ports: ["80/http"]
    default: true
  }
}

routers: default: routes: {
  "/one": "one:80"
  "/two": "two:80"
}

containers: {
  "nginx-one": {
    image: "nginx:latest"
    ports: "80/http"
    files: {
      "/etc/nginx/conf.d/default.conf": """
      server {
        listen 80;
        server_name localhost;
        location /one {
          return 200;
        }
      }
      """
    }
  }
}
```

Run the above Acornfile and link it to the container from the previous one: `acorn run --name one --link two:two .`

This will create an Ingress in `one`'s namespace that routes to an ExternalName Service, which then resolves to a ClusterIP service in `two`'s namespace. I wasn't accounting for this scenario before, so these changes deal with that.

Unfortunately I also had to add a finalizer in here. Since we sometimes create NetworkPolicies in namespaces outside where the owning Ingress lives, if the app that created the Ingress is removed (in this example, `acorn rm one`), the namespace for the app might delete too quickly, and the NetworkPolicy in the other namespace (`two`'s namespace in this example) will get left behind. The finalizer ensures that the NetworkPolicy gets cleaned up properly.

### Checklist

### DIFF
--- a/pkg/controller/appdefinition/networkpolicy.go
+++ b/pkg/controller/appdefinition/networkpolicy.go
@@ -76,12 +76,6 @@ func NetworkPolicyForApp(req router.Request, resp router.Response) error {
 // Acorn apps from the ingress controller. If the ingress controller namespace is not defined, traffic from
 // all namespaces will be allowed instead.
 func NetworkPolicyForIngress(req router.Request, resp router.Response) error {
-	// check if this is being called as a finalizer for a deleted Ingress
-	// we need this because we sometimes create NetworkPolicies in different namespaces from where their owning Ingresses live
-	if !req.Object.GetDeletionTimestamp().IsZero() {
-		return nil
-	}
-
 	cfg, err := config.Get(req.Ctx, req.Client)
 	if err != nil {
 		return err
--- a/pkg/controller/appdefinition/networkpolicy.go
+++ b/pkg/controller/appdefinition/networkpolicy.go
@@ -3,6 +3,7 @@ package appdefinition
 import (
 	"fmt"
 	"strconv"
+	"strings"
 
 	v1 "github.com/acorn-io/acorn/pkg/apis/internal.acorn.io/v1"
 	"github.com/acorn-io/acorn/pkg/config"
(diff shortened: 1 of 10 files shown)


pkg/controller/appdefinition/networkpolicy.go
```diff
@@ -72,6 +76,12 @@ func NetworkPolicyForApp(req router.Request, resp router.Response) error {
 // Acorn apps from the ingress controller. If the ingress controller namespace is not defined, traffic from
 // all namespaces will be allowed instead.
 func NetworkPolicyForIngress(req router.Request, resp router.Response) error {
+	// check if this is being called as a finalizer for a deleted Ingress
+	// we need this because we sometimes create NetworkPolicies in different namespaces from where their owning Ingresses live
+	if !req.Object.GetDeletionTimestamp().IsZero() {
```
cjellick, line 81: Do you need this check now that its not a finalizer?

cjellick (dismissed):
just not sure if you need the deleteTimestamp check anymore

tylerslaton (approved):

pkg/controller/appdefinition/networkpolicy.go
g-linville: Good call!

thedadams (approved):

PR #1415, opened 2023-03-31
Remove project placements

thedadams:
In order to remove project placements, the role of regions is expanded. More objects are designated as regioned and more methods to get and set the region are added.

### DIFF
--- a/pkg/apis/api.acorn.io/v1/types.go
+++ b/pkg/apis/api.acorn.io/v1/types.go
@@ -89,6 +89,29 @@ type ContainerReplicaStatus struct {
 	Image                string                  `json:"image"`
 	ImageID              string                  `json:"imageID"`
 	Started              *bool                   `json:"started,omitempty"`
+
+	Region string `json:"region,omitempty"`
+}
+
+// ForRegion checks or sets the region of a ContainerReplica.
+// If a ContainerReplica's region is unset, ForRegion sets it to the given region and returns true.
+// Otherwise, it returns true if and only if the ContainerReplica belongs to the given region.
+func (in *ContainerReplica) ForRegion(region string) bool {
+	// If the region of a Container Replica is not set, then it hasn't been synced yet. In this case, we assume that it is in
+	// the same region as the app, and return true.
+	if in.Status.Region == "" {
+		in.Status.Region = region
+	}
+
+	return in.Status.Region == region
+}
+
+func (in *ContainerReplica) ForOtherRegions(region string) bool {
+	return in.Status.Region != region
+}
+
+func (in *ContainerReplica) GetRegion() string {
+	return in.Status.Region
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
@@ -394,7 +417,6 @@ type ProjectSpec struct {
 }
 
 type ProjectStatus struct {
-	Placement     string `json:"placement,omitempty"`
 	Namespace     string `json:"namespace,omitempty"`
 	DefaultRegion string `json:"defaultRegion,omitempty"`
 }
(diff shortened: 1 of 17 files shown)


pkg/controller/defaults/region.go
```diff
@@ -3,30 +3,36 @@ package defaults
...
+		if region == "" {
+			if r := ns.Annotations[labels.AcornCalculatedProjectDefaultRegion]; r != "" {
+				region = r
+			} else {
+				region = "local"
+			}
+		}
 	}
 
+	obj.SetDefaultRegion(region)
```
StrongMonkey, line 36: should we move this line above inside the if block so that we only set defaultRegion when obj.GetRegion() is empty?

StrongMonkey (approved):
Looks good! Just several non-blocking comments

pkg/server/registry/apigroups/acorn/regions/strategy.go
```diff
@@ -28,6 +28,11 @@ func (s *strategy) Get(_ context.Context, _, name string) (types.Object, error)
 		ObjectMeta: metav1.ObjectMeta{
 			Name:              "local",
 			CreationTimestamp: s.startTime,
+			OwnerReferences: []metav1.OwnerReference{
```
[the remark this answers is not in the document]
thedadams: This is so that the CLI output correctly denotes that this region is for the "local" cluster.

pkg/controller/defaults/region.go
thedadams: I don't think so. There are situations where we want to change the default region even if it is set. For example: builders. We always want the builder to run in the default region. So, if the default region for a project changes, then this should be updated.

StrongMonkey: Got it. But in this case, should we always get default region from its project. The current logic will set default region to empty string if obj.GetRegion() != ""

thedadams: Yep, you're right. Updated.

pkg/apis/api.acorn.io/v1/types.go
```diff
@@ -89,6 +89,28 @@ type ContainerReplicaStatus struct {
 	Image                string                  `json:"image"`
 	ImageID              string                  `json:"imageID"`
 	Started              *bool                   `json:"started,omitempty"`
+
+	Region string `json:"region,omitempty"`
+}
+
+// ForRegion for a ContainerReplica is called when the ContainerReplica is synced.
+// This ensures the proxy can find the correct ContainerReplica for exec/logs.
+func (in *ContainerReplica) ForRegion(region string) bool {
```
jacobdonenfeld, line 98: NIT: The naming "ForX" isn't intuitive to me. Would make sense to be "IsRegion" and "IsNotRegion". Unsure if there is some naming convention or use for using "For", so might be missing something.

jacobdonenfeld (approved):
One NIT, lgtm, I like the implementation

pkg/apis/api.acorn.io/v1/types.go
thedadams: The reason I used `ForRegion` is that certain objects (like cluster compute classes) actually modify the object to be for the region, other just do a check that would be more like and "is region" approach. I thought `ForRegion` was a good middle ground.

njhale, line 98:
```suggestion
// ForRegion checks or sets the region of a ContainerReplica.
// If a ContainerReplica's region is unset, ForRegion sets it to the given region and returns true; otherwise it returns true IFF a ContainerReplica belongs to the given region.
func (in *ContainerReplica) ForRegion(region string) bool {
```

njhale: Could we reframe the method as `SetRegion(region string) bool` to make it clear at the call site that a mutation _could_ take place?

njhale (approved):
Just a few nits. Otherwise LGTM.

pkg/apis/api.acorn.io/v1/types.go
thedadams: Done.

tylerslaton (approved):

PR #1432, opened 2023-04-04
Disable NetworkPolicies for the nightly EKS test

g-linville:
Unless/until we have EKS using a CNI that enforces NetworkPolicies, we should disable them on our nightly test so that the test that relies on them will skip instead of just failing.

### DIFF
--- a/.github/workflows/aws-eks-test.yml
+++ b/.github/workflows/aws-eks-test.yml
@@ -47,6 +47,7 @@ jobs:
           if acorn check; then
             acorn uninstall -af
           fi
+          # Install Acorn with NetworkPolicies disabled because EKS does not enforce them, so tests will fail if we leave them enabled
           acorn install --image ghcr.io/acorn-io/acorn:main --network-policies=false
           
         env:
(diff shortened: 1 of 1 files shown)


tylerslaton (approved):

.github/workflows/aws-eks-test.yml
```diff
@@ -47,7 +47,7 @@ jobs:
           if acorn check; then
             acorn uninstall -af
           fi
-          acorn install --image ghcr.io/acorn-io/acorn:main
+          acorn install --image ghcr.io/acorn-io/acorn:main --network-policies=false
```
thedadams, line 50: Let's add a comment explaining this for the time being.

thedadams: Thank you!

jacobdonenfeld:
Curious what the error was when installing it without set to false. Possibility of a useful validation step during install?

thedadams:
@jacobdonenfeld The docs are there for the next version.

g-linville:
> Curious what the error was when installing it without set to false. Possibility of a useful validation step during install?

There wasn't an error, but there is an integration test that runs and assumes that NetPols are working if they are configured. So if they are configured to be created, but k8s isn't enforcing them, then the test fails. But if they're disabled, the test just skips.

jacobdonenfeld:
Sounds good to me!

Issue #1398, opened 2023-03-28
Autoupgrade images hosted as private images in docker hub results in a pull every upgrade interval eating into rate limits even when no actual pull of images is done.


sangee2004:
Acorn version -v0.6.0-81-g5856fe34+5856fe34

Steps to reproduce the problem:

login to dockerhub as `<user1>`(`acorn login`)
Deploy app in autoupgrade mode for a private image hosted in docker hub
```acorn run -n mytestautoprivate --interval 10s docker.io/`<user1>`/mytestnewacorn:v#.#.# ```

Notice that the `ratelimit-remaining` reported for the user steadily declines with time even no new versions of the image is made available .

```
% date
Tue Mar 28 09:11:56 PDT 2023

%./ratelimit
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  4511    0  4511    0     0  14568      0 --:--:-- --:--:-- --:--:-- 14693
HTTP/1.1 200 OK
content-length: 2782
content-type: application/vnd.docker.distribution.manifest.v1+prettyjws
docker-content-digest: sha256:767a3815c34823b355bed31760d5fa3daca0aec2ce15b217c9cd83229e0e2020
docker-distribution-api-version: registry/2.0
etag: "sha256:767a3815c34823b355bed31760d5fa3daca0aec2ce15b217c9cd83229e0e2020"
date: Tue, 28 Mar 2023 16:12:01 GMT
strict-transport-security: max-age=31536000
ratelimit-limit: 200;w=21600
**ratelimit-remaining: 154;w=21600**
docker-ratelimit-source: 44a71d06-5a37-4e1f-9073-ce6dad1c3aa3

% date       
Tue Mar 28 09:15:48 PDT 2023
 
% ./ratelimit
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  4511    0  4511    0     0  15002      0 --:--:-- --:--:-- --:--:-- 15188
HTTP/1.1 200 OK
content-length: 2782
content-type: application/vnd.docker.distribution.manifest.v1+prettyjws
docker-content-digest: sha256:767a3815c34823b355bed31760d5fa3daca0aec2ce15b217c9cd83229e0e2020
docker-distribution-api-version: registry/2.0
[...]

PR #1433, opened 2023-04-04
Changes to prevent pull every upgrade interval for private images (#1398)

yashgiri:
### Checklist

### DIFF
--- a/pkg/autoupgrade/daemon_test.go
+++ b/pkg/autoupgrade/daemon_test.go
@@ -185,10 +185,11 @@ func TestRefreshImages(t *testing.T) {
 	thirtySecondsAgo := now.Add(-30 * time.Second)
 	ptrTrue := &[]bool{true}[0]
 	appImages := map[string]string{
-		"test-1":      "acorn/test-1:v#.#.#",
-		"acorn-1":     "docker.io/acorn/acorn-1:v1.1.1-*",
-		"enabled-app": "docker.io/acorn/enabled:latest",
-		"notify-app":  "docker.io/acorn/notify:latest",
+		"test-1":           "acorn/test-1:v#.#.#",
+		"acorn-1":          "docker.io/acorn/acorn-1:v1.1.1-*",
+		"enabled-app":      "docker.io/acorn/enabled:latest",
+		"notify-app":       "docker.io/acorn/notify:latest",
+		"test-autoupgrade": "index.docker.io/acorn/test-autoupgrade:v#.#.#",
 	}
 	apps := make(map[kclient.ObjectKey]v1.AppInstance, len(appImages))
 	for _, entry := range typed.Sorted(appImages) {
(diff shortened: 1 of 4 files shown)


pkg/imagedetails/imagedetails.go
```diff
@@ -10,18 +10,27 @@ import (
...
 
+const defaultNoReg = "xxx-no-reg"
+
 func GetImageDetails(ctx context.Context, c kclient.Client, namespace, imageName string, profiles []string, deployArgs map[string]any, opts ...remote.Option) (*apiv1.ImageDetails, error) {
 	imageName = strings.ReplaceAll(imageName, "+", "/")
 	name := strings.ReplaceAll(imageName, "/", "+")
 
 	if tagPattern, isPattern := autoupgrade.AutoUpgradePattern(imageName); isPattern {
-		if latestImage, found, err := autoupgrade.FindLatestTagForImageWithPattern(ctx, c, namespace, imageName, tagPattern); err != nil {
+		current, err := imagename.ParseReference(imageName, imagename.WithDefaultRegistry(defaultNoReg))
```
thedadams, line 28: I'm not sure if this will work. For example, `v#.#.#` is a supported autoupgrade pattern, but `imagename.ParseReference` will error trying to parse an image with that tag.

thedadams (changes requested):

pkg/autoupgrade/tags.go
```diff
@@ -61,12 +61,12 @@ func findLatestTagForImageWithPattern(ctx context.Context, c daemonClient, names
 		}
 	}
 
-	return newImage, newTag != imagename.DefaultTag, err
+	return newImage, newTag != current, err
```
cjellick, line 64: Im starting to think that there is a more fundamental flaw here. This function is used in two places: the ImageDetails validator and the autoUpgrade daemon. The two callers have different expectations of what the boolean value returned is supposed to mean. The imageDetails caller thinks it means "a matching tag was found". The daemon thinks it means "I found an image you should update to because it is better than your current". They're similar but subtly different.

So, now I don't think passing in the "current" tag is the right thing to do (because that doesnt jive with the usage in imageDetails).

I'm still thinking this through

cc @thedadams

cjellick: Ok, put together a thought on how to tackle this.

First, the boolean return by this function should be defined to mean whether a positive matching tag has been found. We should not pass "current" into this function and attempt to compare here.

Secondly (and somewhat unrelated) `newTag != imagename.DefaultTag` seems like it could introduce a subtle bug to me because `imagename.DefaultTag` ("latest") is a valid tag that could sometimes be the right choice as the latest tag. So, higher up in this function, instead of using imagename.DefaultTag as the placeholder for "a tag cant be found", use a different value that would actually be an invalid tag like `"+notfound+"` (`+` is an illegal character).

Thirdly, in the daemon's use case, you do need to compare the image returned by this function to the app's current image, but just do it there, after you return from this function. Don't try to do it in this function.

@thedadams feel free to chime in on this.

thedadams: Yes, I agree with everything you said here. I will add two thoughts.

Based on what you said, the best way to **guarantee** that we haven't found a matching image is to use something like `+notfound+` as the "starting tag" because we know that will never be a valid matching tag.

Secondly, the use of `+notfound+` is an implementation detail of this function and shouldn't leak out. Meaning that, even if `newTag == "+notfound+"`, the `newImage` returned here shouldn't contain that as its tag. Since `+notfound+` is not a valid tag, returning an image with that tag may produce bugs/errors in other areas, too.

pkg/imagedetails/imagedetails.go
yashgiri: I might have to go back and check on this because when i was running the debugger i think i did pass successfully through the debugger through this line but there is a possibility that i might be thinking of something else. Anyhow since we are removing the usage of current this line will not be present. 👍

pkg/autoupgrade/tags.go
yashgiri: I agree with the first 2 points of Craig as well and i will go ahead with that implementation change as recommended by @thedadams.  👍 

Out of curiosity I was wondering if it would have been better to have a function present in the `tags.go` which would either use `findLatestTagForImageWithPattern` or use similar functionality with added use of current to compare with the latest image. Essentially to handle what has been said before - The imageDetails caller thinks it means "a matching tag was found". The daemon thinks it means "I found an image you should update to because it is better than your current".

thedadams: I had a similar thought. I even wrote up an entire expose on doing something like that.

The reason I didn't post it is the name `findLatestTagForImageWithPattern` implies (to me) that it will return the latest tag. It would be up to the caller to determine what to do with the latest tag.

cjellick: This is looking good. I'm going to resolve this convesation.

cjellick:
Can you add a unit test for `findLatestTagForImageWithPattern`? You should be able to mock the daemonClient interface, which should make this relatively straight forward. @jacobdonenfeld has experience with mocking and our unit tests, so he might be able to provide some guidance.

cjellick:
Regarding mocks - @thedadams already introduced a mockclient in daemon_test.go. You can just use that with some basic modifications

cjellick:
Here is the subtle bug that I think you have:
Say your app is currently on image: `docker.io/some/image:v1.0` and that image has two tags: `v1.0` and `v0.9`. Also, for this experiment, your tag patter for that app is `v#.#`.

Now, say that `c.checkImageAllowed(ctx, namespace, img)` has started returning an error for the `v1.0` tag. That function returns an error when it determines the tagged image doesnt have the required signatures as defined by @iwilltry42's [imageAllowRules feature](https://github.com/acorn-io/acorn/pull/1240).

In this scenario, we DONT want the autoupgrade feature to downgrade to v0.9. We just want to app to stay at v1.0, even though that image no longer has the proper signatures. So, `findLatestTagForImageWithPattern` should return the following values: `docker.io/some/image:v1.0`, `false`, and `nil`.

You should be able to write a test for this using @thedadams's mocked client

thedadams:
@cjellick I was in the middle of writing this scenario up in the Slack channel when I saw your message here. On the same page.

yashgiri:
I guess i understand the issue, this happens because i passed imagename.DefaultTag instead of current. Because of this the tags returned are 0.9 and 1.0, if i had passed `current` only the tags >= current (only 1.0) would have been returned. Is that the correct reasoning here?
My question would be doesn't the latest tag only make the latest image return?

cjellick:
> I guess i understand the issue, this happens because i passed imagename.DefaultTag instead of current. Because of this the tags returned are 0.9 and 1.0, if i had passed current only the tags >= current (only 1.0) would have been returned. Is that the correct reasoning here?
> My question would be doesn't the latest tag only make the latest image return?

That seems to be the case. Definitely write some tests to prove that out.

yashgiri:
Hi @cjellick , @thedadams  I have added a testcase for TestRefreshImages function for the case we discussed above. That seemed the correct place for me to create a testcase. Please let me know if there should be more test cases added elsewhere if need be.

pkg/autoupgrade/daemon_test.go
```diff
@@ -25,6 +26,7 @@ type mockDaemonClient struct {
 	localTags, remoteTags               []string
 	remoteImageDigest, resolvedLocalTag string
 	localTagFound                       bool
+	imageDenyList                       []string
```
cjellick, line 29: how about a map instead since you just need to check for existence

pkg/autoupgrade/daemon_test.go
```diff
@@ -337,6 +341,22 @@ func TestRefreshImages(t *testing.T) {
...
 			appsUpdated:            map[string]string{"notify-app": "docker.io/acorn/notify"},
 		},
+		{
+			name:                   "Auto refresh tag with multiple remote tags with latest tag image denied and current being an older image",
+			client:                 &mockDaemonClient{remoteTags: []string{"v1.1.1", "v1.1.2"}, imageDenyList: map[string]any{"v1.1.2": true}},
+			appKeysPrevCheckBefore: map[kclient.ObjectKey]time.Time{router.Key("acorn", "test-1"): thirtySecondsAgo},
+			imagesToRefresh:        map[imageAndNamespaceKey][]kclient.ObjectKey{{image: "index.docker.io/acorn/test-1:v1.1.1", namespace: "acorn"}: {router.Key("acorn", "test-1")}},
+			appKeysPrevCheckAfter:  map[kclient.ObjectKey]time.Time{router.Key("acorn", "test-1"): now},
+			appsUpdated:            map[string]string{"test-1": "index.docker.io/acorn/test-1:v1.1.2"},
+		},
```
thedadams, line 351: I would expect that if the current tag is `v1.1.1` and the tag `v1.1.2` is denied, then the app would not have its image updated. The issue, I believe, is that the `imageDenyList` keys should be the full image: `index.docker.io/acorn/test-1:v1.1.2`.

Note that doing this in the test below reveals a bug in your implementation.

pkg/autoupgrade/daemon_test.go
```diff
@@ -25,6 +25,7 @@ type mockDaemonClient struct {
 	localTags, remoteTags               []string
 	remoteImageDigest, resolvedLocalTag string
 	localTagFound                       bool
+	imageDenyList                       map[string]any
```
thedadams, line 28: The way I usually do this in Go is 
```suggestion
	imageDenyList                       map[string]struct{}
```
Then, when you check for existence below...

pkg/autoupgrade/daemon_test.go
```diff
@@ -63,8 +64,11 @@ func (m *mockDaemonClient) resolveLocalTag(context.Context, string, string) (str
 	return m.resolvedLocalTag, m.localTagFound, nil
 }
 
-func (m *mockDaemonClient) checkImageAllowed(context.Context, string, string) error {
-	return nil // TODO: use some switch to mock an error e.g. based on mockDaemonClient.imageDenyList
+func (m *mockDaemonClient) checkImageAllowed(_ context.Context, _ string, img string) error {
+	if m.imageDenyList[img] == true {
```
thedadams, line 68:
```suggestion
	if _, ok := m.imageDenyList[img]; ok {
```

thedadams (changes requested):

pkg/autoupgrade/daemon_test.go
yashgiri: Got it, yeah i see that now. Just wanted to clarify that by reveals a bug in your implementation you mean the testing code logic correct?

yashgiri: I realized that not this testcase but in the next one once i fix the image name in the imageDenyList, the `findLatestTagForImageWithPattern` keeps looping currently fixing that logic 👍

pkg/autoupgrade/daemon_test.go
```diff
@@ -337,6 +341,30 @@ func TestRefreshImages(t *testing.T) {
...
+			appKeysPrevCheckAfter:  map[kclient.ObjectKey]time.Time{router.Key("acorn", "test-1"): now},
+			appsUpdated:            map[string]string{"test-1": "index.docker.io/acorn/test-1:v1.1.1"},
+		},
+		{
+			name:                   "Auto refresh tag with multiple remote tags with latest tag image denied and current being latest image",
+			client:                 &mockDaemonClient{remoteTags: []string{"v1.1.1", "v1.1.2"}, imageDenyList: map[string]struct{}{"index.docker.io/acorn/test-1:v1.1.2": {}}},
+			appKeysPrevCheckBefore: map[kclient.ObjectKey]time.Time{router.Key("acorn", "test-1"): thirtySecondsAgo},
+			imagesToRefresh:        map[imageAndNamespaceKey][]kclient.ObjectKey{{image: "index.docker.io/acorn/test-1:v1.1.2", namespace: "acorn"}: {router.Key("acorn", "test-1")}},
+			appKeysPrevCheckAfter:  map[kclient.ObjectKey]time.Time{router.Key("acorn", "test-1"): now},
+			appsUpdated:            map[string]string{"test-1": "index.docker.io/acorn/test-1:v1.1.2"},
```
cjellick, line 358: This is saying that you expect the app`test-1` to be updated to image `index.docker.io/acorn/test-1:v1.1.2` in this scenario.

But since `v1.1.2` is on the imageDenyList, that is not the desired outcome in this scenario. The desired outcome is that  the app is not updated (and so, the `appsUpdated` map should be empty).

pkg/autoupgrade/daemon_test.go
```diff
@@ -337,6 +341,30 @@ func TestRefreshImages(t *testing.T) {
 			appKeysPrevCheckAfter:  map[kclient.ObjectKey]time.Time{router.Key("acorn", "notify-app"): now},
 			appsUpdated:            map[string]string{"notify-app": "docker.io/acorn/notify"},
 		},
+		{
+			name:                   "Auto refresh tag with multiple remote tags with latest tag image denied and current being an older image",
+			client:                 &mockDaemonClient{remoteTags: []string{"v1.1.1", "v1.1.2"}, imageDenyList: map[string]struct{}{"index.docker.io/acorn/test-1:v1.1.2": {}}},
+			appKeysPrevCheckBefore: map[kclient.ObjectKey]time.Time{router.Key("acorn", "test-1"): thirtySecondsAgo},
+			imagesToRefresh:        map[imageAndNamespaceKey][]kclient.ObjectKey{{image: "index.docker.io/acorn/test-1:v1.1.1", namespace: "acorn"}: {router.Key("acorn", "test-1")}},
+			appKeysPrevCheckAfter:  map[kclient.ObjectKey]time.Time{router.Key("acorn", "test-1"): now},
+			appsUpdated:            map[string]string{"test-1": "index.docker.io/acorn/test-1:v1.1.1"},
```
cjellick, line 350: This is saying that you expect the app `test-1` to be updated to `v1.1.1`, but it is already at `v1.1.1`. So, we want **no** update to the app, right? This `appsUpdated` map should be empty in this test case.

pkg/autoupgrade/daemon_test.go
yashgiri: that makes sense, I was under the assumption that appsUpdated will need some value to check against with. But essentially there should be no change.

pkg/autoupgrade/tags.go
```diff
@@ -35,17 +37,22 @@ func getTagsForImagePattern(ctx context.Context, c daemonClient, namespace, imag
...
 		return "", false, err
 	}
 
-	newTag := imagename.DefaultTag
+	var newTag string
+	if current == "" {
+		newTag = invalidTag
+	} else {
+		newTag = current
+	}
```
thedadams, line 51: nit: Let's simplify this:
```suggestion
	newTag := current
	if newTag == "" {
		newTag = invalidTag
	}
```

cjellick (approved):

thedadams (approved):

PR #1436, opened 2023-04-04
Fix faulty merging and update tests for Config

tylerslaton:
### Checklist
- [ ]  PR has at least two approvals before merging (or a reasonable exception, like it's just a docs change)

### DIFF
--- a/docs/docs/30-installation/02-options.md
+++ b/docs/docs/30-installation/02-options.md
@@ -102,19 +102,25 @@ By default, Acorn workloads that publish ports that use HTTP will be allowed to 
 To allow traffic from a specific namespace to all Acorn apps in the cluster, use `--allow-traffic-from-namespace=<namespace>`. This is useful if there is a monitoring namespace, for example, that needs to be able to connect to all the pods created by Acorn in order to scrape metrics.
 
 ## Working with external LoadBalancer controllers
-If you are using an external `LoadBalancer` controller that requires annotations on `LoadBalancer` Services to operate, such as the `aws-load-balancer-controller`, you can pass the `--service-lb-annotations` flag to `acorn install`. This will cause Acorn to add the specified annotations to all `LoadBalancer` Services it creates. The value of the flag should be a comma-separated list of key-value pairs, where the key is the annotation name and the value is the annotation value. For example:
+If you are using an external `LoadBalancer` controller that requires annotations on `LoadBalancer` Services to operate, such as the `aws-load-balancer-controller`, you can pass the `--service-lb-annotation` flag to `acorn install`. This will cause Acorn to add the specified annotations to all `LoadBalancer` Services it creates. The value of the flag should be a comma-separated list of key-value pairs, where the key is the annotation name and the value is the annotation value. For example:
 
 ```bash
-acorn install --service-lb-annotations service.beta.kubernetes.io/aws-load-balancer-type=external,service.beta.kubernetes.io/aws-load-balancer-scheme=internet-facing,service.beta.kubernetes.io/aws-load-balancer-nlb-target-type=instance
+acorn install --service-lb-annotation service.beta.kubernetes.io/aws-load-balancer-type=external,service.beta.kubernetes.io/aws-load-balancer-scheme=internet-facing,service.beta.kubernetes.io/aws-load-balancer-nlb-target-type=instance
 ```
 
 For readability, you can also pass the flag multiple times, and the values will be concatenated. For example:
 
 ```bash
-acorn install
-    --service-lb-annotations service.beta.kubernetes.io/aws-load-balancer-type=external
-    --service-lb-annotations service.beta.kubernetes.io/aws-load-balancer-scheme=internet-facing
-    --service-lb-annotations service.beta.kubernetes.io/aws-load-balancer-nlb-target-type=instance
+acorn install \
+    --service-lb-annotation service.beta.kubernetes.io/aws-load-balancer-type=external \
+    --service-lb-annotation service.beta.kubernetes.io/aws-load-balancer-scheme=internet-facing \
+    --service-lb-annotation service.beta.kubernetes.io/aws-load-balancer-nlb-target-type=instance \
+```
+
+Lastly, you can unset the annotations defined by the `--service-lb-annotation` flag by passing an empty string to the flag. For example:
+
+```bash
+acorn install --service-lb-annotation ""
 ```
 
 :::note
--- a/pkg/apis/api.acorn.io/v1/types.go
+++ b/pkg/apis/api.acorn.io/v1/types.go
@@ -336,35 +336,35 @@ type Config struct {
 	// For repeatable flags, ensure the struct and json fields are plural and the flag name is singular.
 	// See ClusterDomains as an example.
 
-	IngressClassName               *string           `json:"ingressClassName" usage:"The ingress class name to assign to all created ingress resources (default '')"`
-	ClusterDomains                 []string          `json:"clusterDomains" name:"cluster-domain" usage:"The externally addressable cluster domain (default .on-acorn.io)"`
-	LetsEncrypt                    *string           `json:"letsEncrypt" name:"lets-encrypt" usage:"enabled|disabled|staging. If enabled, acorn generated endpoints will be secured using TLS certificate from Let's Encrypt. Staging uses Let's Encrypt's staging environment. (default disabled)"`
-	LetsEncryptEmail               string            `json:"letsEncryptEmail" name:"lets-encrypt-email" usage:"Required if --lets-encrypt=enabled. The email address to use for Let's Encrypt registration(default '')"`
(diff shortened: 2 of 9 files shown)


docs/docs/30-installation/02-options.md
```diff
@@ -117,6 +117,12 @@ acorn install
     --service-lb-annotations service.beta.kubernetes.io/aws-load-balancer-nlb-target-type=instance
 ```
 
+Lastly, you can unset the annotations defined by the `--service-lb-annotations` flag by passing an empty annotation to the flag. For example:
+
+```bash
+acorn install --service-lb-annotations " "=" "
```
tylerslaton, line 123: I couldn't find anything immediately easier than this when working with maps. The rest of our CLI doesn't currently use maps as flags instead opting for parsing logic that runs against string flags. I'd love a better approach here if we can think of one.

pkg/config/config.go
```diff
@@ -316,9 +324,15 @@ func merge(oldConfig, newConfig *apiv1.Config) *apiv1.Config {
 		mergedConfig.PropagateProjectLabels = newConfig.PropagateProjectLabels
 	}
 
-	if len(newConfig.ServiceLBAnnotations) == 0 {
+	if len(newConfig.AllowTrafficFromNamespace) > 0 && newConfig.AllowTrafficFromNamespace[0] == "" {
+		mergedConfig.AllowTrafficFromNamespace = nil
+	} else if len(newConfig.PropagateProjectLabels) > 0 {
```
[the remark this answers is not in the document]
tylerslaton: 👀

pkg/apis/api.acorn.io/v1/types.go
```diff
@@ -336,35 +336,35 @@ type Config struct {
...
+	ManageVolumeClasses            *bool    `json:"manageVolumeClasses" name:"manage-volume-classes" usage:"Manually manage volume classes rather than sync with storage classes, setting to 'true' will delete Acorn-created volume classes"`
+	NetworkPolicies                *bool    `json:"networkPolicies" name:"network-policies" usage:"Create Kubernetes NetworkPolicies which block cross-project network traffic (default true)"`
+	IngressControllerNamespace     *string  `json:"ingressControllerNamespace" name:"ingress-controller-namespace" usage:"The namespace where the ingress controller runs - used to secure published HTTP ports with NetworkPolicies."`
+	AllowTrafficFromNamespace      []string `json:"allowTrafficFromNamespace" name:"allow-traffic-from-namespace" usage:"Namespaces that are allowed to send network traffic to all Acorn apps"`
+	ServiceLBAnnotations           []string `json:"serviceLBAnnotations" name:"service-lb-annotations" usage:"Annotations to add to the service of type LoadBalancer. Defaults to empty. (example key=value)"`
```
StrongMonkey, line 367: the name has to be singular format to match UX consistency, so `service-lb-annotation`

docs/docs/30-installation/02-options.md
```diff
@@ -111,10 +111,16 @@ acorn install --service-lb-annotations service.beta.kubernetes.io/aws-load-balan
 For readability, you can also pass the flag multiple times, and the values will be concatenated. For example:
 
 ```bash
-acorn install
-    --service-lb-annotations service.beta.kubernetes.io/aws-load-balancer-type=external
-    --service-lb-annotations service.beta.kubernetes.io/aws-load-balancer-scheme=internet-facing
-    --service-lb-annotations service.beta.kubernetes.io/aws-load-balancer-nlb-target-type=instance
+acorn install \
+    --service-lb-annotations service.beta.kubernetes.io/aws-load-balancer-type=external \
```
StrongMonkey, line 115:
```suggestion
    --service-lb-annotation service.beta.kubernetes.io/aws-load-balancer-type=external \
```

pkg/config/config_test.go
cjellick, line 1: Why do you have test additions for AllowTrafficFromNamespace but not for the service lb annotation field?

pkg/publish/servicelb.go
```diff
@@ -45,8 +48,14 @@ func ServiceLoadBalancer(req router.Request, svc *v1.ServiceInstance) (result []
 		svc.Spec.Annotations = map[string]string{}
 	}
 
-	for _, entry := range typed.Sorted(cfg.ServiceLBAnnotations) {
-		svc.Spec.Annotations[entry.Key] = entry.Value
+	sort.Strings(cfg.ServiceLBAnnotations)
+	for _, annotation := range cfg.ServiceLBAnnotations {
+		split := strings.Split(annotation, "=")
```
StrongMonkey, line 53: you can just do `strings.Cut`. I think value can have `=`?

pkg/config/config.go
```diff
@@ -316,9 +324,15 @@ func merge(oldConfig, newConfig *apiv1.Config) *apiv1.Config {
 		mergedConfig.PropagateProjectLabels = newConfig.PropagateProjectLabels
 	}
 
-	if len(newConfig.ServiceLBAnnotations) == 0 {
+	if len(newConfig.AllowTrafficFromNamespace) > 0 && newConfig.AllowTrafficFromNamespace[0] == "" {
+		mergedConfig.AllowTrafficFromNamespace = nil
+	} else if len(newConfig.AllowTrafficFromNamespace) > 0 {
+		mergedConfig.AllowTrafficFromNamespace = newConfig.AllowTrafficFromNamespace
+	}
```
thedadams, line 331: nit:
```suggestion
	if len(newConfig.AllowTrafficFromNamespace) == 0 || newConfig.AllowTrafficFromNamespace[0] == "" {
		mergedConfig.AllowTrafficFromNamespace = nil
	} else {
		mergedConfig.AllowTrafficFromNamespace = newConfig.AllowTrafficFromNamespace
	}
```

pkg/install/install.go
```diff
@@ -258,6 +262,16 @@ func Install(ctx context.Context, image string, opts *Options) error {
 	return nil
 }
 
+func validateServiceLBAnnotations(annotations []string) error {
+	for _, annotation := range annotations {
+		split := strings.Split(annotation, "=")
```
thedadams, line 267: Is it valid for the value of an annotation to have an `=`? If so, I think `strings.Cut` here makes sense instead.

pkg/publish/servicelb.go
```diff
@@ -45,8 +48,14 @@ func ServiceLoadBalancer(req router.Request, svc *v1.ServiceInstance) (result []
 		svc.Spec.Annotations = map[string]string{}
 	}
 
-	for _, entry := range typed.Sorted(cfg.ServiceLBAnnotations) {
-		svc.Spec.Annotations[entry.Key] = entry.Value
+	sort.Strings(cfg.ServiceLBAnnotations)
```
thedadams, line 51: No need to sort the slice. The result will be consistent because the items in the slice come out in the same order each time.

pkg/install/install.go
```diff
@@ -258,6 +262,16 @@ func Install(ctx context.Context, image string, opts *Options) error {
...
+		if len(split) != 2 {
```
cjellick, line 268: `=` should be a valid, supported value in the value part of the annotation.

pkg/publish/servicelb.go
thedadams, line 53: Same question about `strings.Cut` here.

pkg/install/install.go
cjellick: (I think we can just use strings.SplitN)

pkg/publish/servicelb.go
cjellick, line 53: same as above

pkg/config/config.go
cjellick: Are you sure @thedadams ? Look at the above logic for similar scenarios:
```
if len(newConfig.PropagateProjectAnnotations) > 0 && newConfig.PropagateProjectAnnotations[0] == "" {
```

you cannot set mergedConfig.property to nil if the newConfig.property is just length zero. That would cause values to get stomped on on subsequent runs of acorn install if all the original flags are not reprovided, which is not the desired behavior

cjellick: So our reset logic for slice type arguments is to pass the argument with a single empty value:
```
acorn install --propogate-project-annotation ""
```

thedadams: Yep. My mistake. Thanks for double-checking me here.

pkg/install/install.go
tylerslaton: I opted for using `strings.Cut` here.

pkg/install/install.go
tylerslaton: Yep! Its valid. Great catch.

pkg/config/config_test.go
tylerslaton: Fixed!

StrongMonkey (approved):
💯

thedadams (approved):

tylerslaton:
Hitting rate limiting for a image we didn't add to the mirror.

cjellick (approved):

Issue #1428, opened 2023-04-04
App get auto updated to image versions  that donot satisfy imageAllowRules.


sangee2004:
Acorn version - v0.6.0-106-g09f636a2+09f636a2

Steps to reproduce the problem:
1. Create imageAllowRules for the project.
2. Build and push "latest" tag for image.
```acorn build -t registry-myreg-f5d03834.a-sangee2004-acorn-b4c4fc02-paul1.stg-on-acorn.io/mytest2:latest --push .```
3. Cosign the image with all the rules from imageAllowRules  created in step1.
 ```cosign sign --key santest2.key -a qatag=ok registry-myreg-f5d03834.a-sangee2004-acorn-b4c4fc02-paul1.stg-on-acorn.io/mytest2:latest```
4. Deploy app with autoupgrade option - `acorn run --auto-upgrade --interval 15s -n testlatestnew registry-myreg-f5d03834.a-sangee2004-acorn-b4c4fc02-paul1.stg-on-acorn.io/mytest2:latest`

5.App gets deployed with the latest version of the image as expected.

6. Build and push a new version of  "latest" tag for image.

I have not yet co-signed the image.

App gets redeployed with the latest version of the image.  There is an error message presented in the "Message" column.

```
acorn apps                                                 
NAME            IMAGE                                                                                      HEALTHY   UP-TO-DATE   CREATED     ENDPOINTS                                                             MESSAGE
testlatestnew   registry-myreg-f5d03834.a-sangee2004-acorn-b4c4fc02-paul1.stg-on-acorn.io/mytest2:latest   1         1            3m23s ago   http://myweb1-testlatestnew-7beff6a7.local.on-acorn.io => myweb1:80   [image-allowed: failed to check if image is allowed: error ensuring references for image registry-myreg-f5d03834.a-sangee2004-acorn-b4c4fc02-paul1.stg-on-acorn.io/mytest2:latest: failed to ensure signature artifact: no matching signatures: expected signature artifact registry-myreg-f5d03834.a-sangee2004-acorn-b4c4fc02-paul1.stg-on-acorn.io/mytest2:sha256-eff1495068d10176fe60eec5298fb8b8b24e1e85c69c7daebf6c64de44ae97c3.sig not found]
```

Expected Behavior:
App should not get deployed with the latest version which does not satisfy imageAllowRules.

Note:
This happens only when using --auto-upgrade  option and pushing the same image tag.
[...]

PR #1451, opened 2023-04-06
change: also check imageAllowRules for auto-upgrade without pattern (#1428)

iwilltry42:
Ref #1428

@cjellick & @thedadams I guess when the nextAppImage fails validation, we should still set a new updateTime, so it won't keep trying forever?

### Checklist

### DIFF
--- a/pkg/autoupgrade/daemon.go
+++ b/pkg/autoupgrade/daemon.go
@@ -227,7 +227,7 @@ func (d *daemon) refreshImages(ctx context.Context, apps map[kclient.ObjectKey]v
 				if !updated && digest != "" {
 					if err := d.client.checkImageAllowed(ctx, app.Namespace, nextAppImage); err != nil {
 						if errors.Is(err, &imageallowrules.ErrImageNotAllowed{}) {
-							logrus.Warnf("Updated image %s for %s/%s is not allowed: %v", nextAppImage, app.Namespace, app.Name, err)
+							logrus.Debugf("Updated image %s for %s/%s is not allowed: %v", nextAppImage, app.Namespace, app.Name, err)
 							continue
 						}
 						logrus.Errorf("error checking if updated image %s for %s/%s  is allowed: %v", app.Namespace, app.Name, nextAppImage, err)
--- a/pkg/autoupgrade/daemon.go
+++ b/pkg/autoupgrade/daemon.go
@@ -228,6 +228,7 @@ func (d *daemon) refreshImages(ctx context.Context, apps map[kclient.ObjectKey]v
 					if err := d.client.checkImageAllowed(ctx, app.Namespace, nextAppImage); err != nil {
 						if errors.Is(err, &imageallowrules.ErrImageNotAllowed{}) {
 							logrus.Debugf("Updated image %s for %s/%s is not allowed: %v", nextAppImage, app.Namespace, app.Name, err)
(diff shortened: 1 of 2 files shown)


pkg/autoupgrade/daemon.go
```diff
@@ -222,6 +224,17 @@ func (d *daemon) refreshImages(ctx context.Context, apps map[kclient.ObjectKey]v
 			}
 
 			if updated || strings.TrimPrefix(app.Status.AppImage.Digest, "sha256:") != strings.TrimPrefix(digest, "sha256:") {
+				if !updated && digest != "" {
```
cjellick, line 227: why do you need to check for a non empty digest here?

pkg/autoupgrade/daemon.go
```diff
@@ -222,6 +224,17 @@ func (d *daemon) refreshImages(ctx context.Context, apps map[kclient.ObjectKey]v
...
+					if err := d.client.checkImageAllowed(ctx, app.Namespace, nextAppImage); err != nil {
+						if errors.Is(err, &imageallowrules.ErrImageNotAllowed{}) {
+							logrus.Warnf("Updated image %s for %s/%s is not allowed: %v", nextAppImage, app.Namespace, app.Name, err)
```
cjellick, line 230: i think this should be info level, not warn. To me, warn indicates something is wrong with the app. In this case, nothing wrong, simply the image doesnt match criteria.

cjellick:
@iwilltry42 - @thedadams has a pretty good test suite for the the sync() function, including an easy to use mock client that can hide the complexity of the cosign bits.

Can you add a test case for this situation (and while you're in there, if can identify any other low-hanging fruit test cases related to imageAllowRules would be great. Don't have to spend too much time on the additional test cases).

cjellick (changes requested):
see my inline comments

pkg/autoupgrade/daemon.go
iwilltry42: We don't **need** to.
However, if the digest is empty, that means that we didn't find one above, thus the `checkImageAllowed` function should also fail or just be unnecessary, because it's working on the digest.
Low impact though.

pkg/autoupgrade/daemon.go
iwilltry42: Potentially even Debug level, just forgot to change it.
Info level could already be too noisy.
```suggestion
							logrus.Debugf("Updated image %s for %s/%s is not allowed: %v", nextAppImage, app.Namespace, app.Name, err)
```

iwilltry42:
[Copy Paste from description]
@cjellick & @thedadams I guess when the nextAppImage fails validation, we should still set a new updateTime, so it won't keep trying forever?

thedadams:
@iwilltry42 I would say that we should set a new updateTime because we did check for a new image and there wasn't one.

Now that you bring it up, I think we should also be setting a new updateTime in the `enabled` and `notify` cases when we don't update the `app. What do you think about this?

iwilltry42:
@thedadams +1 for your suggestion.
I think it makes sense in every case that is not an error during the check.

iwilltry42:
@cjellick seems like @yashgiri already added some great test cases there, so I only added the simple one using no pattern.

@thedadams Craig agrees with you as well, so I added the updated time stamp to the discussed cases.

cjellick (approved):

PR #1461, opened 2023-04-07
Ensure app is removed before deleting other objects

thedadams:
There are situations when using `acorn rm -af` where the other objects are deleted before the app is deleted. This causes things like generated secrets to be regenerated and left around unexpectedly.

This change introduces a wait for app deletion that only happens when the other objects are also being deleted. This ensures that the other objects will be successfully removed.

### Checklist


g-linville (approved):

njhale (approved):
Some questions/suggestions, but I don't think they should hold up this merging.

pkg/cli/rm_helper.go
```diff
@@ -159,11 +160,21 @@ func removeApp(arg string, c client.Client, cmd *cobra.Command, force bool) erro
...
 	if err != nil {
 		return fmt.Errorf("deleting app %s: %w", arg, err)
 	}
+
 	if app != nil {
+		if wait {
+			fmt.Printf("Waiting for app %s to be removed...\n", arg)
+			for {
+				if _, err = c.AppGet(cmd.Context(), arg); apierrors.IsNotFound(err) {
+					break
```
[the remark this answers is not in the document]
thedadams: You're right. I need to add something to handle when the context is canceled.

pkg/cli/rm_helper.go
```diff
@@ -159,11 +160,21 @@ func removeApp(arg string, c client.Client, cmd *cobra.Command, force bool) erro
...
+				}
+				time.Sleep(time.Second)
+			}
```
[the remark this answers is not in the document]
thedadams: Done. Thanks for the suggestion.

pkg/cli/rm_helper.go
```diff
@@ -150,12 +151,31 @@ func removeApp(arg string, c client.Client, cmd *cobra.Command, force bool) erro
...
+				case <-cmd.Context().Done():
+					return cmd.Context().Err()
+				default:
+					if _, err = c.AppGet(cmd.Context(), arg); apierrors.IsNotFound(err) {
+						return nil
+					}
+				}
+				time.Sleep(time.Second)
+			}
+		}
```
njhale, line 178: nit: just thought of this, but we could reduce nesting a lot here

```go
func removeApp(arg string, c client.Client, cmd *cobra.Command, force, wait bool) error {
	// ...
	if app == nil {
	  fmt.Printf("Error: No such app: %s\n", arg)
	  return nil
	}
	
	defer func() {
      // Ensure this gets printed whether we wait or not
      if err == nil || apierrors.IsNotFound(err) {
        fmt.Println("Removed: " + arg)
	  }
    }()
	
	if !wait {
	  // Don't wait for the app to be removed before returning
	  return nil
    }
    
    // Wait for app to be removed before returning
    fmt.Printf("Waiting for app %s to be removed...\n", arg)
    for {
      // ...
    }
    
	return nil
}
```

pkg/cli/rm_helper.go
```diff
@@ -150,12 +151,31 @@ func removeApp(arg string, c client.Client, cmd *cobra.Command, force bool) erro
...
+		if wait {
+			fmt.Printf("Waiting for app %s to be removed...\n", arg)
+			for {
+				select {
+				case <-cmd.Context().Done():
+					return cmd.Context().Err()
+				default:
+					if _, err = c.AppGet(cmd.Context(), arg); apierrors.IsNotFound(err) {
+						return nil
+					}
```
njhale, line 174: I think we also want to return an error -- or at least print one --  when `err != nil && !err.IsNotFound(err)`; i.e. any error other than NOT_FOUND.

njhale (approved):

pkg/cli/rm_helper.go
thedadams: Logging at debug level.

pkg/cli/rm_helper.go
```diff
@@ -150,14 +152,36 @@ func removeApp(arg string, c client.Client, cmd *cobra.Command, force bool) erro
...
+		fmt.Printf("Waiting for app %s to be removed...\n", arg)
+		for {
+			select {
+			case <-cmd.Context().Done():
+				return cmd.Context().Err()
+			default:
+				if _, err = c.AppGet(cmd.Context(), arg); apierrors.IsNotFound(err) {
+					return nil
+				} else if err != nil {
+					logrus.Debugf("Error getting app for removal check: %v", err)
```
njhale, line 180: @thedadams and I talked about this on Slack. Here's the discussion for posterity:

>  We are trying to abstract Kubernetes away as much as possible, so printing the error seemed counter to that goal.

That's a great point.

I think acorn should probably say something if it encounters an error here. For the short term, I feel like that could just be some boilerplate "command failed" message; but long-term, we'll probably want to give the user a more useful message.

Zooming out at the big picture, for any `$PLATFORM` that `acorn` supports -- here `PLATFORM=k8s` -- there are two primary approaches that jump out at me:
A) Maintain mappings between user-facing `acorn` error types and `$PLATFORM` error types
B) Be transparent with `$PLATFORM` error types

and besides the obvious toil on our end, my intuition is that there are cases where `A` isn't possible without leaking `$PLATFORM` specific details into `acorn` error types anyway.

I think `B` still holds up the abstraction goal as long as all platforms get the same treatment; i.e. none of their errors becomes codified in `acorn`'s top-level error types.

so I'd probably opt for `B`, but hide the output behind the --debug flag.

njhale (approved):
Sweet, looks great!

Issue #1362, opened 2023-03-21
accept project in resource prefix for any CLI command


jacobdonenfeld:
Allow a resource to be specified as <project>::<resource> and parse it to scope into the specified project.

Adding expanded scope of the `::` the separator to be consistent across resource types such that anywhere a resource is passed in, the fully qualified version works as expected.

PR #1469, opened 2023-04-11
accept project::resource format, fix acorn images -A (#1362)

jacobdonenfeld:
This adds the ability to use the project::resource format for cross project resources, corresponding cli autocomplete, as well as fixes acorn images -A to list cross project images.

### Overview:

The following changes were done:
- cli/image.go
  - had to parse out project:: before the resource, because images can be looked up via tag, SHA, or prefix, and libraries used for parsing those would not work with the string beginning with project::
  - after calling ImageList, images objects are returned from the client as project::name, which needed to be parsed out and appended to the repository instead.
- added autocomplete for project::resource format. Works as a function that wraps around another autocomplete function. Requires using that function on each cli file's completion
- multiclient.go:
  - Made a function to replace most calls to ForProject within multi.go, that decides which project to return based on the current default project, and whether the input was in project::resource format. This also strips off the project from the input and returns the new input.
  - modified listing functions to return the name in project::resource format when called with -A
- added ./bin to the path before calling make test in CI, to add the acorn binary to path. This was needed for tests for this PR, as they rely on cli calls to the acorn binary
- refactored integration/client/client.go to clean up images at the end of the test, as well as cleaned up some repeated code.
- cleaned up code in integration/run/run_test.go while debugging why some tests were hanging
- added docstings to functions I found myself referencing often, so my IDE could intelligently display those comments
- Corrected spelling on some comments throughout the repository

### Discussion Points:

1. Could add more tests throughout more cli commands. Integration tests could be expanded upon to use more CLI commands.
1. This increases CLI autocomplete runtime. 
1. This won't autocomplete `projectname:` to fill in the extra `:`, is that okay?
1. Testing project autocomplete generally is very difficult because it generates new clients, so passing in the mock client doesn't do anything, outside of a full integration test.

#### Tagging an image using cross-project naming, adds the new tag on the cross-project specified, not the current project, but only works with the ID/SHA.

> acorn tag project-x::image1 newname:newtag
> ✗ ERROR: images.api.acorn.io "image1" not found
> 
> acorn tag project-x::xyz newname:newtag
> 
>     Creates a new tag in project-x newname:newtag
> 
> acorn tag project-x::xyz project-y::newname:newtag
> 
>     ✗ ERROR: could not parse reference: ...

#### Acorn run with project syntax only works if you're specifying the same project you are already on.
[Details](https://github.com/acorn-io/acorn/pull/1469#issuecomment-1524480528)

> Future ideas, could prompt user if they'd like to switch projects and run it in the specified project, or have some flow to pull it from one project into the current project and then run it.

### Checklist

Addresses #1362 and #1185


pkg/cli/completion.go
```diff
@@ -85,6 +86,98 @@ func (a *completion) complete(cmd *cobra.Command, args []string, toComplete stri
...
+			fmt.Printf("could not read CLI config.\n%+v", err)
+		}
+	}
+	newCompletionFunc := func(ctx context.Context, c client.Client, toComplete string) ([]string, error) {
+		if parsedProject, after, found := strings.Cut(toComplete, "::"); found {
+			// check that the provided project exists, return a helpful error otherwise
+			foundProject, err := c.ProjectGet(ctx, parsedProject)
+			if foundProject == nil || err != nil {
+				return []string{}, nil
+			}
```
g-linville, line 110: The comment says that this is returning an error, but the `err` value being returned here is nil.

pkg/cli/completion.go
```diff
@@ -85,6 +86,98 @@ func (a *completion) complete(cmd *cobra.Command, args []string, toComplete stri
...
+		if err != nil {
+			fmt.Printf("could not read CLI config.\n%+v", err)
+		}
+	}
+	newCompletionFunc := func(ctx context.Context, c client.Client, toComplete string) ([]string, error) {
+		if parsedProject, after, found := strings.Cut(toComplete, "::"); found {
+			// check that the provided project exists, return a helpful error otherwise
+			foundProject, err := c.ProjectGet(ctx, parsedProject)
+			if foundProject == nil || err != nil {
+				return []string{}, nil
```
tylerslaton, line 109: We should be returning the error in this case, right?
```suggestion
				return []string{}, err
```

pkg/cli/completion.go
```diff
@@ -85,6 +86,98 @@ func (a *completion) complete(cmd *cobra.Command, args []string, toComplete stri
...
+	options := a.client.Options()
+	if options.CLIConfig == nil {
+		cfg, err := config.ReadCLIConfig()
+		options.CLIConfig = cfg
+		if err != nil {
+			fmt.Printf("could not read CLI config.\n%+v", err)
+		}
+	}
+	newCompletionFunc := func(ctx context.Context, c client.Client, toComplete string) ([]string, error) {
+		if parsedProject, after, found := strings.Cut(toComplete, "::"); found {
```
tylerslaton, line 105: nit: We can shorten this down to just project
```suggestion
		if project, after, found := strings.Cut(toComplete, "::"); found {
```

pkg/cli/completion.go
```diff
@@ -85,6 +86,98 @@ func (a *completion) complete(cmd *cobra.Command, args []string, toComplete stri
...
+			if foundProject == nil || err != nil {
+				return []string{}, nil
+			}
+
+			// > completion for the string after the ::
+			// create a new client scoped for the provided project name
+			options.Project = parsedProject
+			options.AllProjects = false
+			parsedProjectScopedClient, err := project.Client(ctx,
+				options)
```
tylerslaton, line 117: nit:
```suggestion
			parsedProjectScopedClient, err := project.Client(ctx, options)
```

pkg/cli/completion.go
```diff
@@ -85,6 +86,98 @@ func (a *completion) complete(cmd *cobra.Command, args []string, toComplete stri
...
+			// If we were doing a list operation cross project, all values except those that match the current project
+			// would already have project::resource format. However, since we specify a specific project, we know
+			// it will only have the resource form and no project.
+			if len(completions) > 0 {
+				for i := range completions {
+					completions[i] = parsedProject + "::" + completions[i]
+				}
+			}
+			return completions, nil
+		} else {
```
tylerslaton, line 137: Since the `if` block ends in a return, the `else` is not needed.

pkg/cli/images.go
```diff
@@ -46,12 +47,19 @@ func (a *Image) Run(cmd *cobra.Command, args []string) error {
...
 
+	allProjects := a.client.Options().AllProjects
 	var images []apiv1.Image
 	var image *apiv1.Image
 	tagToMatch := ""
 	repoToMatch := ""
 
+	// Image was explicitly provided
 	if len(args) == 1 {
+		c, args[0], err = parseArgGetClient(a.client, cmd, args[0])
```
g-linville, line 58: Assigning a value to an item in the `args` array seems weird to me. Might be better to declare a new variable to hold that.

pkg/cli/images_all_test.go
```diff
@@ -0,0 +1,246 @@
...
+		},
+	}
+
+	image2Tagged := image2Untagged
+	image2Tagged.Tags = []string{"testtag2:latest2"}
+
+	image3Untagged := apiv1.Image{
+		TypeMeta: metav1.TypeMeta{},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "acorn.com/acc/Name3::image-3-129394",
```
g-linville, line 49: I'd suggest using `example.com` here. I know that in simple testing like this, nothing is going to try to reach out to that domain, but it's probably be best to avoid using something like that that we don't own. (Not that we own example.com, but it's meant to be used as an example of course.)

pkg/cli/images_all_test.go
```diff
@@ -0,0 +1,246 @@
...
+			Namespace: "Name3",
+		},
+	}
+
+	image3Tagged := image3Untagged
+	image3Tagged.Tags = []string{"testtag3:latest3"}
+
+	allImagesHaveContaines := func(f *mocks.MockClient) {
```
g-linville, line 57: typo/refactor:

```suggestion
	allImagesHaveContainers := func(f *mocks.MockClient) {
```

pkg/client/multi.go
```diff
@@ -40,6 +40,22 @@ func aggregate[T any, V ObjectPointer[T]](ctx context.Context, factory ProjectCl
 	return aggregateOptionalNaming[T, V](ctx, true, factory, cb)
 }
 
+// parseProject accepts an argument name that is either of the form resource or project::resource, and ClientFactory
+// If the name is not of project::resource form, it returns defaultProject, name, nil
+// otherwise, it parses off the project name and returns project, name without project::, nil
+func parseProject(ctx context.Context, name string, factory ProjectClientFactory) (pc Client, parsedArg string, err error) {
```
tylerslaton, line 46: nit: Slight rewording to make it easier to follow
```suggestion
// parseProject accepts a name string (in the form of resource or project::resource) and a ClientFactory.
// If the name string is not of project::resource form, it returns a client for the default project, the name, and
// nil. Otherwise, it parses off the project name and returns a client for that project, the name without project::, 
// and nil
func parseProject(ctx context.Context, name string, factory ProjectClientFactory) (pc Client, parsedArg string, err error) {
```

pkg/cli/images_all_test.go
```diff
@@ -0,0 +1,246 @@
...
+			} else if err != nil && tt.wantErr {
+				assert.Equal(t, tt.wantOut, err.Error())
+			} else {
+				err := w.Close()
+				if err != nil {
+					t.Fatal(err)
+				}
+				out, _ := io.ReadAll(r)
+				assert.Equal(t, tt.wantOut, string(out))
+			}
```
g-linville, line 243: Doesn't look like this is checking for the case where we wanted an error but did not get one.

pkg/client/multi.go
```diff
@@ -40,6 +40,22 @@ func aggregate[T any, V ObjectPointer[T]](ctx context.Context, factory ProjectCl
...
+	var parsedProject string
+	if possibleProject, possibleArg, found := strings.Cut(name, "::"); found {
+		parsedProject = possibleProject
+		parsedArg = possibleArg
+	} else {
+		parsedProject = factory.DefaultProject()
+		parsedArg = name
+	}
+	pc, err = factory.ForProject(ctx, parsedProject)
+	return pc, parsedArg, err
```
g-linville, line 56: No need to do an explicit return here since you are using named return variables (or whatever the term for that is, lol).

pkg/client/multi.go
```diff
@@ -68,15 +84,10 @@ func isNil(obj kclient.Object) bool {
...
+		result []T
 	)
-	i := strings.LastIndex(name, "/")
-	if i != -1 {
-		projectName = name[0:i]
-		name = name[i+1:]
-	}
-	client, err := factory.ForProject(ctx, projectName)
+	// parse the form project::resource
+	client, name, err := parseProject(ctx, name, factory)
```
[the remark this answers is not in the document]
tylerslaton: The `ForFoo` probably doesn't make as much sense here since the `factory` prefix is no longer used. If I had to make a call here, I'd pick something along the lines of `resourceClient`.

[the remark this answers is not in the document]
tylerslaton: Alternatively, you could add the `parseProject` function into the `factory` package and change the name to `ForResource`.

pkg/cli/context.go
```diff
@@ -38,3 +41,52 @@ func (c *CommandClientFactory) Options() project.Options {
...
+	localClient, err := localClientFactory.CreateDefault()
+	if err != nil {
+		return nil, "", err
+	}
+	return localClient, arg, nil
+}
+
+// parseProjectsOffName parses the projectName::actualName format,
+// strips it, and adds it onto the AcornProjectName label
+func parseProjectsOffName(images []apiv1.Image, err error) ([]apiv1.Image, error) {
```
tylerslaton, line 71: Could we rename this to be a bit clearer that it is specific to images?

pkg/cli/completion_test.go
```diff
@@ -17,6 +17,62 @@ import (
...
+		wantNames     []string
+		wantDirective cobra.ShellCompDirective
+	}{
+		{
+			name:          "Project Prefix Doesn't Exist",
+			toComplete:    "nonexist::tatatat",
+			wantNames:     []string{},
+			wantDirective: cobra.ShellCompDirectiveNoFileComp,
+		},
+		{
```
tylerslaton, line 52: What about a test case where it autocompletes a project without the `::`?

integration/client/apps/update_test.go
```diff
@@ -30,7 +30,7 @@ func TestUpdatePull(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	app, err := c.AppRun(ctx, "foo", &client.AppRunOptions{Name: "test"})
+	app, err := c.AppRun(ctx, "foo:latest", &client.AppRunOptions{Name: "test"})
```
tylerslaton, line 33: What is the necessity of adding `latest` here?

pkg/cli/images.go
tylerslaton, line 58: If you do end up moving the `parseProject` resource to the factory package and renaming it to `factory.ForResource`, could you use it here as well?

tylerslaton (commented):
Can we create a constant that is used everywhere for the `::` token? This will make moving from it less of a pain should we need to and it keep things consistent.

pkg/client/multi.go
```diff
@@ -324,63 +333,69 @@ func (m *MultiClient) VolumeDelete(ctx context.Context, name string) (*apiv1.Vol
...
 func (m *MultiClient) ImageTag(ctx context.Context, image, tag string) error {
-	c, err := m.Factory.ForProject(ctx, m.Factory.DefaultProject())
+	c, image, err := parseProject(ctx, image, m.Factory)
 	if err != nil {
 		return err
 	}
 	return c.ImageTag(ctx, image, tag)
 }
 
 func (m *MultiClient) ImageDetails(ctx context.Context, imageName string, opts *ImageDetailsOptions) (result *ImageDetails, err error) {
```
g-linville, line 381: Question: on this function, should we be looking for and parsing the `project::` prefix?

g-linville (commented):
Sorry if any of my comments are duplicates of something someone else said - I saw that other comments were coming in while I was in the middle of reviewing.

integration/helper/namespace.go
```diff
@@ -84,6 +49,7 @@ func tempCreateHelper(t *testing.T, client client.Client, namespaceObject corev1
 
 	t.Cleanup(func() {
 		namespaceDeleting := namespaceObject.Name
+		_ = client.Get(ctx, k8sclient.ObjectKey{Name: namespaceObject.Name}, &namespaceObject)
```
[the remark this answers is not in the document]
jacobdonenfeld: Thought it could not be worth handling errors in the cleanup function. But could be useful doing to print a message if it does fail to clean up. I'll add something!

pkg/cli/completion.go
```diff
@@ -85,6 +86,98 @@ func (a *completion) complete(cmd *cobra.Command, args []string, toComplete stri
 	return removeExistingArgs(result, args), a.successDirective
 }
 
+// checkProjectPrefix
+// called on a completion object, which already has the completion function defined
+// this function modifies that completion function, wraps it with some logic to handle projects
+func (a *completion) checkProjectPrefix() *completion {
```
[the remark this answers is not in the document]
jacobdonenfeld: Yes, this can be refactored into the .complete(). There are maybe 2 or so completions where we do not want project completion however. Can make a condition to disable it, and have it enabled by default. Great idea, thanks!

pkg/cli/context.go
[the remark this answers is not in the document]
jacobdonenfeld: will do

cjellick:
This use case doesn't work, but should:
```
$ acorn run hub.stg.acornlabs.com/cjellick/kc-demo-us::cj/simpledemo:1
  ✗  ERROR:  could not parse reference: hub.stg.acornlabs.com/cjellick/kc-demo-us::cj/simpledemo:1
```
In this scenario, I am set to use the kc-demo-us project
```
NAME                                         DEFAULT   REGIONS
acorn                                                  local*
foo                                                    local*
hub.stg.acornlabs.com/cjellick/kc-demo-all             aws-eu-north-1*,aws-us-east-2,sandbox
hub.stg.acornlabs.com/cjellick/kc-demo-eu              aws-eu-north-1*
hub.stg.acornlabs.com/cjellick/kc-demo-us    *         aws-us-east-2*
hub.stg.acornlabs.com/cjellick/kc-sandbox1             sandbox*
```
So, I'm not crossing project boundaries or anything like that.

integration/client/images/images_test.go
```diff
@@ -409,3 +411,54 @@ func TestImageBadTag(t *testing.T) {
...
+	ctx := helper.GetCTX(t)
+	kclient := helper.MustReturn(kclient.Default)
+	ns := helper.TempProject(t, kclient)
+
+	c, err := client.New(restConfig, ns.Name, ns.Name)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	_ = client2.NewImage(t, ns.Name)
```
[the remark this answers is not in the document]
jacobdonenfeld: Yes. Client2 either returns the image ID, or calls t.Fatal inside the function. I chose not to receive the ID.

pkg/cli/completion.go
jacobdonenfeld: Hm, curious what we'd actually want in this case. 

Say a user is autocompleting DoesNotExist::ok, should this throw an error, or just fail to autocomplete?

It currently doesn't return an error, to have the function fail to autocomplete without an error.

pkg/cli/completion.go
jacobdonenfeld: that would shadow "github.com/acorn-io/acorn/pkg/project" unfortunately

pkg/cli/completion.go
jacobdonenfeld: ty

integration/client/client.go
```diff
@@ -37,11 +36,29 @@ func NewImage(t *testing.T, namespace string) string {
...
+	image, err := c.AcornImageBuild(ctx, path+"/Acornfile", &client.AcornImageBuildOptions{
+		Cwd: path,
+	})
+	if err != nil {
+		return nil, err
+	}
+	t.Cleanup(func() {
+		_, err1 := c.ImageDelete(ctx, image.ID, &client.ImageDeleteOptions{Force: true})
+		_, err2 := c.AcornImageBuildDelete(ctx, image.ID)
+		if err1 != nil && err2 != nil {
```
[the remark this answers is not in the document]
jacobdonenfeld: Also unsure, figured if one worked, I'd be happy with it. Checking for errors, so there's only an issue if both are non-nil.

integration/helper/namespace.go
```diff
@@ -23,41 +23,6 @@ func TempNamespace(t *testing.T, client client.Client) *corev1.Namespace {
 				labels.AcornProject:       "true",
 			},
 		},
-		Spec:   corev1.NamespaceSpec{},
-		Status: corev1.NamespaceStatus{},
```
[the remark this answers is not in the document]
jacobdonenfeld: They did not need to be initialized. This is cleaning up unnecessary lines.

integration/helper/namespace.go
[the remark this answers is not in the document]
jacobdonenfeld: Great catch!

integration/client/apps/update_test.go
jacobdonenfeld: Doesn't seem to pass without it. Will need to look into this more.

pkg/cli/completion.go
jacobdonenfeld: If it doesn't find anything to complete, there isn't really an error, just no completions.

pkg/cli/images.go
jacobdonenfeld: Complicates clarity too much imo, would like to keep it as is.

pkg/cli/images_all_test.go
jacobdonenfeld: Done, fun logic :)

integration/client/client.go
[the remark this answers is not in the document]
g-linville: Yeah if they are both trying to do the same thing then I am good with this as it is.

pkg/cli/completion.go
tylerslaton: Ah, thanks for clearing this up for me. The way it currently is written makes more sense. Thank you!

.github/workflows/test.yaml
```diff
@@ -30,4 +30,4 @@ jobs:
       - run: docker buildx install
       - run: make setup-ci-image
       - run: ./bin/acorn install --image acorn:v-ci --skip-checks --acorn-dns=disabled
-      - run: TEST_ACORN_CONTROLLER=external make test
+      - run: export PATH=$PATH:$(pwd)/bin && TEST_ACORN_CONTROLLER=external make test
```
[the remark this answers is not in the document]
jacobdonenfeld: You're right. Refactored many tests to the client. Will also add cli specific tests.

jacobdonenfeld:
Refactored RESTClient to be an interface. Didn't end up using it, but seems like best practice to accept an interface, and allows us to do mocking in the future.

pkg/client/multi.go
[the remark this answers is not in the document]
jacobdonenfeld: Don't want to move it under factory, as I want it to remain a function that accepts the factory interface. This was carefully implemented to not break mocks. Renamed to getResourceAndClient

pkg/client/multi.go
```diff
@@ -40,6 +40,22 @@ func aggregate[T any, V ObjectPointer[T]](ctx context.Context, factory ProjectCl
...
+// parseProject accepts an argument name that is either of the form resource or project::resource, and ClientFactory
+// If the name is not of project::resource form, it returns defaultProject, name, nil
+// otherwise, it parses off the project name and returns project, name without project::, nil
+func parseProject(ctx context.Context, name string, factory ProjectClientFactory) (pc Client, parsedArg string, err error) {
+	var parsedProject string
+	if possibleProject, possibleArg, found := strings.Cut(name, "::"); found {
+		parsedProject = possibleProject
+		parsedArg = possibleArg
+	} else {
+		parsedProject = factory.DefaultProject()
```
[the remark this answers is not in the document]
jacobdonenfeld: That's a great point. Now that this function is returning the client, don't have to worry about it's changed default value.

pkg/cli/completion_test.go
jacobdonenfeld: I can't write any successful tests because of how this is implemented unfortunately. :(

pkg/cli/images.go
jacobdonenfeld: Not moved :(

jacobdonenfeld:
acorn project use acorn
acorn build ../acornfilepath
acorn images -a (to get image id)
acorn tag ID okay:test
acorn project create acorn2
acorn project use acorn2
➜ acorn run acorn::okay     
  ✗  ERROR:  GET https://index.docker.io/v2/library/okay/manifests/latest: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:library/okay Type:repository]]
➜  acorn run acorn::okay:test
  ✗  ERROR:  GET https://index.docker.io/v2/library/okay/manifests/test: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:library/okay Type:repository]]

*created image three:four on the current project, acorn2
➜ acorn run acorn2::three          
  ✗  ERROR:  GET https://index.docker.io/v2/library/three/manifests/latest: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:library/three Type:repository]]
➜ acorn run acorn2::three:four
**Works**

jacobdonenfeld:
acorn use project-y
acorn images -A
> project-x:image1:tag1   ID: xyz
> project-y:image2:tag2  ID: abc

acorn tag project-x::image1  newname:newtag
  ✗  ERROR:  images.api.acorn.io "image1" not found

acorn tag project-x::xyz newname:newtag
> Creates a new tag in project-x newname:newtag

acorn tag project-x::xyz project-y::newname:newtag
> ✗  ERROR:  could not parse reference:

jacobdonenfeld:
acorn pull cross project results in 
  ✗  ERROR:  could not parse reference: project-x::imagex

integration/client/apps/update_test.go
jacobdonenfeld: Fixed.

cjellick:
```
acorn projects use acorn
acorn build -t xyz:v1 ../simple-demo

acorn project use hub.stg.acornlabs.com/cjellick/kc-demo-us
acorn build -t xyz:v1 ../simple-demo

acorn projects use acorn
acorn run hub.stg.acornlabs.com/cjellick/kc-demo-us::xyz:v1
```
That last step will actually launch an app successfully using the image from the second step.

Similarly, this will just straight up launch the image i have in docker hub at `cjellick/simple-demo:v1
```
acorn run hub.stg.acornlabs.com/cjellick/kc-demo-us::cjellick/simple-demo:v1
```

These bugs fall into the realm of security issues where the user could potentially launch an image they were not expecting to launch.

cjellick:
Anything that "crosses" a project boundary should explicitly and clearly fail. So, the above scenarios I list as well as these:
```
# 1 try to use image from project X in project Y 
acorn project use project-y
acorn run project-x::my-image

# 2 try to tag across project boundaries
acorn projects use project-y
# 2A
acorn tag project-x::my-image my-image2
# 2B
acorn tag project-x::my-image project-y::my-image2

# 3 an image pull and push
acorn projects use project-y
acorn pull projext-x::my-image
acorn push my-image
```

need to just fail with explicit validation errors detecting cross project behavior.

jacobdonenfeld:
> I have to look into the implementation a bit more because it seems like you must be just discarding the project prefix in this case. Is that what's happening?

That is what was happening. It is now explicitly failing via a call to 
`args, err = noCrossProjectArgs(args, c.GetProject())`

Issue #1474, opened 2023-04-12
Autoupgrade pattern like v2.#.# updates to some untagged image


iwilltry42:
`acorn run --auto-upgrade 'ghcr.io/acorn-io/library/hello-world:v2.#.#'` should not start any app, because no tag matching this pattern exists:
![Screenshot 2023-04-12 at 07-30-30 library_hello-world versions · acorn-io](https://user-images.githubusercontent.com/25345277/231359689-423e1498-6204-48ea-8386-194c51072841.png)
Still, an app with the following data gets deployed:

```yaml
apiVersion: internal.acorn.io/v1
kind: AppInstance
metadata:
  creationTimestamp: "2023-04-12T05:24:02Z"
  generation: 1
  name: blue-rain
  namespace: acorn
  resourceVersion: "184496"
  uid: 35581950-a2fc-4fa5-ba50-7e45d2af101f
spec:
  autoUpgrade: true
  image: ghcr.io/acorn-io/library/hello-world:v2.#.#
status:
  appImage:
    acornfile: "containers: {\n\twebapp: {\n\t\timage: \"nginx:latest\"\n\t\tports:
      {\n\t\t\tpublish: \"80/http\"\n\t\t}\n\t\tscale: 1\n\t\tdirs: {\n\t\t\t\"/usr/share/nginx/html\":
      \"./html\"\n\t\t\t\"/etc/nginx/conf.d\": \"./nginx-conf\"\n\t\t}\n\t}\n}\n"
    digest: sha256:1a6c64d2ccd0bb035f9c8196d3bfe72a7fdbddc4530dfcb3ab2a0ab8afb57eeb
    id: 1a6c64d2ccd0bb035f9c8196d3bfe72a7fdbddc4530dfcb3ab2a0ab8afb57eeb
    imageData:
      containers:
        webapp:
          image: sha256:32d62e0f74960530b98c1f8479ecb16e6bd4ba9f221522aec838d0dd72009805
    name: ghcr.io/acorn-io/library/hello-world
    vcs: {}
  appSpec:
    ...
        image: sha256:32d62e0f74960530b98c1f8479ecb16e6bd4ba9f221522aec838d0dd72009805
...
  ready: true
```

[...]

Issue #1394, opened 2023-03-27
autoupgrade format "**" does not work as expected.


sangee2004:
acorn version v0.6.0-81-g5856fe34+5856fe34

Steps to reproduce the issue:
Have the following image tags:

```
acorn images
REPOSITORY                           TAG          IMAGE-ID
mytest2                              v0.0.1-rc1   8705f66f67e6
mytest2                              v0.0.2       fa11bfd2b180
```
Deploy app with "**" autoupgrade tag in image 

```
 acorn run -n mytestk3 --interval 15s "mytest2:v#.#.#**"
mytestk3
STATUS: ENDPOINTS[] HEALTHY[0] UPTODATE[0] [image-pull: waiting for image to satisfy auto-upgrade tag v#.#.#**]
STATUS: ENDPOINTS[] HEALTHY[0] UPTODATE[0] pending
STATUS: ENDPOINTS[] HEALTHY[0/1] UPTODATE[1] [containers: myweb1 ContainerCreating]
STATUS: ENDPOINTS[http://myweb1-mytestk3-11fcb466.local.on-acorn.io => myweb1:80] HEALTHY[0/1] UPTODATE[1] [containers: myweb1 ContainerCreating]
┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
| STATUS: ENDPOINTS[http://myweb1-mytestk3-11fcb466.local.on-acorn.io => myweb1:80] HEALTHY[1] UPTODATE[1] OK |
└─────────────────────────────────────────────────────────────────────────────────────────────────────────────┘

┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
| STATUS: ENDPOINTS[http://myweb1-mytestk3-11fcb466.local.on-acorn.io => myweb1:80] HEALTHY[1] UPTODATE[1] OK |
└─────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

App gets deployed with `mytest2:v0.0.1-rc1` image.

Expected Behavior:
App should get deployed with `mytest2:v0.0.2` image.

thedadams:
Never mind. I found a very easy fix for this.

sangee2004:
Tested with acorn version v0.6.0-150-g76e7d0a4+76e7d0a4

```
acorn images
mynew                                    v0.0.1-rc1   df447c241e1a
mynew                                    v0.0.2       ced7ab440ace
```

```
[...]

Issue #1377, opened 2023-03-24
autoupgrade formats dont work as expected when having "*"/ "**" in image.


sangee2004:
acorn version v0.6.0-55-g5d633ce7+5d633ce7

Steps to reproduce the problem:
Have 2 tags , "alpha" and "beta" created for an image.

Deploying app using any of the 2 autoupgrade formats results in app stuck in "waiting for image to satisfy auto-upgrade tag"
```
acorn run -n mytest4 "mytest1:*"
acorn run -n mytest4 "mytest1:**"
```

```
 % acorn images                  
REPOSITORY   TAG       IMAGE-ID
mytest1      alpha     f66612d09d2f
mytest1      beta      f66612d09d2f
 % acorn run -n mytest4 "mytest1:*"
mytest4
STATUS: ENDPOINTS[] HEALTHY[] UPTODATE[] 
STATUS: ENDPOINTS[] HEALTHY[0] UPTODATE[0] [image-pull: waiting for image to satisfy auto-upgrade tag *]
^C  ✗  ERROR:  terminating watch: context canceled


% acorn run -n mytest5 "mytest1:**"
mytest5
STATUS: ENDPOINTS[] HEALTHY[] UPTODATE[] 
STATUS: ENDPOINTS[] HEALTHY[0] UPTODATE[0] [image-pull: waiting for image to satisfy auto-upgrade tag **]
^C  ✗  ERROR:  terminating watch: context canceled

```

sangee2004:
Tested with acorn version `v0.6.0-150-g76e7d0a4+76e7d0a4`
``
 acorn images
REPOSITORY                               TAG       IMAGE-ID
mytest1                                  alpha     35e4e925ff1d
mytest1                                  beta      14fbf57f4832
``

acorn run -n mytest5 "mytest1:*"
acorn run -n mytest6 "mytest1:**"

```
acorn apps
NAME            IMAGE                                            HEALTHY   UP-TO-DATE   CREATED     ENDPOINTS                                                             MESSAGE
mytest6         mytest1:alpha                                    1         1            8m55s ago   http://myweb1-mytest6-f36fa354.local.on-acorn.io => myweb1:80         OK
mytest5         mytest1:beta                                     1         1            9m4s ago    http://myweb1-mytest5-d499ad26.local.on-acorn.io => myweb1:80         OK

```

[...]

PR #1480, opened 2023-04-12
Handle edge-cases around auto-upgrade patterns (#1377)

thedadams:
There were a few cases where auto-upgrade patterns failed to do the expected thing.

1. When using an auto-upgrade pattern that did not match any existing tags, the digest for existing tags would be pulled and used to start the app. This is incorrect when no tag matched the pattern.
2. When using the auto-upgrade pattern of only '*', the matching would fail to match any tag that was not "latest".
3. Using an auto-upgrade pattern ending with '**' implicitly expected that the '**' would match something. So, if the pattern was 'v1.#**' and the existing tags were 'v1.1-rc1' and 'v1.2', then the first tag would match even though the second tag was "better" according to the '#'.

These issues are fixed here.

### Checklist

Issues:
https://github.com/acorn-io/acorn/issues/1394
https://github.com/acorn-io/acorn/issues/1377
https://github.com/acorn-io/acorn/issues/1474

### DIFF
--- a/pkg/autoupgrade/daemon.go
+++ b/pkg/autoupgrade/daemon.go
@@ -178,7 +178,7 @@ func (d *daemon) determineAppsToRefresh(apps map[kclient.ObjectKey]v1.AppInstanc
 // which will trigger the appInstance handlers to pick up the change and deploy the new version of the app
 func (d *daemon) refreshImages(ctx context.Context, apps map[kclient.ObjectKey]v1.AppInstance, imagesToRefresh map[imageAndNamespaceKey][]kclient.ObjectKey, updateTime time.Time) {
 	for imageKey, appsForImage := range imagesToRefresh {
-		current, err := imagename.ParseReference(imageKey.image, imagename.WithDefaultRegistry(defaultNoReg))
+		current, err := imagename.ParseReference(imageKey.image, imagename.WithDefaultRegistry(defaultNoReg), imagename.WithDefaultTag(""))
 		if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/autoupgrade/daemon.go
```diff
@@ -206,9 +206,11 @@ func (d *daemon) refreshImages(ctx context.Context, apps map[kclient.ObjectKey]v
 			// Updated can be false for two reasons:
 			// 1. The tag was a pattern and a newer tag was not found
 			// 2. The tag was not a pattern
-			// In either case, we also want to check to see if new content was pushed to the current tag
+			// In either case, we also want to check to see if new content was pushed to the current tag, if the image has a current tag.
 			// This satisfies the usecase of autoUpgrade with an app's tag is something static, like "latest"
-			if !updated {
+			// However, if the tag is a pattern and the current image has no tag, we don't want to check for a digest because this would
+			// result in a digest upgrade even though no tag matched.
+			if !updated && (!isPattern || current.Identifier() != "") {
```
cjellick, line 213: If I have a pattern, say `v#.#` and the app is currently on `v1.0` and I push a new version of `v1.0`, will the app get updated?

thedadams: Yes, because `current.Identifier() != ""` in that case.

cjellick (approved):
looks good

cjellick:
i think this is fine to merge with just 1 approval, since @thedadams and @yashgiri paired on it

njhale (approved):
lgtm

Issue #1479, opened 2023-04-12
runtime error  when trying to list projects.


sangee2004:
acorn --version
acorn version v0.6.0-145-g1f6d90fd+1f6d90fd

Steps to reproduce the problem:

Tried to list projects as soon as I logged in to remote server.

```
acorn login -p xxx <remote server>
acorn project use <remote server>/<account>/mytestsandbox

  •  Run "acorn projects xxx" to list available projects
  •  Run "acorn project use xxx/<account>/mytestsandbox" to set default project
  ✔  Login to xxxx as <account> succeeded

 % acorn project
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x148 pc=0x32823f9]

goroutine 211 [running]:
github.com/acorn-io/acorn/pkg/project.GetDetails.func1({0xc0005a6de0, 0x28}, {{0xc0005a6de0, 0x28}, {0x0, 0x0}, {0x0, 0x0}, 0x0, 0x0, ...})
	/home/ubuntu/actions-runner/_work/acorn/acorn/pkg/project/operations.go:205 +0x119
created by github.com/acorn-io/acorn/pkg/project.GetDetails
	/home/ubuntu/actions-runner/_work/acorn/acorn/pkg/project/operations.go:199 +0x88

```

On retry, worked fine.

sangee2004:
I hit a similar problem again.


```
acorn projects
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x148 pc=0x3250fb9]

goroutine 182 [running]:
github.com/acorn-io/acorn/pkg/project.GetDetails.func1({0xc0002e7bf0, 0x2b}, {{0xc0002e7bf0, 0x2b}, {0x0, 0x0}, {0x0, 0x0}, 0x0, 0x0, ...})
	/home/runner/actions-runner/_work/acorn/acorn/pkg/project/operations.go:205 +0x119
created by github.com/acorn-io/acorn/pkg/project.GetDetails
	/home/runner/actions-runner/_work/acorn/acorn/pkg/project/operations.go:199 +0x88
sangeethahariharan@Sangeethas-MBP example1 % acorn projects
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x148 pc=0x3250fb9]

goroutine 182 [running]:
[...]

PR #1506, opened 2023-04-15
refactor go routine for project region output (#1479)

jsilverio22:
### Checklist
#1479

### DIFF
--- a/pkg/cli/project.go
+++ b/pkg/cli/project.go
@@ -84,26 +84,31 @@ func (a *Project) Run(cmd *cobra.Command, args []string) error {
 	}
 
 	out := table.NewWriter(tables.ProjectClient, a.Quiet, a.Output)
-	projectDetails, err := project.GetDetails(cmd.Context(), a.client.Options(), projectNames)
-	if err != nil {
-		return err
-	}
+	projectDetails := project.GetDetails(cmd.Context(), a.client.Options(), projectNames)
+
 	sort.Slice(projectDetails, func(i, j int) bool {
 		return projectDetails[i].FullName < projectDetails[j].FullName
 	})
+
 	for _, projectItem := range projectDetails {
-		supportedRegions := projectItem.Project.GetSupportedRegions()
-		defaultRegion := projectItem.Project.GetRegion()
-		for i, supportedRegion := range supportedRegions {
-			if supportedRegion == defaultRegion {
-				supportedRegions[i] = supportedRegion + "*"
-			}
+		if projectItem.Err != nil {
+			logrus.Warnf("Could not list details of project [%s]: %v", projectItem.FullName, projectItem.Err)
+			continue
 		}
-		out.Write(projectEntry{
-			Name:    projectItem.FullName,
-			Default: defaultProject == projectItem.FullName,
-			Regions: supportedRegions,
-		})
+		if projectItem.Project != nil {
+			supportedRegions := projectItem.Project.GetSupportedRegions()
+			defaultRegion := projectItem.Project.GetRegion()
+			for i, supportedRegion := range supportedRegions {
+				if supportedRegion == defaultRegion {
+					supportedRegions[i] = supportedRegion + "*"
+				}
+			}
+			out.Write(projectEntry{
+				Name:    projectItem.FullName,
+				Default: defaultProject == projectItem.FullName,
+				Regions: supportedRegions,
+			})
+		}
 	}
 
 	return out.Err()
--- a/pkg/project/operations.go
+++ b/pkg/project/operations.go
@@ -117,6 +117,7 @@ type listResult struct {
 type DetailProject struct {
 	FullName string
 	Project  *apiv1.Project
+	Err      error
 }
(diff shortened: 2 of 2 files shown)


pkg/project/operations.go
```diff
@@ -190,37 +191,50 @@ func List(ctx context.Context, opts Options) (projects []string, warnings map[st
 	return projects, warnings, nil
 }
 
-func GetDetails(ctx context.Context, opts Options, projectNames []string) (result []DetailProject, err error) {
-	entryCh := make(chan DetailProject)
-	errCh := make(chan error)
+func GetDetails(ctx context.Context, opts Options, projectNames []string) (projects []DetailProject, warnings map[string]error, err error) {
```
ibuildthecloud, line 194: Why have the warnings return when you are just coping the `err` field to the map?  Can you just make the `err` public `Err` and the client read that? I'm not sure what the extra data structure and logic buys you.

jsilverio22: that makes sense, removed the warnings map

Is the purpose of the warnings return in the List function that there are 2 sources they can be coming from local or hub servers and it serves to differentiate where they are coming from?

pkg/project/operations.go
```diff
@@ -190,37 +191,45 @@ func List(ctx context.Context, opts Options) (projects []string, warnings map[st
...
-			opts.Project = proj
+		wg.Add(1)
+		go func(projectName string, opts Options) {
+			defer wg.Done()
+			// Launch a goroutine for each project to retrieve its information
+			opts.Project = projectName
 			c, err := Client(ctx, opts)
 			if err != nil {
-				errCh <- err
+				logrus.Debugf("unable to get client for %s: %s", projectName, err)
```
njhale, line 223: nit: should this be a warn level log? or possibly let this bubble up in a `DetailProject.Err` with a `nil` `Project` field?

pkg/cli/project.go
```diff
@@ -88,6 +88,13 @@ func (a *Project) Run(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+
+	for _, project := range projectDetails {
+		if project.Err != nil {
+			logrus.Warnf("Could not list projects from [%s]: %v", project.FullName, project.Err)
```
njhale, line 94:
```suggestion
			logrus.Warnf("Could not list details of project [%s]: %v", project.FullName, project.Err)
```

pkg/cli/project.go
```diff
@@ -88,6 +88,13 @@ func (a *Project) Run(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+
+	for _, project := range projectDetails {
```
njhale, line 92: nit: I would do this after sorting so we get deterministic-ish output.

pkg/cli/project.go
```diff
@@ -88,6 +88,13 @@ func (a *Project) Run(cmd *cobra.Command, args []string) error {
...
+		if project.Err != nil {
```
njhale, line 93: If we failed to get project details, is it possible that `project.Project == nil`? If so, it looks like the for-loop below can still hit a nil pointer:

```go
for _, projectItem := range projectDetails {
		supportedRegions := projectItem.Project.GetSupportedRegions()
		defaultRegion := projectItem.Project.GetRegion()
		for i, supportedRegion := range supportedRegions {
			if supportedRegion == defaultRegion {
				supportedRegions[i] = supportedRegion + "*"
			}
		}
		out.Write(projectEntry{
			Name:    projectItem.FullName,
			Default: defaultProject == projectItem.FullName,
			Regions: supportedRegions,
		})
}
```

maybe we should add some checks before trying to access it there?

njhale (changes requested):
Nice! Thanks for putting this together

Just a few small questions/comments before I approve.

pkg/project/operations.go
jsilverio22: yeah I was trying to follow the pattern of the List func where it just debug logs when local kubeconfig client is invalid but I think it makes more sense to warn here instead in this case.

pkg/cli/project.go
jsilverio22: good point, I consolidated the err check and now check for nil projectItem in this loop

pkg/project/operations.go
```diff
@@ -190,37 +191,45 @@ func List(ctx context.Context, opts Options) (projects []string, warnings map[st
 	return projects, warnings, nil
 }
 
-func GetDetails(ctx context.Context, opts Options, projectNames []string) (result []DetailProject, err error) {
-	entryCh := make(chan DetailProject)
-	errCh := make(chan error)
+func GetDetails(ctx context.Context, opts Options, projectNames []string) (projects []DetailProject, err error) {
```
njhale, line 194: Ah, forgot to post this comment:

This function always returns a `nil` error, I think it's safe to drop the `error` return param.

njhale (approved):
LGTM! Thanks for the quick turnaround!

PR #1518, opened 2023-04-17
Use Go's built-in DNS resolver to avoid DNS issues when building Acorns

g-linville:
### Checklist

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -21,6 +21,8 @@ builds:
         goarch: arm64
       - goos: windows
         goarch: arm
+    flags:
+      - -tags netgo
     ldflags:
       - -s
       - -w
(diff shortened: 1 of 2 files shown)


ibuildthecloud:
I don't think this makes any difference. AFAIK we are already using the pure go resolver because the binary is not linked to glibc at all (at least on linux).  I could be wrong, but if you get 
```
$ ldd bin/acorn 
	not a dynamic executable
```
I don't see how it could not be using pure go.

cjellick:
oh wow. this 100% fixes it for me locally

cjellick:
i didnt think you had to mess with `-tags netgo` anymore

g-linville:
For me, on macOS, I never hit the issue with `-tags netgo`. I sometimes hit it without that. I agree that it's strange...

cjellick (approved):

thedadams (approved):

Issue #1298, opened 2023-03-01
acron update done after changes are done to computeclass does not get applied (unless there are other changes done to the app)


sangee2004:
acorn version v0.6.0-rc1-6-gb57d7f88+b57d7f88

Steps to reproduce the problem:
1. Create a computeclass (say c1)
2. Deploy an app (test1) using this computeclass c1 (acorn run --update -n test1 --compute-class c1 . )
3. In my case , the app did not get scheduled to any node because of incorrect scheduling rules .
4. Make the required changes to the scheduling rules in the computeclass c1.
5. Update app created in step2  using the same compute class (acorn run --update -n test1 --compute-class c1 . )

This does not result in the changes in compute class to be applied to the apps.
Containers dont get restarted as part of this update.

Note -- If there is any change made to the app ( in changed some args in Acornfile) , then update picks up the changes made to th compute class and containers get redeployed successfully.

Expected Behavior:
Changes done  to computeclass should get picked up on an `acorn update` .

tylerslaton:
Talked briefly with @thedadams offline and the solution here is to likely add a `--force` flag to `acorn update` which would update the `generation` flag even if nothing changes. Then trying to get this desired behavior becomes running `acorn update --force`.

cjellick:
@tylerslaton did you confirm this works? If so, this issue can be closed

tylerslaton:
@cjellick the `--force` flag does not currently exist. It is something we would need to add.

PR #1533, opened 2023-04-21
add generation to spec for defaults (#1298)

jsilverio22:
### Checklist

#1298


pkg/controller/appdefinition/status.go
```diff
@@ -588,3 +588,25 @@ func endpoints(req router.Request, cfg *apiv1.Config, app *v1.AppInstance) (stri
...
+func RefreshStatus(h router.Handler) router.Handler {
+	return router.HandlerFunc(func(req router.Request, resp router.Response) error {
+		appInstance := req.Object.(*v1.AppInstance)
+
+		if appInstance.Annotations["acorn.io/refresh"] == "true" {
+			delete(appInstance.Annotations, "acorn.io/refresh")
+			for i, annotation := range appInstance.Spec.Annotations {
+				if annotation.Key == "acorn.io/refresh" {
+					appInstance.Spec.Annotations = append(appInstance.Spec.Annotations[:i], appInstance.Spec.Annotations[i+1:]...)
+				}
```
jacobdonenfeld, line 601: Could this be deleted the same way as `delete(appInstance.Annotations, "acorn.io/refresh")`

pkg/controller/appdefinition/testdata/status/refresh/expected.yaml
```diff
@@ -0,0 +1,221 @@
...
+apiVersion: v1
+metadata:
+  name: buildimage-pull-1234567890ab
+  namespace: app-created-namespace
+  labels:
+    acorn.io/managed: "true"
+    acorn.io/pull-secret: "true"
+type: "kubernetes.io/dockerconfigjson"
+data:
+  ".dockerconfigjson": eyJhdXRocyI6eyJpbmRleC5kb2NrZXIuaW8iOnsiYXV0aCI6Ik9nPT0ifX19
```
jacobdonenfeld, line 209: Is this some random key, or an actual key?

pkg/controller/appdefinition/status.go
```diff
@@ -588,3 +588,25 @@ func endpoints(req router.Request, cfg *apiv1.Config, app *v1.AppInstance) (stri
...
+
+func RefreshStatus(h router.Handler) router.Handler {
+	return router.HandlerFunc(func(req router.Request, resp router.Response) error {
+		appInstance := req.Object.(*v1.AppInstance)
+
+		if appInstance.Annotations["acorn.io/refresh"] == "true" {
+			delete(appInstance.Annotations, "acorn.io/refresh")
+			for i, annotation := range appInstance.Spec.Annotations {
+				if annotation.Key == "acorn.io/refresh" {
+					appInstance.Spec.Annotations = append(appInstance.Spec.Annotations[:i], appInstance.Spec.Annotations[i+1:]...)
```
jacobdonenfeld, line 600: This likely introduces a weird behavior if appInstance.Spec.Annotations happens to have acorn.io/refresh=true two or more times. I think it would never be able to get rid of its label, and keeps almost doubling the amount of labels.

pkg/controller/appdefinition/status.go
jsilverio22: delete is only for maps I think and the appInstance.Spec.Annotations is a slice of ScopedLabels

pkg/controller/appdefinition/status.go
jsilverio22: good catch, I modified to start from then end and remove which I think resolves this

pkg/controller/appdefinition/testdata/status/refresh/expected.yaml
jsilverio22: it's reused from another test testdata/deployspec/basic just a test one

pkg/controller/appdefinition/status.go
jsilverio22: added a test case for this as well

pkg/cli/run.go
```diff
@@ -314,6 +315,13 @@ func (s *Run) update(ctx context.Context, c client.Client, imageSource imagesour
 	updateOpts.Replace = s.Replace
 	updateOpts.Image = image
 	updateOpts.DeployArgs = deployArgs
+	if s.Refresh {
+		refreshAnnotation, err := v1.ParseScopedLabels("acorn.io/refresh=true")
+		if err != nil {
+			return nil, false, err
+		}
+		updateOpts.Annotations = append(updateOpts.Annotations, refreshAnnotation...)
```
thedadams, line 323: I would like to avoid having to add this annotation to the app spec. Can we do something here to only put the annotation on the `AppInstance` object?

The reason being that this actually changes the app (which bumps the generation).

pkg/controller/appdefinition/status.go
```diff
@@ -588,3 +588,25 @@ func endpoints(req router.Request, cfg *apiv1.Config, app *v1.AppInstance) (stri
...
+
+		if appInstance.Annotations["acorn.io/refresh"] == "true" {
+			delete(appInstance.Annotations, "acorn.io/refresh")
+			for i := len(appInstance.Spec.Annotations) - 1; i >= 0; i-- {
+				if appInstance.Spec.Annotations[i].Key == "acorn.io/refresh" {
+					appInstance.Spec.Annotations = append(appInstance.Spec.Annotations[:i], appInstance.Spec.Annotations[i+1:]...)
+				}
+			}
+
+			err := req.Client.Update(req.Ctx, req.Object)
```
thedadams, line 604: We don't need to update this here; `baaah` will take care of this for us.

pkg/controller/routes.go
```diff
@@ -47,7 +47,7 @@ func routes(router *router.Router, registryTransport http.RoundTripper) {
 	router.HandleFunc(&v1.AppInstance{}, images.CreateImages)
 	router.HandleFunc(&v1.AppInstance{}, appdefinition.ParseAppImage)
 	router.HandleFunc(&v1.AppInstance{}, tls.ProvisionCerts) // Provision TLS certificates for port bindings with user-defined (valid) domains
-	router.Type(&v1.AppInstance{}).Middleware(appdefinition.FilterLabelsAndAnnotationsConfig).HandlerFunc(namespace.AddNamespace)
+	router.Type(&v1.AppInstance{}).Middleware(appdefinition.FilterLabelsAndAnnotationsConfig, appdefinition.RefreshStatus).HandlerFunc(namespace.AddNamespace)
```
thedadams, line 50: This middleware should be added on line 56 instead. By adding it here, before we calculate the defaults, then the defaults won't get recalculated if the generation doesn't change. The reason this is working right now is that you are actually changing the generation.

pkg/controller/appdefinition/status_test.go
```diff
@@ -36,3 +36,11 @@ func TestCheckStatus(t *testing.T) {
...
 	assert.True(t, called, "router handler call expected")
 }
+
+func TestRefreshStatus(t *testing.T) {
+	tester.DefaultTest(t, scheme.Scheme, "testdata/status/refresh", RefreshStatus(router.HandlerFunc(DeploySpec)).Handle)
+}
+
+func TestRefreshStatusMulti(t *testing.T) {
+	tester.DefaultTest(t, scheme.Scheme, "testdata/status/refresh-multi", RefreshStatus(router.HandlerFunc(DeploySpec)).Handle)
+}
```
thedadams, line 46: These test ensure that the annotation you added is getting removed. However, we also want a test (or two) that ensures that the defaults are recalculated when we want them to be.

thedadams (changes requested):

cjellick:
I don't like this approach. I know this is a bit of a corner case, but the solution is very antithetical to how we present acorn

> 1. Run `kubectl edit appinstance <appinstancename> -n <namespace>` to open the configuration file for the app that needs to be updated.


We don't even want the user to know "AppInstance" is a thing. In fact, a normal developer should not have access to appInstance. Can you set generation on App rather than appInstance?

What if we just had an explicit `--generation` flag with a description like "Setting this to a new value will cause the app to be re-evaluated and potentially redeployed.

jsilverio22:
closing until issue is confirmed by user

PR #1539, opened 2023-04-24
Move region to spec of container replicas and volumes

thedadams:
The region was put on the status of container replicas and volumes because the user doesn't directly set the region on these objects. However, the user doesn't directly set anything on these objects; they are generated from the app definition. Therefore, putting the region on the spec makes sense and simplifies many things.


jsilverio22 (approved):
just a knowledge confirmation other than that lgtm

nothing will change from the retrieval of regions perspective other than it's on spec vs status? (like for acorn project update)

thedadams:
@jsilverio22 Correct, with the caveat that projects aren't involved here. A user isn't allowed to create these objects (container replica and volume) directly.

ibuildthecloud:
I'll buy it.  Sounds good.

PR #1554, opened 2023-04-26
Create NetworkPolicies for builders

g-linville:
This PR adds a new handler that creates a NetworkPolicy for each builder. This is what the NetworkPolicy does:

- Allow traffic **only** from the acorn-apiserver, which is responsible for proxying traffic to the builders.
- Allow traffic **only** to the following locations:
  - The Acorn internal image registry
  - CoreDNS
  - Any IP address outside of the cluster (the Internet)

I also renamed the other NetworkPolicy handlers to make the names shorter/less redundant.

### Checklist

### DIFF
--- a/pkg/controller/networkpolicy/networkpolicy.go
+++ b/pkg/controller/networkpolicy/networkpolicy.go
@@ -325,10 +325,8 @@ func PolicyForService(req router.Request, resp router.Response) error {
 // It also only allows outgoing traffic to CoreDNS, the Acorn registry, and the Internet.
 func PolicyForBuilder(req router.Request, resp router.Response) error {
 	cfg, err := config.Get(req.Ctx, req.Client)
-	if err != nil {
+	if err != nil || !*cfg.NetworkPolicies {
 		return err
-	} else if !*cfg.NetworkPolicies {
-		return nil
 	}
 
 	deployment := req.Object.(*appsv1.Deployment)
--- a/pkg/controller/networkpolicy/networkpolicy.go
+++ b/pkg/controller/networkpolicy/networkpolicy.go
@@ -27,9 +27,9 @@ const (
 	awsEC2cidrV6 = "fd00:ec2::254/64"
 )
 
-// PolicyForApp creates a single Kubernetes NetworkPolicy that restricts incoming network traffic
+// ForApp creates a single Kubernetes NetworkPolicy that restricts incoming network traffic
 // to all pods in an app, so that they cannot be reached by pods from other projects.
-func PolicyForApp(req router.Request, resp router.Response) error {
+func ForApp(req router.Request, resp router.Response) error {
 	cfg, err := config.Get(req.Ctx, req.Client)
 	if err != nil {
 		return err
@@ -98,10 +98,10 @@ func PolicyForApp(req router.Request, resp router.Response) error {
 	return nil
 }
 
-// PolicyForIngress creates Kubernetes NetworkPolicies to allow traffic to exposed HTTP ports on
+// ForIngress creates Kubernetes NetworkPolicies to allow traffic to exposed HTTP ports on
 // Acorn apps from the ingress controller. If the ingress controller namespace is not defined, traffic from
 // all namespaces will be allowed instead.
-func PolicyForIngress(req router.Request, resp router.Response) error {
+func ForIngress(req router.Request, resp router.Response) error {
 	cfg, err := config.Get(req.Ctx, req.Client)
 	if err != nil {
 		return err
@@ -236,9 +236,9 @@ func PolicyForIngress(req router.Request, resp router.Response) error {
 	return nil
 }
 
-// PolicyForService creates a Kubernetes NetworkPolicy to allow traffic to published TCP/UDP ports
+// ForService creates a Kubernetes NetworkPolicy to allow traffic to published TCP/UDP ports
 // on Acorn apps that are exposed with LoadBalancer Services.
-func PolicyForService(req router.Request, resp router.Response) error {
+func ForService(req router.Request, resp router.Response) error {
 	cfg, err := config.Get(req.Ctx, req.Client)
 	if err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


pkg/controller/networkpolicy/networkpolicy.go
```diff
@@ -8,19 +8,21 @@ import (
...
 	apierror "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/strings/slices"
 )
 
-// NetworkPolicyForApp creates a single Kubernetes NetworkPolicy that restricts incoming network traffic
+// PolicyForApp creates a single Kubernetes NetworkPolicy that restricts incoming network traffic
 // to all pods in an app, so that they cannot be reached by pods from other projects.
-func NetworkPolicyForApp(req router.Request, resp router.Response) error {
+func PolicyForApp(req router.Request, resp router.Response) error {
```
thedadams, line 25: nit: if we're changing the name, let's trim it all the way down:
```suggestion
func ForApp(req router.Request, resp router.Response) error {
```

pkg/controller/networkpolicy/networkpolicy.go
```diff
@@ -305,3 +296,111 @@ func NetworkPolicyForService(req router.Request, resp router.Response) error {
...
+
+// PolicyForBuilder creates a Kubernetes NetworkPolicy to allow traffic to the buildkitd pods from the acorn-api only.
+// It also only allows outgoing traffic to CoreDNS, the Acorn registry, and the Internet.
+func PolicyForBuilder(req router.Request, resp router.Response) error {
+	cfg, err := config.Get(req.Ctx, req.Client)
+	if err != nil {
+		return err
+	} else if !*cfg.NetworkPolicies {
+		return nil
+	}
```
thedadams, line 308: nit:
```suggestion
	if err != nil || !*cfg.NetworkPolicies {
		return err
	}
```

thedadams (approved):

pkg/controller/networkpolicy/networkpolicy.go
```diff
@@ -305,3 +296,111 @@ func NetworkPolicyForService(req router.Request, resp router.Response) error {
...
+		CIDR: "0.0.0.0/0",
+	}
+
+	// get pod CIDRs from the nodes so that we can only allow IP addresses outside the cluster
+	nodes := corev1.NodeList{}
+	if err := req.Client.List(req.Ctx, &nodes); err != nil {
+		return nil, fmt.Errorf("failed to list nodes: %w", err)
+	}
+	for _, node := range nodes.Items {
+		for _, cidr := range node.Spec.PodCIDRs {
```
StrongMonkey, line 398: Should it block service IP cidr too?

pkg/controller/networkpolicy/networkpolicy.go
```diff
@@ -305,3 +296,111 @@ func NetworkPolicyForService(req router.Request, resp router.Response) error {
...
+			Namespace: deployment.Namespace,
+			Labels: map[string]string{
+				labels.AcornManaged: "true",
+			},
+		},
+		Spec: networkingv1.NetworkPolicySpec{
+			PodSelector: metav1.LabelSelector{
+				MatchLabels: podLabels,
+			},
+			Ingress: []networkingv1.NetworkPolicyIngressRule{{
```
StrongMonkey, line 335: Another usecase for builder is that we can publish builder using ingress and the build will just using ingress traffic. Hub workload cluster uses this mode. Have we accounted for allowing ingress traffic in this case?

g-linville: I was looking at our dev sandbox workload cluster and did not see any ingress resources exposing the builders. Do you know of a cluster where we are doing this so that I can look at it for reference?

StrongMonkey: Sandbox workload cluster should have it. https://github.com/acorn-io/hub-system-charts/blob/main/templates/acorn-install.yaml#L53-L55

You can also try installing acorn using this flag --publishBuilders

pkg/controller/networkpolicy/networkpolicy.go
g-linville: I definitely think we should do this, but the trouble is that I can't find a way to get this information programatically. Apparently it is sometimes in the output of `kubectl cluster-info dump` but it did not appear there for me. We could make this another installation parameter that the user can optionally provide if you think that would be good.

StrongMonkey: yeah it is true. What was the original intention here? To block buildkit pod being accessed from other buildkit?

g-linville: This is what @cjellick wanted originally:

"Builders all go in the same namespace. They aren't isolated form each other. Should do a policy specifically for builders. Isolate them from each other as though they're in their own projects. Should only be able to talk to registry and internet"

Based on that last sentence I decided to restrict egress traffic. I think this makes sense to do, since users can run whatever commands they want to run in their Dockerfiles, which are also (potentially) built during the Acorn build process.

I wish I could set an explicit deny to all traffic in all namespaces by default, and then use the ipBlock to let it get out to the Internet. But I'm not sure if there is a way to do that.

StrongMonkey: Got it. Just curious, with your change, can you still access other buildkit service using service IP? I wonder when kube-proxy translate service IP into pod ip, the network policy will still block the pod ip, then we don't have to worry about it.

g-linville: Just tested this in a local cluster. Thankfully, just using the pod IP CIDR works just fine. I deployed a pod in the default namespace and exposed with with a service. Without this NetPol, I was able to curl that pod just fine, by both its service hostname (`nginx.default.svc.cluster.local`) and by the Service IP that that hostname resolves to. When I added the NetworkPolicy to restrict egress traffic, I could no longer reach that service, by its hostname or its IP. So I think we are good here. 👍

pkg/controller/networkpolicy/networkpolicy.go
g-linville: I tested this out. Thankfully, the `ForIngress` handler steps in and takes care of what we need to ensure that the ingress controller can proxy traffic to the builder without any issues.

njhale (approved):

PR #1566, opened 2023-04-28
Add quota management system

tylerslaton:
This PR adds a process that enforces quota management for an installation of Acorn. In its current implementation, this only applies for `appInstances`. This process is enabled by adding the  `acorn.io/quota-enforced: true` annotation to a project.  Enforcement then comes from two new handlers:

1. EnsureQuotaRequest
2. WaitForAllocation

These two handlers manage a new `admin` type - `QuotaRequest`. This new type is responsible for specifying what resources are being requested. It then halts the provisioning of that resource until a third party updates the status of the `QuotaRequest` to acknowledge allocation of resources.

> **Note**: On its own, `QuotaRequests` will never be satisfied until something outside of the Acorn system updates the `QuotaRequest` to allow the appInstance to proceed.

### Checklist


pkg/controller/quota/quota.go
```diff
@@ -0,0 +1,128 @@
...
+	app := appInstance.Status.AppSpec
+
+	// Add the volume storage needed to the quota request.
+	for _, volume := range app.Volumes {
+		size, err := resource.ParseQuantity(string(volume.Size))
+		if err != nil {
+			return err
+		}
+		quotaRequest.Spec.VolumeStorage.Add(size)
+	}
```
tylerslaton, line 126: Since this is grabbing `Volumes` from the AppSpec, this will only calculate storage for ones that are specified in the Acornfile. This is an issue as it means that `Volumes` specified client side in the `AppInstance` can effectively get around being tracked in the quota. I would just process the `AppInstance` `Volumes` directly but that implements a second source of truth for the state of `Volumes`.

tylerslaton: Actually this isn't a problem as the `Volumes` defined in the `AppInstance` only overwrites existing `Volumes` in the `Acornfile`.

pkg/controller/quota/quota.go
```diff
@@ -0,0 +1,135 @@
...
+	return nil
+}
+
+// addCompute adds the compute resources of the containers passed to the quota request.
+func addCompute(containers map[string]apiv1.Container, appInstance *apiv1.AppInstance, quotaRequest *adminv1.QuotaRequestInstance) {
+	// For each workload, add their memory/cpu requests to the quota request
+	for name, container := range containers {
+		if _, ok := appInstance.Status.Scheduling[name]; ok {
+			quotaRequest.Spec.Resources.CPU.Add(appInstance.Status.Scheduling[name].Requirements.Requests["cpu"])
+			quotaRequest.Spec.Resources.Memory.Add(appInstance.Status.Scheduling[name].Requirements.Requests["memory"])
```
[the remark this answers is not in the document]
tylerslaton: I like it! Reduces the amount of text by a lot.

pkg/controller/quota/quota.go
```diff
@@ -0,0 +1,135 @@
...
+func addStorage(appInstance *apiv1.AppInstance, quotaRequest *adminv1.QuotaRequestInstance) error {
+	app := appInstance.Status.AppSpec
+
+	// Add the volume storage needed to the quota request. We only parse net new volumes, not
+	// existing ones that are then bound client-side.
+	for name, volume := range app.Volumes {
+		size := volume.Size
+		if bound, boundSize := boundSize(name, appInstance.Spec.Volumes); bound {
+			size = boundSize
+		}
```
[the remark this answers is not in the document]
tylerslaton: That's a great question. What I was trying to solve her is the situation where a volume is defined in the Acornfile and a user overwrites it with a new size for it. Any ideas about how to address that while not counting twice for already created volumes?

pkg/controller/quota/quota.go
```diff
@@ -0,0 +1,135 @@
...
+	// Determine how to proceed depending on if the quotaRequest exists and what it has written to its status
+	if quotaRequest.Status.Condition(adminv1.QuotaRequestCondition).Error {
+		status.Error(fmt.Errorf("%v", quotaRequest.Status.Condition(adminv1.QuotaRequestCondition).Message))
+	} else if quotaRequest.Generation != quotaRequest.Status.ObservedGeneration {
+		status.Unknown("waiting for quota allocation")
+	} else if quotaRequest.Status.Condition(adminv1.QuotaRequestCondition).Success {
+		status.Success()
+	}
+
+	return req.Client.Status().Update(req.Ctx, appInstance)
```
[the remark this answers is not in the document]
tylerslaton: This was originally an AppInstance handler and I moved away from it because I was noticing some weird behavior in the integration test where the QuotaRequest status would be updated but this handler would never run since (I think) the AppInstance wasn't actually changed

pkg/apis/api.acorn.io/v1/types.go
```diff
@@ -402,6 +402,7 @@ type Config struct {
 	AllowTrafficFromNamespace      []string `json:"allowTrafficFromNamespace" name:"allow-traffic-from-namespace" usage:"Namespaces that are allowed to send network traffic to all Acorn apps"`
 	ServiceLBAnnotations           []string `json:"serviceLBAnnotations" name:"service-lb-annotation" usage:"Annotation to add to the service of type LoadBalancer. Defaults to empty. (example key=value)"`
 	AWSIdentityProviderARN         *string  `json:"awsIdentityProviderArn" name:"aws-identity-provider-arn" usage:"ARN of cluster's OpenID Connect provider registered in AWS"`
+	QuotaEnforced                  *bool    `json:"quotaEnforced" name:"quota-enforced" usage:"Enable quota tracking for the installation. (default 'false')"`
```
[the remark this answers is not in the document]
tylerslaton: Yeah, I like EnforcedQuota more here too, good callout.

pkg/controller/quota/quota.go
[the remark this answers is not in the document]
thedadams: If the `QuotaRequest` is created by a handler on the `AppInstance`, then changes to the `QuotaRequest` should re-enqueue the corresponding `AppInstance`.

pkg/controller/quota/quota.go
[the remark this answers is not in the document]
tylerslaton: Did a slight variant on what you suggested. Let me know what you think.

pkg/controller/quota/quota.go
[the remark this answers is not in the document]
tylerslaton: Gotcha, I have a fix that should address this case then.

integration/run/run_test.go
```diff
@@ -1521,3 +1521,87 @@ func TestProjectUpdate(t *testing.T) {
...
+	helper.WaitForObject(t, helper.Watcher(t, c), &apiv1.AppList{}, app, func(obj *apiv1.App) bool {
+		return obj.Status.Condition(v1.AppInstanceConditionNamespace).Success &&
+			obj.Status.Condition(v1.AppInstanceConditionQuotaAllocated).Transitioning
+	})
+
+	// Grab the app's QuotaRequest and check that it has the appropriate values set.
+	app, err = c.AppGet(ctx, app.Name)
+	if err != nil {
+		t.Fatal(err)
+	}
```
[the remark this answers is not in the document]
tylerslaton: Good catch. No, we do not need to since `WaitForObject` should update the `app` object.

pkg/controller/quota/quota.go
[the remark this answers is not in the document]
tylerslaton: Alright, I have a fix for this that also has the tests passing. Resolving but let me know if you have other comments.

pkg/apis/internal.admin.acorn.io/v1/quotarequests.go
```diff
@@ -0,0 +1,70 @@
...
+	ProjectEnforcedQuotaAnnotation = "acorn.io/enforced-quota"
+)
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type QuotaRequestInstance struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	SupportedRegions []string                   `json:"supportedRegions,omitempty"`
```
thedadams, line 20: Apps only support one region, so I think these should, too.
```suggestion
	Region string                   `json:"supportedRegions,omitempty"`
```

pkg/controller/quota/quota.go
```diff
@@ -0,0 +1,168 @@
...
+
+// addStorage adds the storage resources of the volumes passed to the quota request.
+func addStorage(appInstance *apiv1.AppInstance, quotaRequest *adminv1.QuotaRequestInstance) error {
+	app := appInstance.Status.AppSpec
+
+	// Add the volume storage needed to the quota request. We only parse net new volumes, not
+	// existing ones that are then bound client-side.
+	for name, volume := range app.Volumes {
+		size := volume.Size
+		if bound, boundSize := boundVolumeSize(name, appInstance.Spec.Volumes); bound {
```
thedadams, line 127: If the volume is being bound, then it already exists. Would considering it here mean that the volume size is getting counted twice?

pkg/controller/quota/quota.go
```diff
@@ -0,0 +1,168 @@
...
+			return true, binding.Size
+		}
+	}
+	return false, "0"
+}
+
+func isEnforced(req router.Request, namespace string) (bool, error) {
+	project := v1.Project{}
+	if err := req.Client.Get(req.Ctx, router.Key("", namespace), &project); err != nil {
+		if errors.IsNotFound(err) {
```
thedadams, line 159: Why do we need to handle a not found error differently?

pkg/controller/quota/quota.go
```diff
@@ -0,0 +1,168 @@
...
+			return false, nil
+		}
+		return false, err
+	}
+	if enabled, found := project.Annotations[adminv1.ProjectEnforcedQuotaAnnotation]; !found || enabled != "true" {
+		return false, nil
+	}
+	return true, nil
```
thedadams, line 167: nit:
```suggestion
	return project.Annotations[adminv1.ProjectEnforcedQuotaAnnotation] == "true", nil
```

pkg/server/registry/apigroups/acorn/images/storage.go
```diff
@@ -11,9 +11,16 @@ import (
 	kclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// type strategyMiddleware func(strategy strategy.CompleteStrategy) strategy.CompleteStrategy
```
thedadams, line 14: Can these be removed?

pkg/server/registry/apigroups/acorn/secrets/storage.go
```diff
@@ -5,18 +5,26 @@ import (
...
+	"github.com/acorn-io/mink/pkg/strategy"
 	"github.com/acorn-io/mink/pkg/strategy/remote"
 	"github.com/acorn-io/mink/pkg/strategy/translation"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apiserver/pkg/registry/rest"
 	kclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func NewStorage(c kclient.WithWatch) rest.Storage {
+type strategyMiddleware func(strategy strategy.CompleteStrategy) strategy.CompleteStrategy
```
thedadams, line 16: This type is defined in several different places. Can we move this to be defined in one central place?

pkg/server/registry/apigroups/admin/quotarequest/translator.go
thedadams, line 1: Are quota requests exposed publicly? In other words, can we get away with only have `QuotaRequestInstance`?

pkg/apis/internal.acorn.io/v1/appinstance.go
```diff
@@ -205,6 +206,7 @@ type AppInstanceStatus struct {
 	Conditions             []Condition                `json:"conditions,omitempty"`
 	Endpoints              []Endpoint                 `json:"endpoints,omitempty"`
 	Defaults               Defaults                   `json:"defaults,omitempty"`
+	QuotaRequest           string                     `json:"quotaRequest,omitempty"`
```
thedadams, line 209: I didn't see that this was used anywhere.

pkg/server/registry/apigroups/admin/quotarequest/translator.go
tylerslaton: Yeah, I thought about this. It really depends on how the any review here goes. Short answer though - no, it isn't right now. I'd say its probably we don't need to have a storage layer for this at all since it's CRD is getting created.

integration/run/run_test.go
```diff
@@ -1554,3 +1554,89 @@ func TestProjectUpdate(t *testing.T) {
...
+		t.Fatal("error while getting rest config:", err)
+	}
+	// Create a project.
+	kclient := helper.MustReturn(kclient.Default)
+	project := helper.TempProject(t, kclient)
+	t.Cleanup(func() {
+		if err := kclient.Delete(ctx, project); err != nil {
+			t.Fatal(err)
+		}
+	})
```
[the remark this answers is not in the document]
tylerslaton: Yep, good call.

pkg/apis/internal.admin.acorn.io/v1/quotarequests.go
tylerslaton: Makes sense. Updated the region functions to reflect this change as well.

pkg/controller/quota/quota.go
```diff
@@ -0,0 +1,168 @@
...
+		return err
+	}
+
+	// Determine how to proceed depending on if the quotaRequest exists and what it has written to its status
+	if quotaRequest.Status.FailedResources != nil {
+		resp.DisablePrune()
+		status.Error(fmt.Errorf("failed to provision the following resources: %v", quotaRequest.Status.FailedResources.ToString()))
+	} else if quotaRequest.Status.Condition(adminv1.QuotaRequestCondition).Error {
+		status.Error(fmt.Errorf("error occurred while trying to allocate quota: %w", err))
+	} else if !equality.Semantic.DeepEqual(quotaRequest.Spec.Resources, quotaRequest.Status.AllocatedResources) {
```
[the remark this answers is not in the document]
tylerslaton: Yep!

pkg/controller/quota/quota.go
tylerslaton: Ah, right. I wrote this awhile ago so the current logic may get around this but this made it so that the quota request wouldn't error out if the project wasn't created yet. Basically the status would be updated with the equivalent of `project not found` for quota allocation which was a confusing UX since it happened consistently enough.

pkg/controller/quota/quota.go
```diff
@@ -0,0 +1,168 @@
...
+	if err != nil && !errors.IsNotFound(err) {
+		return err
+	}
+
+	// Determine how to proceed depending on if the quotaRequest exists and what it has written to its status
+	if quotaRequest.Status.FailedResources != nil {
+		resp.DisablePrune()
+		status.Error(fmt.Errorf("failed to provision the following resources: %v", quotaRequest.Status.FailedResources.ToString()))
+	} else if quotaRequest.Status.Condition(adminv1.QuotaRequestCondition).Error {
+		status.Error(fmt.Errorf("error occurred while trying to allocate quota: %w", err))
```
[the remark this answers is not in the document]
tylerslaton: Does it make more sense to be more opaque and just use `FailedResources` as a "something went wrong" communicator?

pkg/controller/quota/quota.go
tylerslaton: Ah, I think that the name I made is a bit confusing. This logic isn't actually determining if we'll bind to any existing volumes, it just determines if any Acornfile defined Volumes will get overwritten sizes. That brings me to two action items:

1. Rename this function
2. Do some querying logic to determine if this will bind to some existing Volume.

pkg/server/registry/apigroups/acorn/secrets/storage.go
tylerslaton: Yep, I made a `middleware` package for this now. What do you think?

pkg/server/registry/apigroups/admin/quotarequest/translator.go
tylerslaton: I went ahead and removed it. If anyone wants it to exist then we can come back to it.

pkg/controller/quota/quota.go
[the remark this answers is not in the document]
thedadams: You can bubble up the error however you'd like. What I am saying here is that I think the variable `err` will always be `nil` here.

pkg/controller/quota/quota.go
thedadams: I don't think any querying logic is necessary: `appInstance.Spec.Volumes` will tell you if the volume is being bound to an existing volume (i.e. the `Volume` field will be set).

pkg/controller/quota/quota.go
thedadams: You're saying that you were able to create an app or quota in a project that didn't exist?

pkg/controller/quota/quota.go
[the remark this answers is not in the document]
tylerslaton: Ah, I see now. Thank you and yes, you are correct.

pkg/controller/quota/quota.go
tylerslaton: I just tested this locally and you are right, there is not a need to handle a not-found error differently here. I'll go ahead and amend this change.

pkg/apis/internal.admin.acorn.io/v1/quotarequests.go
```diff
@@ -0,0 +1,74 @@
...
+	return in.Spec.Region
+}
+
+type QuotaRequestInstanceSpec struct {
+	Region    string    `json:"region,omitempty"`
+	Resources Resources `json:"resources,omitempty"`
+}
+
+type QuotaRequestInstanceStatus struct {
+	AllocatedResources Resources      `json:"approvedResources,omitempty"`
```
StrongMonkey, line 50: The json tag is not consistent, is that intentional?

pkg/controller/quota/quota.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+		status.Success()
+		return nil
+	}
+
+	// Attempt to get the quotaRequest for this appInstance. It should exist with the name and namespace of the
+	// appInstance being processed.
+	quotaRequest := &adminv1.QuotaRequestInstance{}
+	err := req.Client.Get(req.Ctx, router.Key(appInstance.Namespace, appInstance.Name), quotaRequest)
+	if err != nil && !errors.IsNotFound(err) {
```
StrongMonkey, line 43: are we igoring not found error but the quotarequest should always exist?

pkg/controller/quota/quota.go
```diff
@@ -0,0 +1,165 @@
...
+		return err
+	}
+
+	// Determine how to proceed depending on if the quotaRequest exists and what it has written to its status
+	if quotaRequest.Status.FailedResources != nil {
```
StrongMonkey, line 48: This part of the code takes me a while to understand.. would be nice to add some comments explaining each scenario? It is always good if you have a big if else statement and comment helps a lot in understanding it.

pkg/controller/quota/quota.go
```diff
@@ -0,0 +1,165 @@
...
+
+	return nil
+}
+
+// EnsureQuotaRequest ensures that the quota request exists and is up to date.
+func EnsureQuotaRequest(req router.Request, resp router.Response) error {
+	appInstance := req.Object.(*apiv1.AppInstance)
+
+	// Don't do anything if quota isn't enabled for this project
+	if enforced, err := isEnforced(req, appInstance.Namespace); !enforced {
```
StrongMonkey, line 67: looks like the err is not being checked

pkg/controller/quota/quota.go
StrongMonkey: Yeah, if project is guaranteed to exist, we should not assert not found error, and just let it fail if error happens.

pkg/roles/roles.go
```diff
@@ -147,6 +147,8 @@ var (
 					"clustervolumeclasses",
 					"projectcomputeclasses",
 					"clustercomputeclasses",
+					"quotarequests",
+					"imageallowrules",
```
StrongMonkey, line 151: Duplicated `imageallowrules`

pkg/server/registry/apigroups/acorn/apps/storage.go
```diff
@@ -14,11 +15,16 @@ import (
...
 )
 
-func NewStorage(c kclient.WithWatch, clientFactory *client.Factory, recorder event.Recorder) rest.Storage {
+func NewStorage(c kclient.WithWatch, clientFactory *client.Factory, recorder event.Recorder, middlewares ...middleware.CompleteStrategy) rest.Storage {
 	remoteResource := remote.NewRemote(&v1.AppInstance{}, c)
 	strategy := translation.NewSimpleTranslationStrategy(&Translator{}, remoteResource)
 	strategy = publicname.NewStrategy(strategy)
 	strategy = newEventRecordingStrategy(strategy, recorder)
+
+	for i := len(middlewares) - 1; i >= 0; i-- {
```
StrongMonkey, line 24:
```suggestion
	for _, middleware := range middleware
```

pkg/controller/quota/quota.go
tylerslaton: Talked offline and this function is now updated to work properly. I've also added a new function to handle Secrets and their bindings in a similar way.

pkg/apis/internal.admin.acorn.io/v1/quotarequests.go
tylerslaton: Good catch. Not intentional and will fix now.

pkg/controller/quota/quota.go
tylerslaton: QuotaRequests will not always exist, at least not immediately for AppInstances. However, in that event we enter the if tree after this line and will report the `status.Unknown()`. Doing it this way just saves some lines. What do you think though?

pkg/controller/quota/quota.go
tylerslaton: Great point. Let me add that and demystify this code a bit.

pkg/roles/roles.go
tylerslaton: Ah, must be a leftover merge conflict artifact. Thanks!

pkg/controller/routes.go
```diff
@@ -58,6 +60,8 @@ func routes(router *router.Router, registryTransport http.RoundTripper, recorder
 	appRouter := router.Type(&v1.AppInstance{}).Middleware(appdefinition.RequireNamespace, appdefinition.IgnoreTerminatingNamespace, appdefinition.FilterLabelsAndAnnotationsConfig)
 	appRouter.HandlerFunc(defaults.Calculate)
 	appRouter.HandlerFunc(scheduling.Calculate)
+	appRouter.HandlerFunc(quota.EnsureQuotaRequest)
```
StrongMonkey, line 63: The problem with this approach is that, the quota object will always be created after all the handlers are processed. This means that `quota.WaitForAllocation` is always going to fail for the first several times since quota object is not created...

pkg/controller/quota/quota.go
StrongMonkey: Got it. I commented on the other line. The way how handler are processed will make sure that the quota object is only created after all the handlers are processed

pkg/controller/routes.go
StrongMonkey: Is the quota going to be effective even if it has to be create in the end? 

Now I am thinking an app could possibly escape the quota because the quota are not created at first?

pkg/controller/quota/quota.go
tylerslaton: If i'm understanding your concern correctly, the below line will prevent the App from provisioning until it has the `QuotaAllocated` status set as to `Success`. This. will only occur if the `QuotaRequest` has been created and processed in such a way that denotes success (i.e the Spec.Resources and Status.AllocatedResources are equal).

https://github.com/acorn-io/acorn/blob/fb98a440b6e1b44e5c23636d87ed8244df162210/pkg/controller/appstatus/status.go#L41

StrongMonkey: Ok. That's good then.

pkg/server/registry/apigroups/acorn/apps/storage.go
StrongMonkey: actually, if you are applying in reverse order, then nvm

pkg/controller/routes.go
tylerslaton: Left a comment above (I believe) addressing these concerns,

pkg/server/registry/apigroups/acorn/secrets/storage.go
```diff
@@ -12,11 +13,16 @@ import (
...
-func NewStorage(c kclient.WithWatch) rest.Storage {
+func NewStorage(c kclient.WithWatch, middlewares ...middleware.CompleteStrategy) rest.Storage {
 	translated := translation.NewTranslationStrategy(&Translator{
 		c: c,
 	}, remote.NewRemote(&corev1.Secret{}, c))
 	remoteResource := publicname.NewStrategy(translated)
+
+	for i := len(middlewares) - 1; i >= 0; i-- {
+		remoteResource = middlewares[i](remoteResource)
+	}
```
thedadams, line 24: Since you moved the type to its own package, I'd like to see an `ApplyMiddlewares` function there that does this.

tylerslaton: Right on, I'll add that.

integration/run/run_test.go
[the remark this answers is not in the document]
thedadams: I don't think this comment was addressed.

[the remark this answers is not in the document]
tylerslaton: It was but I added it back in an attempt to fix the breaking tests 😓 This will be removed soon once I get those fixed.

pkg/apis/api.acorn.io/v1/types.go
```diff
@@ -273,9 +273,9 @@ type VolumeColumns struct {
 	AccessModes string `json:"accessModes,omitempty"`
 }
 
-// EnsureRegion checks or sets the region of a ContainerReplica.
-// If a ContainerReplica's region is unset, EnsureRegion sets it to the given region and returns true.
-// Otherwise, it returns true if and only if the ContainerReplica belongs to the given region.
+// EnsureRegion checks or sets the region of a Volume.
```
thedadams, line 276: 🙏🏻

pkg/apis/internal.admin.acorn.io/v1/quotarequests.go
```diff
@@ -0,0 +1,75 @@
+package v1
+
+import (
+	v1 "github.com/acorn-io/acorn/pkg/apis/internal.acorn.io/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	QuotaRequestCondition          = "quota-request"
+	ProjectEnforcedQuotaAnnotation = "acorn.io/enforced-quota"
```
thedadams, line 10: Can this be moved to the `labels` package with the other labels and annotations?

pkg/controller/quota/quota.go
```diff
@@ -0,0 +1,187 @@
...
+		quotaRequest.Spec.Resources.VolumeStorage.Add(parsedSize)
+	}
+
+	// Add the secrets needed to the quota request. We only parse net new secrets, not
+	// existing ones that are then bound client-side.
+	for name := range app.Secrets {
+		if boundSecret(name, appInstance.Spec.Secrets) {
+			continue
+		}
+		quotaRequest.Spec.Resources.Secrets = +1
```
thedadams, line 154: Most importantly, should this be `+=` instead of `= +`?
```suggestion
		if !boundSecret(name, appInstance.Spec.Secrets) {
			quotaRequest.Spec.Resources.Secrets += 1
		}
```

tylerslaton: That's super strange, I didn't even know that was supported behavior If I'm honest. Yeah I'll fix this, thank you for catching it.

pkg/controller/quota/quota.go
tylerslaton: I believe this not existing was leading to quite weird number of test failures. After revert it back to normal everything is passing again. Thoughts? Looking at the controller logs I saw a number of failures to sync Projects.

CC: @StrongMonkey @thedadams

pkg/server/registry/apigroups/acorn/secrets/storage.go
tylerslaton: Added. Thoughts?

pkg/controller/quota/quota.go
thedadams: The issue here seems to be that you are looking for the project at all. Since this is in the controller, the controller shouldn't be looking for `api.acorn.io` objects. Instead, it should be looking at the underlying/internal objects. In this case, the namespace.

tylerslaton: Gotcha, I updated this to use a Namespace instead. Let's see if the tests get through properly.

tylerslaton: Yep! That seems to have fixed it.

thedadams (approved):

pkg/controller/quota/quota.go
StrongMonkey: 💯

PR #1589, opened 2023-05-09
Determine app readiness based on deployment spec.Replicas rather than status.Replicas

g-linville:
Before this change, we were checking for app readiness by ensuring that the app's Deployment had matching `status.Replicas` and `status.ReadyReplicas`. If for some reason the Deployment fails to create any replicas (such as a mutating webhook that fails, if Istiod is down for example), then both `status.Replicas` and `status.ReadyReplicas` are set to 0, and Acorn thinks the app is ready even though it is not running at all.

With this change, we look at the Deployment spec instead to see the desired number of replicas.

### Checklist

### DIFF
--- a/pkg/controller/appstatus/status.go
+++ b/pkg/controller/appstatus/status.go
@@ -484,7 +484,7 @@ func AppStatus(req router.Request, resp router.Response) error {
 		if dep.Spec.Replicas != nil {
 			status.ReadyDesired = *dep.Spec.Replicas
 		} else {
-			status.ReadyDesired = 0
+			status.ReadyDesired = 1
 		}
 		status.UpToDate = dep.Status.UpdatedReplicas
 		status.Created = true
--- a/pkg/controller/appstatus/status.go
+++ b/pkg/controller/appstatus/status.go
@@ -481,7 +481,11 @@ func AppStatus(req router.Request, resp router.Response) error {
 
 		status := container[containerName]
 		status.Ready = dep.Status.ReadyReplicas
-		status.ReadyDesired = dep.Status.Replicas
+		if dep.Spec.Replicas != nil {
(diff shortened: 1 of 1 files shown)


pkg/controller/appstatus/status.go
```diff
@@ -481,7 +481,11 @@ func AppStatus(req router.Request, resp router.Response) error {
 
 		status := container[containerName]
 		status.Ready = dep.Status.ReadyReplicas
-		status.ReadyDesired = dep.Status.Replicas
+		if dep.Spec.Replicas != nil {
```
StrongMonkey, line 484: Hmm this should be dep.status.Replica?

g-linville: No, the `status.Replicas` has the possibility to be 0 (not set) if something goes wrong. `spec.Replicas` is a pointer so I have to do the nil check first before proceeding.

StrongMonkey: But dep.spec.replicas will never be the replica that is ready? It is the desired replica for a deployment, not the ready replica.

StrongMonkey: oh wait, this is `ReadyDesired`.... yeah setting spec.replica makes sense...

tylerslaton (approved):

pkg/controller/appstatus/status.go
```diff
@@ -481,7 +481,11 @@ func AppStatus(req router.Request, resp router.Response) error {
...
+			status.ReadyDesired = *dep.Spec.Replicas
+		} else {
+			status.ReadyDesired = 0
```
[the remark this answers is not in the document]
g-linville: Good call, fixed :+1:

iwilltry42 (approved):

Issue #1562, opened 2023-04-28
Add basic `acorn events` command


njhale:
<details>
    <summary> ORIGINAL ISSUE HERE </summary>
Add an `acorn events` command that prints a chronologically ordered set of `Events` related to Acorn resources within the currently selected project.

## Usage

<details>
    <summary> e.g. after creating and deleting a `Secret` </summary>

```sh
$ acorn --project acorn events -o json
[
  {
    "kind": "Event",
    "apiGroup": "api.acorn.io/v1",
    "metadata": {
      "name": "secret.1759efb44780fe3e",
      "namespace": "acorn"
    },
    "eventTime": "2023-04-27T23:51:33.389882Z",
    "type": "Normal",
    "action": "CreateSucceeded",
    "actor": "njhale",
    "note": "Secret was successfully created",
    "regarding": {
      "ref": {
        "apiVersion": "api.acorn.io/v1",
        "kind": "Secret",
        "name": "my-secret",
        "namespace": "acorn",
        "resourceVersion": "123222",
        "uid": "8630051f-4fb1-402f-8db4-4df444c59c9a"
      }
    }
  },
  {
    "kind": "Event",
    "apiGroup": "api.acorn.io/v1",
    "metadata": {
      "name": "secret.1759efb887a6s7d8ff",
      "namespace": "acorn"
    },
    "eventTime": "2023-04-27T23:52:32.389882Z",
    "type": "Normal",
    "action": "DeleteSucceeded",
    "actor": "njhale",
    "note": "Secret was successfully deleted",
    "regarding": {
      "ref": {
        "apiVersion": "api.acorn.io/v1",
        "kind": "Secret",
        "name": "my-secret",
        "namespace": "acorn",
        "resourceVersion": "123222",
        "uid": "8630051f-4fb1-402f-8db4-4df444c59c9a"
      }
    }
  }
]
```

</details>

<details>
  <summary> e.g. after creating an `App` </summary>

```sh
$ acorn --project acorn events -o json
[
  {
    "kind": "Event",
    "apiGroup": "api.acorn.io/v1",
    "metadata": {
      "name": "app.1759ef887as98da",
      "namespace": "acorn"
    },
    "eventTime": "2023-04-27T23:51:33.389882Z",
    "type": "Normal",
    "action": "CreateSucceeded",
    "actor": "njhale",
    "note": "App was successfully created",
    "regarding": {
      "obj": {
        "metadata": {
          "name": "cupule",
[...]

PR #1604, opened 2023-05-15
Add basic `acorn events` subcommand (#1562)

njhale:
### Description

Adds a basic `acorn events` subcommand (and respective server-side machinery).

### Features

The initial command supports:
- Listing events within a namespace
- Listing events across all namespaces (`-A`)
- Printing additional "event details" (`-d`)
- Limiting the number of events returned (`--tail`)

It **does not yet support**:
- Filtering of any kind
- Watching for new events
- Event deduplication

### Initial Events

In addition to the subcommand, this change causes acorn to emit several new events related to the life cycle of `Apps`:

- `AppSpecUpdate`
- `AppCreate`
- `AppDelete`
- `AppImagePullFailure`
- `AppImagePullSuccess`

### Example Help

```sh
$ acorn events --help
List events about Acorn resources

Usage:
  acorn events [flags]

Examples:
# List all events in the current project
  acorn events

  # List events across all projects
  acorn -A events

  # List the last 10 events 
  acorn events --tail 10

# Getting Details 
  # The 'details' field provides additional information about an event.
  # By default, this field is elided from this command's output, but can be enabled via the '--details' flag.
  acorn events --details


Flags:
  -d, --details         Don't strip event details from response
  -h, --help            help for events
  -o, --output string   Output format (json, yaml, {{gotemplate}})
  -t, --tail int        Return this number of latest events

Global Flags:
  -A, --all-projects        Use all known projects
      --debug               Enable debug logging
      --debug-level int     Debug log level (valid 0-9) (default 7)
      --kubeconfig string   Explicitly use kubeconfig file, overriding current project
  -j, --project string      Project to work in

```

### Example Output

```sh
$ acorn events --tail 2
NAME                               TYPE      ACTOR     SOURCE    OBSERVED    DESCRIPTION
8e2d224e7f506f2dff433d852462f732   Foo       bar       /         3d22h ago   A test fired, creating an App
bdf5b5d33ed0e5bdadb3b404e7c9872c   Foo       bar       /         3d22h ago   A test fired, creating an App
```

### Checklist


pkg/client/client.go
```diff
@@ -314,6 +318,27 @@ type ContainerReplicaListOptions struct {
...
+	Tail    *int `json:"tail,omitempty"`
+}
+
+func (o EventStreamOptions) ListOptions() *kclient.ListOptions {
+	var listOpts kclient.ListOptions
+	if o.Tail != nil {
+		listOpts.Limit = int64(*o.Tail)
+	}
+
+	// TODO(njhale): Only "withContext" needs to be sent to the server.
```
njhale, line 332: nit: old comment, drop it

pkg/client/event.go
```diff
@@ -0,0 +1,32 @@
+package client
+
+import (
+	"context"
+
+	apiv1 "github.com/acorn-io/acorn/pkg/apis/api.acorn.io/v1"
+)
+
+func (c *DefaultClient) EventStream(ctx context.Context, opts *EventStreamOptions) (<-chan apiv1.Event, error) {
```
njhale, line 9: I opted for returning a channel so we can implement watching later without needing a separate method.

pkg/client/multi.go
```diff
@@ -511,6 +513,43 @@ func (m *MultiClient) Info(ctx context.Context) ([]apiv1.Info, error) {
...
+				}
+			}
+
+			return nil
+		})
+	}
+	go func() {
+		defer close(all)
+		if err := g.Wait(); err != nil {
+			logrus.Errorf("Error streaming events: %s", err.Error())
```
njhale, line 546: Should I be returning errors like these over the channel instead of logging? (would require a small refactor to a broader channel type)

pkg/controller/appdefinition/pullappimage.go
```diff
@@ -96,3 +112,62 @@ func determineTargetImage(appInstance *v1.AppInstance) (string, string) {
...
+	// ResolvedImage is the image, post resolution.
+	// +optional
+	ResolvedImage string `json:"resolvedImage,omitempty"`
+
+	// Err the error that occurred during resolution, if any.
+	// +optional
+	Err string `json:"err,omitempty"`
+}
+
+func recordResolutionEvent(ctx context.Context, recorder event.Recorder, obj kclient.Object, err error, targetImage, resolvedImage string) {
```
cjellick, line 138: Shared directly, but wanted to post here too. this definitely makes sense.

Here's some additional information we could add:
- Whether this was triggered by auto-upgrade or not (you could make `determineTargetImage` return an extra field to accomplish this)
- app.status.appImage.Digest and app.status.appImage.VCS
- Would it be overkill to also record the "previous values" for targetImage, resolvedImage, digest, and vcs? that would be fairly powerful to have
- TargetImage and ResolvedImage suck as names to expose to the user, but i think the meaning of these two fields is slightly different depending on whether the app auto-upgrades or not. let me confirm that and get back to you with slightly better names

cjellick: Ok, ran through a few scenarios. "ResolvedImage" is kind of an implementation detail that I think we should drop. It is used to get an image ID if the image in question is "local" (not from a remote registry). If the image ISNT local, the value is the same as targetImage. Furthermore, the ID happens to be the same value as the digest field with the "sha256:" prefix stripped off. So, I dont think it holds much value. Just drop it.


I think targetImage is a good name to keep for the other field, or could just call it image.

njhale: I refactored this and tried to hit all your suggestions.

The largest semantic change is moving the "topic" of the events generated here from "Resolution" to "Pull".

pkg/event/id.go
```diff
@@ -0,0 +1,42 @@
...
+		e.Description,
+		e.Observed.UTC().Format(time.RFC3339),
+	}, ",")
+
+	h := fnv.New128a()
+	if _, err := h.Write([]byte(fieldSet)); err != nil {
+		return "", err
+	}
+
+	digest := h.Sum([]byte{})
```
thedadams, line 31: nit: I have always seen this
```suggestion
	digest := h.Sum(nil)
```

pkg/server/registry/apigroups/acorn/apps/events.go
```diff
@@ -0,0 +1,198 @@
...
+		CompleteStrategy: s,
+		recorder:         recorder,
+	}
+}
+
+func (s *eventRecordingStrategy) Create(ctx context.Context, obj types.Object) (types.Object, error) {
+	created, err := s.CompleteStrategy.Create(ctx, obj)
+	if err != nil {
+		// Return created because CompleteStrategy.Create is a black box; i.e. we can't assume
+		// updated is nil when err is non-nil.
```
thedadams, line 64: nit (and same below)
```suggestion
		// created is nil when err is non-nil.
```

pkg/server/registry/apigroups/acorn/apps/events.go
```diff
@@ -0,0 +1,198 @@
...
+		return created, err
+	}
+
+	defer func() {
```
thedadams, line 68: Why is this deferred when nothing seems to happen afterwards? Same below.

pkg/server/registry/apigroups/acorn/apps/events.go
```diff
@@ -0,0 +1,198 @@
...
+		details, err := v1.Mapify(AppSpecCreateEventDetails{
+			ResourceVersion: created.GetResourceVersion(),
+		})
```
thedadams, line 71: Is there a benefit to using concrete types here rather than just a `GenericMap` directly?

pkg/server/registry/apigroups/acorn/events/strategy.go
```diff
@@ -0,0 +1,80 @@
...
+			return f, v, nil
+		}
+
+		return "", "", err
+	})
+	if err != nil {
+		return nil, err
+	}
+	opts.Predicate.Field = stripped
+	logrus.Warnf("query: [%v]", q)
```
thedadams, line 39: Remove?

pkg/server/registry/apigroups/acorn/events/strategy.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+
+		if !q.details {
+			event.Details = nil
+		}
+
+		result = append(result, event)
+	}
+
+	list.Items = result // TODO(njhale): Will this cause an inconsistent metav1.ListMeta?
```
thedadams, line 77: nit, does this simplify like so?
```suggestion
	if q.details {
		return list, nil
	}

	for i, event := range list.Items {
		event.Details = nil
		list.Items[i] = event
	}
```

pkg/server/registry/apigroups/acorn/events/validator.go
```diff
@@ -0,0 +1,36 @@
...
+	if err != nil {
+		result = append(result, field.InternalError(
+			field.NewPath("metadata", "name"),
+			fmt.Errorf("failed to generate content ID for event: %w", err),
+		))
+		return
+	}
+
+	if e.Name != id {
+		result = append(result, field.NotSupported(
```
thedadams, line 28: Is `NotSupported` correct here? It seems to me that the name would be invalid?

pkg/apis/internal.acorn.io/v1/event.go
```diff
@@ -0,0 +1,82 @@
...
+type EventSeverity string
+
+// EventSource identifies an object related to an Event.
+//
+// The referenced object may or may not exist.
+//
+// Note: corev1.ObjectReference was explicitly avoided because its use in new schemas is discouraged.
+// See https://github.com/kubernetes/api/blob/cdff1d4efea5d7ddc52c4085f82748c5f3e5cc8e/core/v1/types.go#L5919
+// for more details.
+type EventSource struct {
```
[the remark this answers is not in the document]
njhale: > Is the assumption that the group would always be api.acorn.io?

That's the assumption, yes.

pkg/cli/events.go
```diff
@@ -0,0 +1,63 @@
...
+`})
+	return cmd
+}
+
+type Events struct {
+	Tail    *int   `usage:"Return this number of latest events" short:"t"`
+	Details bool   `usage:"Don't strip event details from response" short:"d"`
+	Output  string `usage:"Output format (json, yaml, {{gotemplate}})" short:"o"`
+	client  ClientFactory
+}
```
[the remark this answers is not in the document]
njhale: I was planning on introducing an optional positional filter argument -- in a separate PR -- that lets you filter on source type, source name, and the event name. Here's what the help for that looks like:

```
# Filtering by Resource 
  # The FILTER argument comes in the form '<kind>/<name>' and, when given, prunes the result set to contain events related to Acorn resources with the respective kind and name.
  # List events related to the 'hello' app in the current project
  acorn events app/hello
  
  # The '/<name>' suffix is optional and, when omitted, matches events related to any resource of the given kind.
  # List events related to any volume in the current project
  acorn events vol

# Filtering by Name Prefix
  # If FILTER doesn't have a '/name' suffix and isn't a known Acorn kind, it's value is interpreted as an event name prefix.
  # List events with names starting with '175d849699e0d4c5' in the current project
  acorn events 175d849699e0d4c5

```

pkg/client/event.go
```diff
@@ -0,0 +1,32 @@
...
+	listOpts := opts.ListOptions()
+	listOpts.Namespace = c.Namespace
+
+	var current apiv1.EventList
+	if err := c.Client.List(ctx, &current, listOpts); err != nil {
```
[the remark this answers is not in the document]
njhale: yup, so I actually implemented that _first_

```go
func (c *DefaultClient) EventStream(ctx context.Context, opts *EventStreamOptions) (<-chan apiv1.Event, error) {
	listOpts := opts.ListOptions()
	listOpts.Namespace = c.Namespace

	var (
		current apiv1.EventList
		w       watch.Interface
		err     error
	)
	if opts.Watch {
		w, err = c.Client.Watch(ctx, &apiv1.EventList{}, &kclient.ListOptions{})
	} else {
		err = c.Client.List(ctx, &current, listOpts)
	}
	if err != nil {
		return nil, err
	}

	events := make(chan apiv1.Event)
	go func() {
		defer close(events)
		// Send the current set of events
		for _, c := range current.Items {
			select {
			case <-ctx.Done():
				return
			case events <- c:
			}
		}

		if w == nil {
			// Don't need to watch
			return
		}

		defer w.Stop()

		for {
			select {
			case <-ctx.Done():
				return
			case e := <-w.ResultChan():
				switch e.Type {
				case kwatch.Error:
					return
				case kwatch.Added, kwatch.Modified:
					events <- *(e.Object.(*apiv1.Event))
				}
			}
		}
	}()

	return events, nil
}
```

I decided to roll this back because I figured it'd be contentious + table formatting gets a bit funky when we start flushing multiple times.

As it stands, I'm _very_ on the fence; should I just implement the vanilla list client instead?

pkg/controller/appdefinition/pullappimage_test.go
```diff
@@ -70,3 +81,209 @@ func app(specImage, statusImageID, statusAvailableImage, statusConfirmUpgradeIma
...
+			assert.Len(t, recording, 0)
+			return
+		}
+
+		assert.Len(t, recording, 1)
+		assert.EqualValues(t, expect, recording[0])
+	})
+}
+
+// func testRecordPullEvent(t *testing.T, obj kclient.Object, autoUpgradeOn bool, err error, previousImage, targetImage v1.AppImage, expect *apiv1.Event) {
```
njhale, line 273: nit: drop old comment

pkg/controller/appdefinition/pullappimage_test.go
```diff
@@ -70,3 +81,209 @@ func app(specImage, statusImageID, statusAvailableImage, statusConfirmUpgradeIma
...
+	// Error during local image resolution records NO events
+	testRecordPullEvent(t,
+		"ResolutionError",
+		withMeta("foo", "", "", app("acorn.io/img:1", "", "", "", false, false)),
+		resolveImageTo("", false, fmt.Errorf("resolution error!")),
+		nil,
+		now,
+		nil,
+	)
+	// testTargetImage(t, app("acorn.io/img:#", "acorn.io/img:1", "", "acorn.io/img:2", false, true), "", "confirm upgrade to acorn.io/img:2")
```
njhale, line 201: nit: drop comment

pkg/server/registry/apigroups/acorn/events/strategy.go
njhale: I believe so.

For context, this code originally had more filtering logic to drop specific events from the result set. I left it like this so I could quickly add that logic back.

That all being said, I think it's probably smarter to follow your suggestion to simplify this.

pkg/apis/internal.acorn.io/v1/event.go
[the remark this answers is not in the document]
thedadams: I could see us adding events for other groups, like when an AWS account is create/updated/deleted.

We can cross that bridge when we come to it, but it also might be a straightforward add to these changes. Your call.

pkg/client/event.go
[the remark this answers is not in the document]
thedadams: I think it's fine.

pkg/server/registry/apigroups/acorn/events/strategy.go
```diff
@@ -0,0 +1,80 @@
...
+
+		return "", "", err
+	})
+	if err != nil {
+		return nil, err
+	}
+	opts.Predicate.Field = stripped
+	logrus.Warnf("query: [%v]", q)
+
+	// TODO(njhale): Filtering like this adds an extra O(n*lgn) time and O(n) space to every List call.
```
tylerslaton, line 41: For now this is fine. I think we can optimize if this becomes a major scaling factor for us that'll be a contained segment of work when we do so.

pkg/apis/internal.acorn.io/v1/event.go
```diff
@@ -0,0 +1,82 @@
...
+	// It's typically used to embed the subject resource, in its entirety, at the time the Event occurred,
+	// but can be used to hold any data related to the event.
+	//
+	// +optional
+	Details GenericMap `json:"details,omitempty"`
+}
+
+const (
+	EventSeverityInfo EventSeverity = "info"
+	EventSeverityWarn EventSeverity = "warn"
```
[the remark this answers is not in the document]
tylerslaton: This could then be used in the sections I commented about an `error` occurring during the `CRUD` operations of an object.

pkg/client/multi.go
thedadams: This may be fine?

If an error occurs here, then the error will be printed and then the channel will be closed. Closing the channel will indicate to any "watchers" that nothing else is coming.

Is this OK?

njhale: yeah, I think that part is okay -- I'm mostly concerned about how the error looks to users -- this skips any formatting that happens downstream

pkg/server/registry/apigroups/acorn/apps/events.go
njhale: I think there's a good case to be made for documenting event details.

pkg/server/registry/apigroups/acorn/apps/events.go
njhale: dropped

pkg/server/registry/apigroups/acorn/apps/events.go
```diff
@@ -0,0 +1,198 @@
...
+func newEventRecordingStrategy(s strategy.CompleteStrategy, recorder event.Recorder) *eventRecordingStrategy {
+	return &eventRecordingStrategy{
+		CompleteStrategy: s,
+		recorder:         recorder,
+	}
+}
+
+func (s *eventRecordingStrategy) Create(ctx context.Context, obj types.Object) (types.Object, error) {
+	created, err := s.CompleteStrategy.Create(ctx, obj)
+	if err != nil {
```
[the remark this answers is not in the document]
njhale: I think it definitely could, but I'd prefer to add that in a follow up PR if we decide to add it. WDYT?

pkg/server/registry/apigroups/acorn/apps/events.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+	}()
+
+	return created, nil
+}
+
+func (s *eventRecordingStrategy) Delete(ctx context.Context, obj types.Object) (types.Object, error) {
+	deleted, err := s.CompleteStrategy.Delete(ctx, obj)
+	if err != nil {
+		// Return deleted because CompleteStrategy.Delete is a black box; i.e. we can't assume
```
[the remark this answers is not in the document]
njhale: see https://github.com/acorn-io/acorn/pull/1604#discussion_r1198403358

pkg/server/registry/apigroups/acorn/events/strategy.go
njhale: I concur, I dropped the comment.

pkg/apis/internal.acorn.io/v1/event.go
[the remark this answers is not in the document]
njhale: My preference is to keep the API surface as tight as possible for now. Like you said, adding a field is generally a backwards compatible change anyway; let's cross that bridge when we come to it.

njhale:
I believe I've addressed all the feedback and a few odds and ends. Asking for re-reviews.

pkg/server/registry/apigroups/acorn/apps/events.go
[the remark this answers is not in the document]
tylerslaton: Sure, don't want to delay this if we don't need to.

g-linville (approved):
Great work!

Issue #1561, opened 2023-04-28
Not able to bind apps to existing volumes in "released" state.


sangee2004:
acorn version v0.7.0-alpha1-16-g9007dba3+9007dba3

Steps to reproduce the problem:
1. Deploy app with volumes.
2. Deplete app (without deleting volumes)
3. Volumes get to released state.
4. Deploy app by binding this volume.

`acorn run -n testvolbind -v testvoldata.my-data:my-data -f AcornfileBind`
AcornfileBind
```
containers: {
mynginx: {
    ports: publish: "80/http"
    image: "nginx"
    dirs: {
        "/usr/share/nginx/html/test1": "volume://my-data"
     }
 }
}
```

App deployment fails to succeed with following error:

```
 acorn apps
NAME                           IMAGE                                                           HEALTHY   UP-TO-DATE   CREATED    ENDPOINTS                                                                          MESSAGE
testvolbind                    ddca10edbf20                                                    0         0            18s ago                                                                                       [defined: PersistentVolume "testvoldata.my-data" not found] [controller: [routes.go:57] PersistentVolume "testvoldata.my-data" not found]
```

```
acorn volumes
NAME                  APP-NAME              BOUND-VOLUME   CAPACITY   VOLUME-CLASS   STATUS     ACCESS-MODES   CREATED
testvoldata.my-data   testvoldata.my-data   my-data        10G        local-path     released   RWO            14m ago
```

sangee2004:
E2E Test case failure:
volume_test.go->TestVolumeBind

sangee2004:
Tested with `acorn version v0.7.0-alpha1-125-g071f902a+071f902a` 

This issue is not seen any more.

Able to bind apps to existing volumes successfully.

```
acorn volumes
NAME                                       APP-NAME              BOUND-VOLUME   CAPACITY   VOLUME-CLASS   STATUS    ACCESS-MODES   CREATED
pvc-360d6a99-72f4-4ca4-a2ef-0f0e51296f6c   testvolbind.my-data   my-data-bind   10Gi       ebs-retain     bound     RWO            5m56s ago
```

PR #1615, opened 2023-05-16
Fix volume binding name bug (#1561)

g-linville:
#1561

Before this fix, Acorn was expecting the user to use the name of the PersistentVolume in Kubernetes when binding an existing volume into a new app, for example:

```bash
acorn run -n myApp -v pvc-447f1f93-5b81-417c-a938-918d6f68e7ed:vol .
```

PV names in k8s are very long and we don't even expose them to the user, so it is better to also make the public volume name an option for binding.

With this fix, the user can now use the public volume name when binding an existing volume. For example:

```bash
❯ acorn volume
NAME           APP-NAME       BOUND-VOLUME   CAPACITY   VOLUME-CLASS   STATUS     ACCESS-MODES   CREATED
voltest.data   voltest.data   data           1G         local-path     released   RWO            32m ago

❯ acorn run -n myapp -v voltest.data:vol .
```

(Using the PV name also still works, as long as the PV is managed by Acorn and belongs to the same project as the app.)

Acorn will use labels to find the corresponding PersistentVolume and then reassign it to the new app. As that happens, the public name of the existing PV gets changed (along with the other labels).

### Tests

The new testdata YAML files that I added are big, but mostly just copies of the ones we already had for ephemeral and named binding. The only thing I changed is the `volume` parameter of the VolumeBinding so that we have one set of tests that uses the real PV name and another set that uses the public name.

I also created unit tests specifically for my new function that locates the PV for a new volume binding.

### Checklist


pkg/controller/appdefinition/volume.go
```diff
@@ -155,7 +156,31 @@ func toPVCs(req router.Request, appInstance *v1.AppInstance) (result []kclient.O
...
+
+			pvList := new(corev1.PersistentVolumeList)
+			if err := req.Client.List(req.Ctx, pvList, &kclient.ListOptions{LabelSelector: selector}); err != nil {
+				return nil, err
+			}
+			if len(pvList.Items) != 1 {
+				return nil, fmt.Errorf("expected 1 PV for volume %s, found %d", volumeBinding.Volume, len(pvList.Items))
+			}
+
+			pvc.Spec.VolumeName = pvList.Items[0].Name
```
thedadams, line 183: Spoke "offline" about this. These changes make it impossible for a user to bind an existing volume created outside of Acorn. We discussed the approach of how to handle these two cases.

pkg/controller/appdefinition/volume.go
```diff
@@ -155,7 +156,31 @@ func toPVCs(req router.Request, appInstance *v1.AppInstance) (result []kclient.O
 
 		if bind {
 			pvc.Name = bindName(vol)
-			pvc.Spec.VolumeName = volumeBinding.Volume
+
```
cjellick, line 159: I've thought through the below logic in the context of whether or not you could use this to "steal" another unbound volume from another acorn project. Looks like the
```
				labels.AcornAppNamespace: appInstance.Namespace,
```
bit of the selector prevents that. Is that right?

g-linville:
The logic should be good now. Going to work on adding more test cases though.

pkg/controller/appdefinition/volume.go
cjellick: Didn't realize we supported that! Are we sure we want to? You have any concrete use cases in mind?

thedadams: It is essentially the same thing that happens [here](https://www.acorn.io/hosting-the-pi-hole-ad-blocker-on-my-kubernetes-home-lab/) except that you could bind to a volume with a different name.

pkg/controller/appdefinition/volume.go
g-linville: Yeah, I think with the way things are right now you can steal an unbound PV from another project. My changes should prevent that from being possible.

ibuildthecloud:
No context comments: 

A user should only be able to bind a PV that has the acorn managed label. 

Cross project binding should definitely not be supported right now. 

If the volume name changes after binding that’s weird. Can it just always have the original name?

g-linville:
> If the volume name changes after binding that’s weird. Can it just always have the original name?

It's not the name of the PV that changes, but the value of the `acorn.io/public-name` label on the PV. This value becomes the same thing that is shown to the user when they list volumes by running `acorn volume`.

g-linville:
Ready for review again. Changed the logic based on our discussion earlier today, so that all apps can only bind volumes that are managed by Acorn and part of the same project as the app - no way to "adopt" a volume from the outside until we add that as a different subcommand (unless you manually throw labels onto the PV).

g-linville:
Converting to draft and unassigning all reviewers. I just remembered that I still need to make sure that binding volumes inside of the Acornfile is properly locked down. My top priority issue changed so I will need to come back to this later.

pkg/controller/appdefinition/volume.go
```diff
@@ -206,6 +209,55 @@ func toPVCs(req router.Request, appInstance *v1.AppInstance) (result []kclient.O
...
+	pvList := new(corev1.PersistentVolumeList)
+	if err := req.Client.List(req.Ctx, pvList, &kclient.ListOptions{LabelSelector: selector}); err != nil {
+		return nil, err
+	}
+	if len(pvList.Items) != 1 {
+		if len(pvList.Items) == 0 {
+			return nil, apierrors.NewNotFound(schema.GroupResource{Group: "", Resource: "persistentvolumes"}, binding.Volume)
+		}
+		return nil, fmt.Errorf("expected 1 PV for volume %s, found %d", binding.Volume, len(pvList.Items))
```
tylerslaton, line 255: What conditions will lead to this happening? Are we able to have more than one PV with the same public name?

tylerslaton (approved):

tylerslaton (changes requested):
Oops, didn't mean to approved.

pkg/controller/appdefinition/volume.go
g-linville: As far as I am aware, each PV public name should be unique within a project. I'm technically not even sure that this is possible, but I figured it would be best to check for that case anyway.

tylerslaton: Yeah, doesn't hurt. Just wanted to check 👍🏻

pkg/controller/appdefinition/volume.go
```diff
@@ -206,6 +209,55 @@ func toPVCs(req router.Request, appInstance *v1.AppInstance) (result []kclient.O
 	return
 }
 
+func getPVForVolumeBinding(req router.Request, appInstance *v1.AppInstance, binding v1.VolumeBinding) (*corev1.PersistentVolume, error) {
```
tylerslaton, line 212: Does it makes sense for this to be a public function? I ask because I have a use case where I would use this exact function in a different controller I'm writing.

tylerslaton: Or rather, I would use the "detect if this binding will bind to an existing volume" logic.

g-linville: It's public now

g-linville:
From talking with Darren, it looks like there isn't any other way to bind volumes currently besides using `-v`. So I believe that all these changes are sufficient.

pkg/controller/appdefinition/volume.go
```diff
@@ -206,6 +209,55 @@ func toPVCs(req router.Request, appInstance *v1.AppInstance) (result []kclient.O
 	return
 }
 
+func GetPVForVolumeBinding(req router.Request, appInstance *v1.AppInstance, binding v1.VolumeBinding) (*corev1.PersistentVolume, error) {
```
thedadams, line 212: nit: this looks very specific to this implementation. Should this be unexported?

pkg/controller/appdefinition/volume.go
```diff
@@ -206,6 +209,55 @@ func toPVCs(req router.Request, appInstance *v1.AppInstance) (result []kclient.O
...
+	// binding.Volume can either be the actual name of the PersistentVolume, or its public name in Acorn.
+	// Check for the actual name first.
+	pv := new(corev1.PersistentVolume)
+	if err := req.Client.Get(req.Ctx, kclient.ObjectKey{Name: binding.Volume}, pv); err != nil && !apierrors.IsNotFound(err) {
+		return nil, err
+	} else if err == nil {
+		// make sure this PV is managed by acorn
+		if _, ok := pv.Labels[labels.AcornManaged]; !ok {
+			return nil, apierrors.NewNotFound(schema.GroupResource{Group: "", Resource: "persistentvolumes"}, binding.Volume)
```
thedadams, line 221: Returning not-found errors here could be confusing for a user.

Let's say someone is running Acorn on a cluster they have admin controls over. If they create a volume and then try to bind said volume, they would see a status indicating that the volume was not found. However, they would be able to see that the volume would exist, and, likely, create an issue with Acorn.

Could this error be changed to indicate that no "acorn managed" volume was found? Does the error returned here need to be a Kubernetes API error?

I have the same comments/questions about the other errors. Obviously, the error needn't be too descriptive, but something other than a plain "not found" would be better, in my opinion.

pkg/controller/appdefinition/volume_test.go
```diff
@@ -165,3 +167,165 @@ func TestVolumeLabelsAnnotations(t *testing.T) {
...
+				Labels: map[string]string{},
+			},
+		},
+	}
+
+	// I don't know of a better way to convert this to generic kclient.Object
+	var objList []kclient.Object
+	for _, pv := range pvList {
+		objList = append(objList, pv.DeepCopy())
+	}
```
thedadams, line 328: Does this work?
```suggestion
	pvList := []kclient.Object{
		&corev1.PersistentVolume{
			ObjectMeta: metav1.ObjectMeta{
				Name: "pv2",
				Labels: map[string]string{
					labels.AcornAppNamespace: "proj1",
					labels.AcornPublicName:   "app2.volume",
					labels.AcornManaged:      "true",
				},
			},
		},
		&corev1.PersistentVolume{
			ObjectMeta: metav1.ObjectMeta{
				Name: "pv3",
				Labels: map[string]string{
					labels.AcornAppNamespace: "other-project",
					labels.AcornPublicName:   "app3.volume",
					labels.AcornManaged:      "true",
				},
			},
		},
		&corev1.PersistentVolume{
			ObjectMeta: metav1.ObjectMeta{
				Name:   "not-acorn-managed",
				Labels: map[string]string{},
			},
		},
	}
```

docs/docs/50-running/04-volumes.md
```diff
@@ -30,14 +30,27 @@ The volume class used, including the default, may have restrictions on the size
 
 You can see a list of available volume classes and their restrictions, if any, with the [`acorn offerings volumeclasses`](100-reference/01-command-line/acorn_offerings_volumeclasses.md) command.
 
-## Using precreated volumes
+## Using pre-existing volumes
```
thedadams, line 33: Do we want to add something about not being able to use pre-existing Persistent Volumes created outside of Acorn?

pkg/controller/appdefinition/volume.go
g-linville: Yeah that sounds like a good idea. My rationale for plain NotFound errors was that I did not want to indicate whether a PV existed with the provided name, as ordinary users would be able to abuse that to try to guess names of other volumes that exist in the cluster (and also the names of other apps). But returning a more Acorn-specific error sounds good as long as I am still properly obscuring that information.

pkg/controller/appdefinition/volume_test.go
g-linville: That works, thanks!

pkg/controller/appdefinition/volume.go
iwilltry42: This confused me during review. Actually @tylerslaton asked for it to be exported as he could use it elsewhere: https://github.com/acorn-io/acorn/pull/1615#discussion_r1199160483

iwilltry42 (approved):

thedadams (approved):

PR #1624, opened 2023-05-19
Only build test summary when tests run

tylerslaton:
This makes it so that the test summary doesn't run everytime, only when the no previous steps (including the running of the tests) failed.


njhale (approved):
💯

Issue #1622, opened 2023-05-18
Improve VCS info captured by acorn when building images


njhale:
When building images, acorn captures some very basic VCS info about the repository in the current working directory. Today, this denotes only:
1. The commit hash of `HEAD`
2. Whether the working directory has uncommitted changes

We want to enrich the set of info captured to include:
- A reference to the upstream [remote]
- Whether the workspace is "clean" -- meaning NO non-ignored local files -- or not

See also: [vcs.go](https://github.com/acorn-io/acorn/blob/1d155cbe8be6bc4888ce7d9e78c625052107e390/pkg/vcs/vcs.go#L8)

sangee2004:
Tested with - acorn version v0.7.0-alpha1-118-g6adb3fe9+6adb3fe9

`AppImagePullSuccess` event now includes the following information in "vcs" section:

```
            "vcs": {
                "modified": true,
                "remotes": [
                    "git@github.com:sangee2004/hub.git",
                    "git@github.com:acorn-io/hub.git"
                ],
                "revision": "c6a2c32111362fad04967b7832cd0854662c83b4"
            }

There is 1 issue with "modified" always being reported as "true".

Scenarios:
1. Deploy app from git hub repo without making any local changes. 

previous:
```
            "vcs": {}
        },
```
target:
```
            "vcs": {
                "modified": true,
                "remotes": [
                    xx
                ],
                "revision": "c6a2c32111362fad04967b7832cd0854662c83b4"
            }
```
 
In this case we expect "modified": false for "target"

2. Make some local changes but do not commit the changes.

```
        "previous": {
            "digest": "sha256:19eef54625c3132a759374490db3af209e5e974acb794381dc9c5a9be9f006fe",
            "name": "19eef54625c3132a759374490db3af209e5e974acb794381dc9c5a9be9f006fe",
            "vcs": {
                "modified": true,
                "remotes": [xxx
                ],
                "revision": "32c5239dc55b4a8f5981b657c5822237a4589f7d"
            }
        },
        "resourceVersion": "1222348",
        "target": {
            "digest": "sha256:80c8da7939201347c9f2e7060dbdaee55fba8f38a13eee1aabe82e038a531b0b",
            "name": "80c8da7939201347c9f2e7060dbdaee55fba8f38a13eee1aabe82e038a531b0b",
            "vcs": {
                "modified": true,
                "remotes": [
                   xxx
                ],
[...]

PR #1640, opened 2023-05-23
Improve VCS info (#1622)

njhale:
### Checklist


njhale:
The primary question I have is: Is the remote ref enough? Do we need to also include the remote tracking branch? I'm not sure the tracking branch information is super helpful if, for example, you're working out of a feature branch that's deleted later. That being said, the HEAD commit can also change when merged.

pkg/vcs/vcs.go
```diff
@@ -25,8 +25,20 @@ func VCS(path string) (result v1.VCS) {
...
 	}
 
-	return v1.VCS{
+	result = v1.VCS{
 		Revision: head.Hash().String(),
 		Modified: !s.IsClean(),
 	}
+
+	// Set optional remotes field
+	remotes, err := repo.Remotes()
```
StrongMonkey, line 34: For my own education, what would be the usecase to know where is the remote in the git repo that image is built?

tylerslaton (approved):

tylerslaton:
> The primary question I have is: Is the remote ref enough? Do we need to also include the remote tracking branch? I'm not sure the tracking branch information is super helpful if, for example, you're working out of a feature branch that's deleted later. That being said, the HEAD commit can also change when merged.

Shouldn't the remote + the commit be enough to find exactly what the change is?

pkg/vcs/vcs.go
njhale: So, including remotes means that, in some cases, the HEAD can be traced back to an actual git object. I think there are quite a few usecases that can make use of this, but the immediate example for me is `AppImagePullSuccess`; in which we track the VCS info for images both before and after a successful pull. Given both, a user could potentially produce a source diff to make the update easier to grok.

njhale:
> > The primary question I have is: Is the remote ref enough? Do we need to also include the remote tracking branch? I'm not sure the tracking branch information is super helpful if, for example, you're working out of a feature branch that's deleted later. That being said, the HEAD commit can also change when merged.
> 
> Shouldn't the remote + the revision* be enough to find exactly what the change is?

@tylerslaton in most cases it is. I think we can also get into a situation where the image is built from a local commit that hasn't be pushed to any remote. In that case, our VCS info will show a revision that doesn't exist in any of the listed remotes.

tylerslaton:
> @tylerslaton in most cases it is. I think we can also get into a situation where the image is built from a local commit that hasn't be pushed to any remote. In that case, our VCS info will show a revision that doesn't exist in any of the listed remotes.
@njhale In that particular case, can we just label it something like "local"?

StrongMonkey (approved):

njhale:
> @njhale In that particular case, can we just label it something like "local"?

We totally could, but that would require making some extra calls to the remote repos to check. We can always add this feature in later, so I'm hoping to get this in as is for now.

PR #1641, opened 2023-05-23
Get Project before modifying it in Quota test

tylerslaton:
The nightly integration test failed with the following error:

```
run_test.go:1582: Operation cannot be fulfilled on namespaces "acorn-project-6xvr9": the object has been modified; please apply your changes to the latest version and try again
```

As such I am updating the code to get the Project before modifying it. Additionally, I am fixing an issue where the message that was output for the failing tests did not have the correct count for failing test cases. Lastly, I am changing the format for `gotestsum` to display each test case to make it easier to see especially long test cases.

### Checklist

### DIFF
--- a/integration/run/run_test.go
+++ b/integration/run/run_test.go
@@ -1573,14 +1573,14 @@ func TestEnforcedQuota(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	// Update the project to enforce quota.
-	if project.Annotations == nil {
-		project.Annotations = make(map[string]string)
-	}
-	project.Annotations[labels.ProjectEnforcedQuotaAnnotation] = "true"
-	if err := kclient.Update(ctx, project); err != nil {
-		t.Fatal(err)
-	}
+	// Annotate the project to enforce quota.
+	project = helper.WaitForObject(t, helper.Watcher(t, c), &corev1.NamespaceList{}, project, func(obj *corev1.Namespace) bool {
+		if project.Annotations == nil {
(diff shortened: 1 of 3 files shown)


tylerslaton:
Seeing weird behavior where the run time of the run integration test more than doubled in CI. Looking into that.

integration/run/run_test.go
```diff
@@ -1574,6 +1575,9 @@ func TestEnforcedQuota(t *testing.T) {
 	}
 
 	// Update the project to enforce quota.
+	if err := kclient.Get(ctx, types.NamespacedName{Name: project.Name}, project); err != nil {
```
njhale, line 1578: I think the update will still fail if a controller updates the project between this get and the following update call. With that in mind, I think we'll want to either:
1. switch to an apply
2. perform a get+update repeatedly until you're successful

The later I had to do for some e2e tests for events:
```go
	app = helper.WaitForObject(t, helper.Watcher(t, c), &apiv1.AppList{}, app, func(obj *apiv1.App) bool {
		obj.Spec.Stop = &stop
		return kc.Update(ctx, obj) == nil
	})
```

njhale (commented):
I just have one suggestion here. Other than that, LGTM.

integration/run/run_test.go
tylerslaton: Thanks! I went with the `Wait` approach. Its up now 👍🏻

njhale (approved):

StrongMonkey (approved):

Issue #1597, opened 2023-05-11
Not able to get logs for sidecar containers.


sangee2004:
acorn version v0.7.0-alpha1-43-ge7fb1587+e7fb1587

Steps to reproduce the problem:
Deploy app with sidecars.
Once app is deployed tried to access logs of sidecar containers.
This fails with following error:
```
acorn logs mytest.nginx1-5f665f7f45-mqc7v:sidecar 
  ✗  ERROR:  unable to parse requirement: values[0][acorn.io/public-name]: Invalid value: "mytest.nginx1-5f665f7f45-mqc7v:sidecar": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
```
But accessing logs using appname works.
```
% acorn containers mytest
NAME                                     APP       IMAGE     STATE     RESTARTCOUNT   CREATED   MESSAGE
mytest.nginx1-5f665f7f45-mqc7v           mytest    nginx     running   0              54s ago   
mytest.nginx1-5f665f7f45-mqc7v:sidecar   mytest    ubuntu    running   0              54s ago   
% acorn logs mytest
nginx1-5f665f7f45-mqc7v: /docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
nginx1-5f665f7f45-mqc7v: /docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
nginx1-5f665f7f45-mqc7v: /docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
nginx1-5f665f7f45-mqc7v: 10-listen-on-ipv6-by-default.sh: info: Getting the checksum of /etc/nginx/conf.d/default.conf
nginx1-5f665f7f45-mqc7v: 10-listen-on-ipv6-by-default.sh: info: Enabled listen on IPv6 in /etc/nginx/conf.d/default.conf
nginx1-5f665f7f45-mqc7v: /docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
nginx1-5f665f7f45-mqc7v: /docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh
[...]

PR #1673, opened 2023-05-30
fix: retrieve logs for sidecars (: notation in cli) (#1597)

iwilltry42:
Ref #1597 

## Note

It would be easily fixable with much less code, e.g. by just splitting the name input on `:`, but this feels more robust and is done like this in other places as well.

### Checklist

### DIFF
--- a/pkg/client/app.go
+++ b/pkg/client/app.go
@@ -223,13 +223,24 @@ func translatePermissions(err error) error {
 func (c *DefaultClient) AppLog(ctx context.Context, name string, opts *LogOptions) (<-chan apiv1.LogMessage, error) {
 	appName := name
 
+	var targetNamespace, targetName string
+
 	app, err := c.AppGet(ctx, appName)
 	if apierrors.IsNotFound(err) {
 		appName, _ = publicname.Split(name)
 		app, err = c.AppGet(ctx, appName)
 	}
-	if err != nil {
-		return nil, err
+	if err == nil {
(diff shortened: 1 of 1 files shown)


pkg/client/app.go
```diff
@@ -223,13 +223,24 @@ func translatePermissions(err error) error {
 func (c *DefaultClient) AppLog(ctx context.Context, name string, opts *LogOptions) (<-chan apiv1.LogMessage, error) {
 	appName := name
 
+	var targetNamespace, targetName string
+
 	app, err := c.AppGet(ctx, appName)
```
tylerslaton, line 228: This error isn't really being checked unless it is a not found error. Thoughts on if it should or shouldn't?

tylerslaton (approved):

g-linville (approved):

pkg/client/app.go
iwilltry42: I think we don't need to care too much about it.
If it's a real error, like not being able to contact the server, then subsequent requests like [ContainerReplicaGet](https://github.com/acorn-io/acorn/pull/1673/files/c5dd2cac8aa06789a287d83540e09327fdd87c8d#diff-2b548167cebba53ceca0681a297fd5471f930c043e4f6e50bda1eccaf78e89dcR236) will fail as well and bubble up the error. Other errors related to the input are handled properly by trying adjusted lookups.
Any objections?

PR #1674, opened 2023-05-30
Fix nightly EKS failures

tylerslaton:
Trying out a change that may help fix the tests. The weird thing is that the test works in GHA, but not in EKS. This particular line I am changing was hanging forever which caused timeouts in the EKS runs.

### DIFF
--- a/integration/run/run_test.go
+++ b/integration/run/run_test.go
@@ -1603,8 +1603,7 @@ func TestEnforcedQuota(t *testing.T) {
 	// Wait for the app to set the AppInstanceQuotaAllocated condition to be transitioning and for the namespace
 	// to be ready.
 	helper.WaitForObject(t, helper.Watcher(t, c), &apiv1.AppList{}, app, func(obj *apiv1.App) bool {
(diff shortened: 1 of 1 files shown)


integration/run/run_test.go
```diff
@@ -1603,8 +1603,7 @@ func TestEnforcedQuota(t *testing.T) {
 	// Wait for the app to set the AppInstanceQuotaAllocated condition to be transitioning and for the namespace
 	// to be ready.
 	helper.WaitForObject(t, helper.Watcher(t, c), &apiv1.AppList{}, app, func(obj *apiv1.App) bool {
-		return obj.Status.Condition(v1.AppInstanceConditionNamespace).Success &&
-			obj.Status.Condition(v1.AppInstanceConditionQuotaAllocated).Transitioning
```
njhale, line 1607: My theory is that both conditions are only true transiently, and not eventually; i.e. we can start watching _too late_ and missed when both are true.

njhale (approved):
Left my theory for the behavior you're observing.

tylerslaton:
Merging and am triggering a new run.

PR #1691, opened 2023-06-02
hack: add k8s auditing via acorn events

njhale:
and switch to metav1.MicroTime for events


njhale:
This was useful for debugging purposes, but I don't think we want to continue down this route. IMO, it feels like otel/tracing is the way to go, but we can always reopen this down the road if that's not the case.

Issue #1689, opened 2023-06-01
Services stuck in "pending" state


sangee2004:
acorn version v0.7.0-alpha1-125-g071f902a+071f902a

Steps to reproduce the problem:
Deploy the following service:

```
services: "my-service": {
    generated: job: "create-service"
}
jobs: "create-service": {
    image: "ubuntu:latest"
    entrypoint: "/app/create-service.sh"
    files: "/app/create-service.sh": """
      #!/bin/bash
      cat <<EOF > /run/secrets/output
      services: "my-service": {
        secrets: ["my-service-secret"]
      }
      secrets: "my-service-secret": {
          type: "basic"
          data: {
            username: "foo"
            password: "bar"
          }
      }
      EOF
      """
}
secrets: "my-service-secret": {
    type: "generated"
    params: job: "create-service"
}
```

App gets stuck in "pending create" state.
```
acorn apps
NAME        IMAGE          HEALTHY   UP-TO-DATE   CREATED    ENDPOINTS                                MESSAGE
mydoc       46cd926e088a   0         0            39s ago                                             [services: my-service: [pending create]]
acorn secrets
NAME                              TYPE      KEYS                  CREATED
mydoc.my-service-secret           basic     [password username]   82s ago
```

sangee2004:
Tested with acorn version v0.7.0-alpha1-148-g7909ed41+7909ed41.

Able to deploy the service mentioned in the issue successfully.

PR #1697, opened 2023-06-02
Correctly report status for Acorn services that don't have k8s services (#1689)

thedadams:
If an Acorn service didn't have an associated Kubernetes service, then the Acorn service would not be reported as ready. After this change, we track whether an Acorn service should have a Kubernetes service so that we don't take it into account when we report its status.

### Checklist

### DIFF
--- a/pkg/apis/internal.acorn.io/v1/service.go
+++ b/pkg/apis/internal.acorn.io/v1/service.go
@@ -29,6 +29,7 @@ type ServiceInstance struct {
 type ServiceInstanceStatus struct {
 	Conditions []Condition `json:"conditions,omitempty"`
 	Endpoints  []Endpoint  `json:"endpoints,omitempty"`
+	HasService bool        `json:"hasService,omitempty"`
 }
 
 func (in *ServiceInstance) ShortID() string {
--- a/pkg/controller/service/service.go
+++ b/pkg/controller/service/service.go
@@ -18,12 +18,14 @@ func RenderServices(req router.Request, resp router.Response) error {
 		return err
 	}
 	resp.Objects(objs...)
+	svcInstance.Status.HasService = len(objs) > 0
 
 	objs, err = publish.ServiceLoadBalancer(req, svcInstance)
 	if err != nil {
 		return err
 	}
 	resp.Objects(objs...)
+	svcInstance.Status.HasService = svcInstance.Status.HasService || len(objs) > 0
 
 	objs, err = publish.Ingress(req, svcInstance)
 	if err != nil {
@@ -31,5 +33,6 @@ func RenderServices(req router.Request, resp router.Response) error {
 	}
 	resp.Objects(objs...)
 
+	resp.Objects(svcInstance)
 	return nil
 }
--- a/pkg/controller/service/testdata/ingress/basic/expected.yaml.d/serviceinstance.yaml
+++ b/pkg/controller/service/testdata/ingress/basic/expected.yaml.d/serviceinstance.yaml
@@ -1,0 +1,48 @@
+kind: ServiceInstance
+apiVersion: internal.acorn.io/v1
+metadata:
+  name: oneimage
+  namespace: app-created-namespace
+  uid: 1234567890abcdef
+  labels:
+    "acorn.io/app-namespace": "app-namespace"
+    "acorn.io/app-name": "app-name"
+    "acorn.io/container-name": "oneimage"
+    "acorn.io/managed": "true"
+spec:
+  appName: app-name
+  appNamespace: app-namespace
+  publish:
+    - hostname: localhost
+    - hostname: ci1.acorn.not
+  labels:
+    "acorn.io/app-namespace": "app-namespace"
+    "acorn.io/app-name": "app-name"
+    "acorn.io/container-name": "oneimage"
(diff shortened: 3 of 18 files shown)


pkg/controller/service/service.go
```diff
@@ -18,18 +18,21 @@ func RenderServices(req router.Request, resp router.Response) error {
...
 	resp.Objects(objs...)
+	svcInstance.Status.HasService = svcInstance.Status.HasService || len(objs) > 0
 
 	objs, err = publish.Ingress(req, svcInstance)
 	if err != nil {
 		return err
 	}
 	resp.Objects(objs...)
 
+	resp.Objects(svcInstance)
```
StrongMonkey, line 36: This seems weird but do you want to put svcInstance into it's own `resp.Objects`? I think it should automatically update the status if you set it

pkg/controller/service/service.go
```diff
@@ -18,18 +18,21 @@ func RenderServices(req router.Request, resp router.Response) error {
...
 	}
 	resp.Objects(objs...)
+	svcInstance.Status.HasService = len(objs) > 0
 
 	objs, err = publish.ServiceLoadBalancer(req, svcInstance)
 	if err != nil {
 		return err
 	}
 	resp.Objects(objs...)
+	svcInstance.Status.HasService = svcInstance.Status.HasService || len(objs) > 0
```
StrongMonkey, line 28: Would setting `svcInstance.Status.HasService = len(objs) > 0` at the end of the handler enough?

pkg/controller/service/testdata/ingress/basic/expected.yaml.d/serviceinstance.yaml
```diff
@@ -0,0 +1,48 @@
...
+apiVersion: internal.acorn.io/v1
+metadata:
+  name: oneimage
+  namespace: app-created-namespace
+  uid: 1234567890abcdef
+  labels:
+    "acorn.io/app-namespace": "app-namespace"
+    "acorn.io/app-name": "app-name"
+    "acorn.io/container-name": "oneimage"
+    "acorn.io/managed": "true"
```
njhale, line 11: nit: I don't think we need the quotes here (and in the other fixtures as well).

pkg/controller/service/service.go
thedadams: You're correct that this isn't necessary for running Acorn. However, this allows the unit tests to ensure that fields are set correctly after the handler runs.

StrongMonkey: Hmm ok. It would be weird if we are just adding this for sake of unit test, can the test be asserting the original object instead of the object in resp.Objects?

pkg/controller/service/testdata/ingress/labels/expected.yaml.d/serviceinstance.yaml
```diff
@@ -0,0 +1,47 @@
...
+    "conl1": "value"
+    "conl3": "value"
+    "globall1": "value"
+    "globall2": "value"
+  annotations:
+    "allconsa1": "value"
+    "cona1": "value"
+    "cona3": "value"
+    "globala1": "value"
+    "globala2": "value"
```
njhale, line 29: nit: Do all of these tests require all of these fields set? Not sure it's ultra important for this PR in particular if they aren't, but it would be nice to prune these down to the important bits.

pkg/apis/internal.acorn.io/v1/service.go
```diff
@@ -29,6 +29,7 @@ type ServiceInstance struct {
 type ServiceInstanceStatus struct {
 	Conditions []Condition `json:"conditions,omitempty"`
 	Endpoints  []Endpoint  `json:"endpoints,omitempty"`
+	HasService bool        `json:"hasService,omitempty"`
```
njhale, line 32: nit: it's not conventional here, but if it's not too burdensome, it would be nice to start adding descriptive godocs comments for net-new stuff

I'm assuming this propagates to the CRD's openapi spec and k8s API discovery as well (though I'm not sure if wrangler handle's that in it's CRD generation)

njhale (approved):
LGTM

I left a few comments, but they're nits/optionals/questions.

pkg/controller/service/service.go
njhale: hmm, at a glance, I _think_ you'd get a false negative if the first `objs` returns `len(objs) > 0` but the second does not.

thedadams: No because the second `objs, err =` overwrites the first. So, if the `objs` produces a service, then the next call could overwrite it to nothing.

StrongMonkey: Oh, make sense! For some reason I thought objs are additive and was alwasy appended

pkg/controller/service/service.go
thedadams: That wouldn't be possible with the testing framework we use for these handlers.

Functionally, it doesn't change anything: `baaah` will pull it out and update its status no matter what. This is similar to the way `DeploySpec` works. The `AppInstance` gets add to `resp.Objects` when we set the conditon.

StrongMonkey (approved):

Issue #1694, opened 2023-06-02
Improve `Event` timestamp precision


njhale:
## Problem

The `observed` field of an `Event` is of type `metav1.Time`, which is truncated to seconds precision when serialized. This limitation means that we lose the ability to order events that occur within the same second; e.g. resource contention between controllers, flapping, etc.

## Solution

Use `metav1.MicroTime` instead of `metav1.Time` in `Events`.

njhale:
Fixed by https://github.com/acorn-io/acorn/pull/1701

sangee2004:
acorn version v0.7.0-alpha1-138-gab3dab9e+ab3dab9e

Deploy app.
acorn events
  ✗  ERROR:  template: :1:55: executing "" at <.Observed>: wrong type for value; expected v1.Time; got v1.MicroTime


I get these error in some projects (possibly they have older events ?) 
```
 acorn events
ERRO[0001] Error streaming events: parsing time "2023-06-02T23:58:30Z" as "2006-01-02T15:04:05.000000Z07:00": cannot parse "Z" as ".000000" 
SOURCE    NAME      TYPE      ACTOR     OBSERVED   DESCRIPTION
```

njhale:
@sangee2004 thanks for catching these!

> ✗ ERROR: template: :1:55: executing "" at <.Observed>: wrong type for value; expected v1.Time; got v1.MicroTime

Ah, looks like a few changes that were supposed to go into #1701 got left out by mistake. I've put them in https://github.com/acorn-io/acorn/pull/1703, which fixes the issue locally for me.

> I get these error in some projects (possibly they have older events ?)

I believe this is because there's a skew between the openapi schema of the EventInstance CRD in the `account-api-server` and that of the provisioned clusters. This should be resolved as soon as https://github.com/acorn-io/hub/pull/553 merges.

I'll run some tests really quickly to see if I can verify _before_ merging.

njhale:
It looks like https://github.com/acorn-io/acorn/pull/1701 was a backwards incompatible schema change and breaks Event list when there are EventInstance CRs on the cluster that predate the change.

Added a commit to #1703 to address that as well.

sangee2004:
Tested with  acorn version v0.7.0-alpha1-148-g7909ed41+7909ed41. Errors reported in
https://github.com/acorn-io/acorn/issues/1694#issuecomment-1574458380 is not seen anymore.


[...]

PR #1701, opened 2023-06-02
Give event timestamps microsecond precision (#1694)

njhale:
The `metav1.Time` type used as timestamps in acorn events were
being truncated to seconds precision when serialized. This
limitation meant that we didn't have the ability to order events
that occur within the same second; e.g. resource contention between
controllers, flapping, etc.

To this end, switch to `metadata.MicroTime`.

### DIFF
--- a/pkg/apis/internal.acorn.io/v1/event.go
+++ b/pkg/apis/internal.acorn.io/v1/event.go
@@ -40,7 +40,7 @@ type EventInstance struct {
 	Description string `json:"description,omitempty"`
 
 	// Observed represents the time the Event was first observed.
-	Observed metav1.Time `json:"observed"`
+	Observed metav1.MicroTime `json:"observed" wrangler:"type=string"`
 
 	// Details provides additional information about the cluster at the time the Event occurred.
 	//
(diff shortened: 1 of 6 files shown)


pkg/apis/internal.acorn.io/v1/event.go
```diff
@@ -40,7 +40,7 @@ type EventInstance struct {
 	Description string `json:"description,omitempty"`
 
 	// Observed represents the time the Event was first observed.
-	Observed metav1.Time `json:"observed"`
+	Observed metav1.MicroTime `json:"observed" wrangler:"type=string"`
```
g-linville, line 43: Just curious, why did you need to add this `wrangler` tag to this struct field?

g-linville (approved):

pkg/apis/internal.acorn.io/v1/event.go
njhale: We use `wrangler` to generate the internal CRDs at runtime, and without this, it generates an incorrect openapi spec for `MicroTime` that breaks serialization and conversion; general chaos.

thedadams (approved):

PR #1705, opened 2023-06-03
Revert "fix: resolvelocal should always return full ID if image is found (#1682)"

cjellick:
This reverts commit e74aea6c635ec181abb6c9d05df5c7a4432e7b00.

### DIFF
--- a/pkg/tags/tags.go
+++ b/pkg/tags/tags.go
@@ -86,6 +86,9 @@ func ResolveLocal(ctx context.Context, c kclient.Client, namespace, image string
 		}
 	} else if err != nil {
 		return "", false, err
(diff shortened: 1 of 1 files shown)


njhale (approved):
This change fixes the issue locally for me.

cjellick:
https://github.com/acorn-io/acorn/pull/1682

pkg/tags/tags.go
```diff
@@ -86,6 +86,9 @@ func ResolveLocal(ctx context.Context, c kclient.Client, namespace, image string
 		}
```
iwilltry42, line 86: Adding a `return image, false, nil` below this line also fixes this issue.

PR #1707, opened 2023-06-05
Ensure acme-challenge is available

cjellick:
To request a cert from let's encrypt for a custom domain, we have to
setup an acme HTTP01 challenge. This entails making a specific path
available with a specific value on the specified domain. To do so, we
setup an http server (that knows how to handle the acme protocol), a
service, and an ingress.

This addresses an issue where we were requesting the cert before the
server/servie/ingress combination was fully functional. It does so by
setting up a test path in the server and querying for a 200 response
before attempting to request the cert.

### DIFF
--- a/pkg/controller/tls/letsencrypt.go
+++ b/pkg/controller/tls/letsencrypt.go
@@ -14,6 +14,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"net/http"
 	"strconv"
 	"strings"
 	"sync"
@@ -38,12 +39,14 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/wait"
 	kclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
 	LetsEncryptURLStaging    = "https://acme-staging-v02.api.letsencrypt.org/directory"
 	LetsEncryptURLProduction = "https://acme-v02.api.letsencrypt.org/directory"
+	acmeChallengePath        = "/.well-known/acme-challenge/"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


pkg/controller/tls/letsencrypt.go
```diff
@@ -560,6 +563,44 @@ func (u *LEUser) httpChallenge(ctx context.Context, domain string) (*certificate
...
+	// The provider will only present a single token at a time. Cleanup before attempting to obtain cert.
+	// Need to do this before potentially returning from ExponentialBackoffWithContext's error
+	cleanupErr := httpProviderServer.CleanUp(domain, "test", "test")
+	if cleanupErr != nil {
+		logrus.Errorf("Failed to cleanup test acme-challenge. %v", cleanupErr)
+	}
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to successfully hit test acme-challenge endpoint. %w", err)
+	}
```
njhale, line 602: nit: I think factoring this out to helper function would clean things up a bit:

```go
func ensureChallengeEndpoint(ctx context.Context, provider *http01.ProviderServer, domain string) error {
    if err := provider.Present(domain, "test", "test"); err != nil {
        return nil, fmt.Errorf("failed to present test acme-challenge. %w", err)
    }
    
    defer func() {
        // The provider will only present a single token at a time. Cleanup before attempting to obtain cert.
        // Need to do this before potentially returning from ExponentialBackoffWithContext's error
        if err := httpProviderServer.CleanUp(domain, "test", "test"); err != nil {
            logrus.Errorf("Failed to cleanup test acme-challenge. %v", err)     
        }
    }()
    
    url := "http://" + domain + acmeChallengePath + "test"
	if err := wait.ExponentialBackoffWithContext(ctx, wait.Backoff{
        Duration: 1 * time.Second,
		Factor:   2,
		Steps:    10,
		Cap:      180 * time.Second,
	}, func(ctx context.Context) (bool, error) {
		logrus.Debugf("Attempting to hit test cert path %v", url)
		resp, err := http.Get(url)
		if err != nil {
			logrus.Errorf("Error hitting test acme-challenge endpoint: %v", err)
			return false, nil
		}
		logrus.Debugf("Response code for test acme-challenge endpoint: %v", resp.StatusCode)
		if resp.StatusCode == 200 {
			return true, nil
		}
		return false, nil
	}); err != nil {
        return fmt.Errorf("failed to successfully hit test acme-challenge endpoint. %w", err)
	}
	
	return nil
}
```

njhale (approved):
This makes sense to me. I left one style nit, but I don't think it should hold this up.

pkg/controller/tls/letsencrypt.go
cjellick: i did the refactor

iwilltry42 (approved):
I don't like that this is necessary, as I'd expect that networking is OK once the Ingress object is updated, but if it makes the thing more stable, then I'm all for it.
LGTM

Issue #1710, opened 2023-06-05
Give a useful error message when a user tries to `acorn update` a nested Acorn or service


g-linville:
Related to #1596, #1600, and #1690.

For now, we want to prevent users from using the `acorn update` command to update the Acorn images for nested Acorns and services inside of an Acorn app. Instead, they should update the images referenced within the top-level (parent) app, and then update the top-level app itself.

Users are currently presented with an error like this if they try to update a nested Acorn or service:

```
❯ acorn update --image a5eeb79495bc9dc5701403692730b942e2ac3f43f73b059b3157a70a4404a8c0 consumer.my-service 
  ✗  ERROR:  App.api.acorn.io "consumer.my-service" is invalid: metadata.name: Invalid value: "consumer.my-service": a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')
```

Instead of a name format error, we should detect that they are trying to do an invalid update and give a more useful error message.

sangee2004:
Tested with acorn version v0.7.0-alpha1-152-g1706098f+1706098f

User is presented with following error message when attempting to update a nested acorn app:
```
acorn apps
NAME                 IMAGE                                       HEALTHY   UP-TO-DATE   CREATED     ENDPOINTS                                                                            MESSAGE
mynested             c3d7e415d145                                0         0            5m37s ago                                                                                        OK
mynested.my-acorn1   1a6c64d2ccd0                                1         1            5m37s ago   http://webapp-mynested-my-acorn1-e1848a0c-4eac4b33.local.oss-acorn.io => webapp:80   OK
mynested.my-acorn2   b59c29933ae5                                1         1            5m37s ago   http://myweb1-mynested-my-acorn2-e5efce26-88f3c3c5.local.oss-acorn.io => myweb1:80   OK
```

```
 acorn update --image <image>:v0.0.2 mynested.my-acorn1  
 ✗  ERROR:  App.api.acorn.io "mynested.my-acorn1" is invalid: metadata.name: Invalid value: "mynested.my-acorn1": invalid name
               To update a nested Acorn or a service, update the parent Acorn instead.
```

sangee2004:
Wanted to confirm if any of these issues were also meant to get addressed as part of this issue?

1. Error message in case of `--pull` is not a useful error message.
```
[...]

PR #1711, opened 2023-06-05
Return a more useful error if the user tries to update a dependent acorn (#1710)

g-linville:
for #1710

Two things I'm not sure of:

1. I used the client to check to see whether the name given by the user references a real dependency. To save execution time, would we rather just assume that any name with a `.` in it references a dependency, and send this specific error message every time?
2. Is my error message clear and concise enough?

Would like feedback on the above.

### Checklist

### DIFF
--- a/pkg/cli/update.go
+++ b/pkg/cli/update.go
@@ -1,8 +1,10 @@
 package cli
 
 import (
+	"context"
 	"fmt"
 	"io"
+	"strings"
 
 	cli "github.com/acorn-io/acorn/pkg/cli/builder"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


cjellick (changes requested):
I envisioned this being a server side validation. Is that possible or is there some technical limitation?

g-linville:
> I envisioned this being a server side validation. Is that possible or is there some technical limitation?

@cjellick the current error is being generated by the CLI when it tries to parse the name. So unless we want to modify that behavior, this also has to be done client-side.

g-linville:
> > I envisioned this being a server side validation. Is that possible or is there some technical limitation?
> 
> @cjellick the current error is being generated by the CLI when it tries to parse the name. So unless we want to modify that behavior, this also has to be done client-side.

Actually...looking at the code again, I'm no longer as sure about this. I'm gonna go try some stuff and I'll report back.

g-linville:
Moved the logic to ValidateUpdate.

cjellick (commented):
Since @ibuildthecloud is back, i'm waiting to hear what he thinks about upgrade/update nested acorns

thedadams (approved):

cjellick (approved):

Issue #1502, opened 2023-04-15
2 entries for the same `repository:tag` but with different digest


sangee2004:
acorn version v0.6.0-170-gc3906ab1+c3906ab1

Steps to reproduce the problem:

1.Deploy an app with image any image - in my case `index.docker.io/sangeetha/myfirstacorn:v0.0.1`
2. Built an image with the same tag as above - `acorn build -t index.docker.io/sangeetha/myfirstacorn:v0.0.1 .`

This results in 2 image entries being created.

```
% acorn images
REPOSITORY                                            TAG       IMAGE-ID
index.docker.io/sangeetha/myfirstacorn                v0.0.1    cb54602c8b8e
index.docker.io/sangeetha/myfirstacorn                v0.0.1    ee8a243e3e7a
```

iwilltry42:
Confirmed.
```bash
$ acorn run index.docker.io/sangeetha/myfirstacorn:v0.0.1
# ...

$ acorn images                                           
REPOSITORY                               TAG       IMAGE-ID
index.docker.io/sangeetha/myfirstacorn   v0.0.1    ec318908dbc4

$ acorn build -t index.docker.io/sangeetha/myfirstacorn:v0.0.1 .
# ...
e3d5974dea7e979ddea7b5e3b9f08675bd661a883d01dab250d7e10096c20a14

$ acorn images
REPOSITORY                               TAG       IMAGE-ID
index.docker.io/sangeetha/myfirstacorn   v0.0.1    e3d5974dea7e
index.docker.io/sangeetha/myfirstacorn   v0.0.1    ec318908dbc4

$ kubectl describe images   
Name:         e3d5974dea7e979ddea7b5e3b9f08675bd661a883d01dab250d7e10096c20a14
Namespace:    acorn
Labels:       <none>
Annotations:  apply.acorn.io/applied:
                H4sIAAAAAAAA/5zNu62EMBBG4V4mJrABP+u4Dfz2DHeRsEHYGyF6X6HtYLMTfeciXv+ldYrUXhiNjTKxCW5mgZPgAj+RjEwpLMpbZxJbq+H9xEoz0mgUO9FKhRl2sd4FGqhIB6OD4k...
API Version:  api.acorn.io/v1
Digest:       sha256:e3d5974dea7e979ddea7b5e3b9f08675bd661a883d01dab250d7e10096c20a14
Kind:         Image
Metadata:
  Creation Timestamp:  2023-05-15T13:47:00Z
[...]

PR #1714, opened 2023-06-05
Stop processing tags on app creation (#1502)

thedadams:
Based on which image was pulled when setting an image for an app, the handler would append tags to the current image and (attempt to) remove tags from other images. This causes confusion in the following scenario:
1. Create an app from an image tag.
2. Build an image with the same tag as above.
3. Depending on the tag, you may have two images with the same tag or the image created in step one would remain tagged and the built image would not.

The outcomes in step 3 are confusing. Therefore, after these changes, the app handler will not try to "take" a tag from an existing image. The idea is that if the desired tag exists when the app is run, then we would have used that image. If the image that the app is using doesn't  have the tag, then a new image must have been built and tagged, which "took over" the tag in question.

### Checklist

### DIFF
--- a/pkg/controller/images/images.go
+++ b/pkg/controller/images/images.go
@@ -13,7 +13,7 @@ import (
 	kclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func CreateImages(req router.Request, resp router.Response) error {
+func CreateImages(req router.Request, _ router.Response) error {
 	app := req.Object.(*v1.AppInstance)
 	if app.Status.AppImage.ID == "" || app.Status.AppImage.Digest == "" {
 		return nil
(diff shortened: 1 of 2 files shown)


thedadams:
I found a problem with this implementation.

pkg/controller/images/images.go
```diff
@@ -103,14 +103,15 @@ func createImageForSelf(req router.Request, app *v1.AppInstance) (*v1.ImageInsta
...
 		}
-	}
 
-	// remove tag from existing images
-	if err := removeTag(req, app, imageName); err != nil {
-		return nil, err
-	}
+		// If a new image is built with the same tag as the image used for this app,
+		// then we don't want to take the tag from the newly built image.
+		if found, err := checkExistingImages(req, app, imageName); err != nil {
```
StrongMonkey, line 109: Nit: imageName seems confusing as it is just a tag in the method input?

pkg/controller/images/images.go
```diff
@@ -139,35 +140,22 @@ func createImageForSelf(req router.Request, app *v1.AppInstance) (*v1.ImageInsta
...
+		return false, err
 	}
 
 	for _, image := range images.Items {
-		if len(image.Tags) == 0 {
-			continue
-		}
-
-		newTags := make([]string, 0, len(image.Tags))
 		for _, existingTag := range image.Tags {
```
StrongMonkey, line 153: I am sure you tested, but how does this logic compare the repo? Is that listing all images and only compares on tags?

pkg/controller/images/images.go
thedadams: Tags for Acorn `ImageInstances` is the whole image name. For example, `index.docker.io/thedadams/my-image:v1.0.0` is the Acorn tag.

pkg/controller/images/images.go
thedadams: I think my response to your other question also addresses this.

StrongMonkey (approved):
🔥

cjellick:
attaching issue https://github.com/acorn-io/acorn/issues/1502

g-linville (approved):

Issue #1563, opened 2023-04-28
Not able to deploy apps using remote image entries 


sangee2004:
acorn version - acorn version v0.7.0-alpha1-16-g9007dba3+9007dba3

Steps to reproduce the problem:

Deploy app using image from dockerhub.
Once the app is deployed successfully, we will have an image entry created for it.
Try to deploy app using this image.
App deployment fails with ``` [image-pull: GET https://index.docker.io/v2/library/6e60b20b1d34/manifests/latest: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:library/6e60b20b1d34 Type:repository]]]```

```
acorn run -n mytestr14a `<docker-account>/<repo>:<tag>`
acorn images  `<docker-account>/<repo>:<tag> `  
REPOSITORY               TAG       IMAGE-ID
<docker-account>/<repo>  <tag>   6e60b20b1d34
% acorn run -n mytestr14im 6e60b20b1d34 
```

iwilltry42:
FWIW, I figured that this only happens when addressing the image by it's shortened ID.
If you reference it by the full name, it works:

```
$ acorn run -n issue1563-1 ghcr.io/acorn-io/library/hello-world:latest
issue1563-1
STATUS: ENDPOINTS[] HEALTHY[] UPTODATE[] 
STATUS: ENDPOINTS[] HEALTHY[0] UPTODATE[0] [containers: webapp: [pending create]]
STATUS: ENDPOINTS[http://webapp-issue1563-1-5ab9c939.local.oss-acorn.io => webapp:80] HEALTHY[0/1] UPTODATE[1] [containers: webapp: [webapp ContainerCreating]]
┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
| STATUS: ENDPOINTS[http://webapp-issue1563-1-5ab9c939.local.oss-acorn.io => webapp:80] HEALTHY[1] UPTODATE[1] OK |
└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘

┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
| STATUS: ENDPOINTS[http://webapp-issue1563-1-5ab9c939.local.oss-acorn.io => webapp:80] HEALTHY[1] UPTODATE[1] OK |
└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘

$ acorn images                                                        
REPOSITORY                             TAG       IMAGE-ID
[...]

Issue #1704, opened 2023-06-03
Not able to deploy apps from images.


sangee2004:
acorn version v0.7.0-alpha1-141-g5197ed75+5197ed75

Steps to reproduce the problem:
```
acorn run -n mytestnew2 sangeetha/myfirstacorn:v0.0.1
mytestnew2
STATUS: ENDPOINTS[] HEALTHY[] UPTODATE[] 
STATUS: ENDPOINTS[] HEALTHY[0] UPTODATE[0] [image-pull: could not parse reference: ]
^C%                                                          
```
```
 acorn run -n mytest1 ghcr.io/acorn-io/library/hello-world
mytest1
STATUS: ENDPOINTS[] HEALTHY[0] UPTODATE[0] [image-pull: could not parse reference: ]
^C%                                                              
```

njhale:
Looks like @cjellick got this! Unassigned myself

PR #1718, opened 2023-06-06
fix: resolvelocal should always return full ID if image is found (#1563)

iwilltry42:
Ref #1563

### Checklist

### DIFF
--- a/integration/client/apps/apps_test.go
+++ b/integration/client/apps/apps_test.go
@@ -1,6 +1,7 @@
 package apps
 
 import (
+	"fmt"
 	"strings"
 	"testing"
 
@@ -460,4 +461,38 @@ func TestAppRun(t *testing.T) {
 	assert.Equal(t, "volume", app.Spec.Volumes[0].Volume)
 	assert.Equal(t, "secret", app.Spec.Secrets[0].Secret)
 	assert.Equal(t, "value", app.Spec.DeployArgs["key"])
+}
+
+func TestAppRunImageVariations(t *testing.T) {
+	helper.EnsureCRDs(t)
+	restConfig := helper.StartAPI(t)
+
+	ctx := helper.GetCTX(t)
+	kclient := helper.MustReturn(kclient.Default)
+	ns := helper.TempNamespace(t, kclient)
+
+	imageID := client2.NewImage(t, ns.Name)
+
+	c, err := client.New(restConfig, "", ns.Name)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if err := c.ImageTag(ctx, imageID, "foo/bar:baz"); err != nil {
+		t.Fatal(err)
+	}
+
+	imageNames := []string{
+		"foo/bar:baz",
+		imageID,
+		fmt.Sprintf("sha256:%s", imageID),
+		imageID[:8],
+	}
+
+	for _, imageName := range imageNames {
+		_, err := c.AppRun(ctx, imageName, nil)
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
 }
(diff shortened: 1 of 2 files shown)


cjellick:
I'm attaching this issue: https://github.com/acorn-io/acorn/issues/1704

The cause of that issue was the original fix to this that was reverted. So, this issue should get sent to QA when this new PR lands.

pkg/tags/tags.go
```diff
@@ -84,6 +84,7 @@ func ResolveLocal(ctx context.Context, c kclient.Client, namespace, image string
 		if IsLocalReference(image) {
 			return "", false, err
 		}
+		return image, false, nil
```
cjellick, line 87: @iwilltry42 is this accurate? you just want This single line change? This does not align with what was in your original change: https://github.com/acorn-io/acorn/pull/1682/files

thedadams: As I understand it, this is correct.

If the image is not found and not a local reference, then we essentially want to return "this image is not found and that's OK."

By not returning here, we'll go to the `else if !localImage.Remote` branch and essentially return `image, true, nil`.

thedadams (dismissed):

pkg/tags/tags.go
g-linville: But if the first condition is true, won't it just jump to the end of the whole `if else if` chain and return `image, false, nil` anyway?

thedadams: Oh, geez. You're 100% correct. I had my nesting confused.

g-linville: Yeah I think that's correct too

iwilltry42: Huh, that's weird.
It looked different in the diff I first viewed.
It's only this one line but added to my previous change actually.
I just checked it again (on my phone right now) due to your comments and here I don't see the previous changes.
I'll fix it tomorrow morning.

iwilltry42: I guess it dropped my cherry-pick of the previous change.
It's a Git mystery.

iwilltry42: Replayed the revert of @cjellick's revert and rebased :heavy_check_mark:

g-linville (approved):

integration/client/apps/apps_test.go
```diff
@@ -461,3 +462,37 @@ func TestAppRun(t *testing.T) {
...
+		imageID[:8],
+	}
+
+	for _, imageName := range imageNames {
+		_, err := c.AppRun(ctx, imageName, nil)
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+}
```
njhale, line 498: nit: I wonder if we could use sub tests to parallelize these test cases

```suggestion
func TestAppRunImageVariations(t *testing.T) {
	helper.EnsureCRDs(t)
	restConfig := helper.StartAPI(t)

	ctx := helper.GetCTX(t)
	kclient := helper.MustReturn(kclient.Default)
	ns := helper.TempNamespace(t, kclient)

	imageID := client2.NewImage(t, ns.Name)

	c, err := client.New(restConfig, "", ns.Name)
	if err != nil {
		t.Fatal(err)
	}

	if err := c.ImageTag(ctx, imageID, "foo/bar:baz"); err != nil {
		t.Fatal(err)
	}

    for desc, imageName := range map[string]string{
        "ref": "foo/bar:baz",
        "id": imageID,
        "sha256": fmt.Sprintf("sha256:%s", imageID),
        "short": imageID[:8],
    } {
        t.Run(desc, func(t *testing.T) {
             t.Parallel()
             _, err := c.AppRun(ctx, imageName, nil)
             assert.NoError(t, err)
        }
    }
}
```

njhale (approved):
suggestion on the tests, but totally not pressing

lgtm

integration/client/apps/apps_test.go
njhale: https://github.com/acorn-io/acorn/pull/1731

Issue #1708, opened 2023-06-05
Event TTL


njhale:
## Description

Introduce a TTL (Time To Live) for events and ensure that any event existing for a _longer_ period of time than that TTL is deleted.

## Motivation

Today, acorn events "live forever" on a cluster, this causes two problems:
- users will see events even when they're well past their useful life time; i.e. the related resources can be long gone
- there's no upper bound on the number of events so long-running clusters have the potential to be inundated with events; eventually consuming lots of resources which causes its own slew of deleterious effects

sangee2004:
Tested with `acorn version v0.7.0-alpha1-163-g7882f4c1+7882f4c1`

Able set TTL using `acorn install --event-ttl 20m`


Events get deleted as per the TTL setting.

```
acorn events 
SOURCE                             NAME                               TYPE                  ACTOR          OBSERVED   DESCRIPTION
app/mytestservice-mysec-3a60f938   6fd69c1cf85166cf76ed898f7412c014   AppImagePullSuccess   acorn-system   11m ago    Pulled index.docker.io/sangeetha/myfirstservice:v0.0.3
app/mytestservice-mysec-3a60f938   142e48e94c4501c06759e87d0f2a5ffc   AppImagePullSuccess   acorn-system   17m ago    Pulled index.docker.io/sangeetha/myfirstservice:v0.0.2
app/mytestservice                  60c237520f1cf423d11b34635a1490bd   AppImagePullSuccess   acorn-system   17m ago    Pulled a4972fa5903fc463526f877158060c4bf6ab457c3d6ecb1904c7551c760fa242
app/mytestservice                  bc64a9614e57ecb5f0be71f28b8e7891   AppCreate             system:admin   17m ago    App acorn/mytestservice created
app/mytestnew1                     a5c89c1ef99429d36f3ed93e78c80a41   AppImagePullSuccess   acorn-system   19m ago    Pulled sangeetha/myfirstacorn:v0.0.2
app/mytestnew1                     b6bd86a1407fb8f6cb93dd9f12f7eea6   AppCreate             system:admin   19m ago    App acorn/mytestnew1 created
app/mytestapp1                     23f5ec9057646f919557df9a251a8b8e   AppImagePullSuccess   acorn-system   22m ago    Pulled 9cb84f3628cc
app/mytestapp1                     7086f48542647a71a023e4037d74a727   AppImagePullSuccess   acorn-system   25m ago    Pulled 9cb84f3628cc
[...]

PR #1719, opened 2023-06-06
Add TTL for Events (#1708)

njhale:
- Introduce a TTL (Time To Live) for events and ensure that any event
existing for a longer period of time than that TTL is deleted.
- Expose an EventTTL config field
- Default TTL to 168 hours (7 days)

Vanquishes #1708 

### DIFF
--- a/docs/docs/100-reference/01-command-line/acorn_install.md
+++ b/docs/docs/100-reference/01-command-line/acorn_install.md
@@ -31,6 +31,7 @@ acorn install
       --cluster-domain strings                          The externally addressable cluster domain (default .oss-acorn.io)
       --controller-replicas int                         acorn-controller deployment replica count
       --controller-service-account-annotation strings   annotation to apply to the acorn-system service account
+      --event-ttl string                                Amount of time an Acorn event will be stored before being deleted (default '168h' - 7 days)
       --features strings                                Enable or disable features. (example foo=true,bar=false)
   -h, --help                                            help for install
       --http-endpoint-pattern string                    Go template for formatting application http endpoints. Valid variables to use are: App, Container, Namespace, Hash and ClusterDomain. (default pattern is {{hashConcat 8 .Container .App .Namespace | truncate}}.{{.ClusterDomain}})
--- a/pkg/apis/api.acorn.io/v1/types.go
+++ b/pkg/apis/api.acorn.io/v1/types.go
@@ -427,6 +427,7 @@ type Config struct {
 	AllowTrafficFromNamespace      []string        `json:"allowTrafficFromNamespace" name:"allow-traffic-from-namespace" usage:"Namespaces that are allowed to send network traffic to all Acorn apps"`
 	ServiceLBAnnotations           []string        `json:"serviceLBAnnotations" name:"service-lb-annotation" usage:"Annotation to add to the service of type LoadBalancer. Defaults to empty. (example key=value)"`
 	AWSIdentityProviderARN         *string         `json:"awsIdentityProviderArn" name:"aws-identity-provider-arn" usage:"ARN of cluster's OpenID Connect provider registered in AWS"`
+	EventTTL                       *string         `json:"eventTTL" name:"event-ttl" usage:"Amount of time an Acorn event will be stored before being deleted (default '168h' - 7 days)"`
 	Features                       map[string]bool `json:"features" name:"features" boolmap:"true" usage:"Enable or disable features. (example foo=true,bar=false)"`
 }
 
--- a/pkg/controller/eventinstance/eventinstance.go
+++ b/pkg/controller/eventinstance/eventinstance.go
@@ -1,0 +1,91 @@
+package eventinstance
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	v1 "github.com/acorn-io/acorn/pkg/apis/internal.acorn.io/v1"
+	"github.com/acorn-io/acorn/pkg/config"
+	"github.com/acorn-io/baaah/pkg/router"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	kclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const defaultTTL = 168 * time.Hour // 7 days
+
+func GCExpired() router.HandlerFunc {
+	// last stores pre-parsed TTLs from the configuration.
+	last := new(struct {
+		mu  sync.Mutex
+		raw string
+		ttl time.Duration
+	})
+
+	return handler{
+		getTTL: func(ctx context.Context, getter kclient.Reader) (time.Duration, error) {
+			cfg, err := config.Get(ctx, getter)
+			if err != nil {
+				return 0, err
+			}
+
+			cfgTTL := cfg.EventTTL
+			if cfgTTL == nil || *cfgTTL == "" {
+				return defaultTTL, nil
+			}
+
+			last.mu.Lock()
+			defer last.mu.Unlock()
+
+			if last.raw != *cfgTTL {
+				// This is a new TTL, parse and memoize
+				ttl, err := time.ParseDuration(*cfgTTL)
+				if err != nil {
+					return 0, err
+				}
+
+				last.raw, last.ttl = *cfgTTL, ttl
+			}
+
+			return last.ttl, nil
+		},
+	}.gcExpired
+}
+
+// ttlFunc is a function that returns the TTL to use for event expiration.
+type ttlFunc func(context.Context, kclient.Reader) (time.Duration, error)
+
+type handler struct {
+	getTTL ttlFunc
+}
+
+func (h handler) gcExpired(req router.Request, resp router.Response) error {
+	e := req.Object.(*v1.EventInstance)
+
+	// Get the currently configured TTL
+	ttl, err := h.getTTL(req.Ctx, req.Client)
+	if err != nil {
+		return fmt.Errorf("failed to get event ttl: %w", err)
+	}
+
+	// Check expiration
+	if now, expiration := time.Now(), e.Observed.Add(ttl); now.Before(expiration) {
+		// Still fresh, wait until expiration
+		resp.RetryAfter(time.Until(expiration))
+		return nil
+	}
+
+	// Expired, delete the event
+	if err := req.Client.Delete(req.Ctx, e, kclient.Preconditions{
+		// Adding these preconditions prevents us from deleting an event based on old information.
+		// e.g. The observed time has been updated and the event is no longer expired.
+		UID:             &e.UID,
+		ResourceVersion: &e.ResourceVersion,
+	}); err != nil && !apierrors.IsNotFound(err) {
+		// Assume any error other than not found is transient, return error to requeue w/ backoff
+		return err
+	}
+
+	return nil
+}
--- a/pkg/controller/eventinstance/eventinstance.go
+++ b/pkg/controller/eventinstance/eventinstance.go
@@ -3,25 +3,26 @@ package eventinstance
 import (
 	"context"
 	"fmt"
-	"sync"
 	"time"
 
 	v1 "github.com/acorn-io/acorn/pkg/apis/internal.acorn.io/v1"
 	"github.com/acorn-io/acorn/pkg/config"
 	"github.com/acorn-io/baaah/pkg/router"
+	"golang.org/x/sync/semaphore"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	kclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-const defaultTTL = 168 * time.Hour // 7 days
+const defaultTTL = 7 * 24 * time.Hour
 
 func GCExpired() router.HandlerFunc {
-	// last stores pre-parsed TTLs from the configuration.
-	last := new(struct {
-		mu  sync.Mutex
+	sem := semaphore.NewWeighted(1)
+
+	// parsed stores pre-parsed TTLs from the configuration.
+	var parsed struct {
 		raw string
 		ttl time.Duration
-	})
+	}
 
 	return handler{
 		getTTL: func(ctx context.Context, getter kclient.Reader) (time.Duration, error) {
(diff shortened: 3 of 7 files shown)


pkg/apis/api.acorn.io/v1/types.go
```diff
@@ -427,6 +427,7 @@ type Config struct {
 	AllowTrafficFromNamespace      []string        `json:"allowTrafficFromNamespace" name:"allow-traffic-from-namespace" usage:"Namespaces that are allowed to send network traffic to all Acorn apps"`
 	ServiceLBAnnotations           []string        `json:"serviceLBAnnotations" name:"service-lb-annotation" usage:"Annotation to add to the service of type LoadBalancer. Defaults to empty. (example key=value)"`
 	AWSIdentityProviderARN         *string         `json:"awsIdentityProviderArn" name:"aws-identity-provider-arn" usage:"ARN of cluster's OpenID Connect provider registered in AWS"`
+	EventTTL                       *string         `json:"eventTTL" name:"event-ttl" usage:"Amount of time an Acorn event will be stored before being deleted (default '24h' - 24 hours)"`
```
njhale, line 430: I chose to follow the convention and typed this as `*string`, but IMO it would be nice if we could use `*metav1.Duration`.

pkg/controller/eventinstance/eventinstance.go
```diff
@@ -0,0 +1,91 @@
...
+	"time"
+
+	v1 "github.com/acorn-io/acorn/pkg/apis/internal.acorn.io/v1"
+	"github.com/acorn-io/acorn/pkg/config"
+	"github.com/acorn-io/baaah/pkg/router"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	kclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const defaultTTL = 24 * time.Hour
```
njhale, line 16: I could use an opinion on this; personally, I feel like 48 is a little better.

pkg/controller/eventinstance/eventinstance.go
```diff
@@ -0,0 +1,91 @@
...
+			}
+
+			last.Lock()
+			defer last.Unlock()
+
+			if last.raw != *cfgTTL {
+				// This is a new TTL, parse and memoize
+				ttl, err := time.ParseDuration(*cfgTTL)
+				if err != nil {
+					return 0, err
```
njhale, line 45: This implies that, as long as there's a bad TTL in the config, expired events won't get cleaned up. Is that an acceptable failure mode?

client-side validation should make it uncommon, but the edge case will always exist as long as there's no server-side validation.

docs/docs/100-reference/01-command-line/acorn_install.md
```diff
@@ -31,6 +31,7 @@ acorn install
       --cluster-domain strings                          The externally addressable cluster domain (default .oss-acorn.io)
       --controller-replicas int                         acorn-controller deployment replica count
       --controller-service-account-annotation strings   annotation to apply to the acorn-system service account
+      --event-ttl string                                Amount of time an Acorn event will be stored before being deleted (default '24h' - 24 hours)
```
cjellick, line 34: can we do seven days or will that generate too much data?

pkg/controller/eventinstance/eventinstance.go
cjellick: id prefer 7 days

pkg/controller/eventinstance/eventinstance.go
```diff
@@ -0,0 +1,91 @@
...
+			return last.ttl, nil
+		},
+	}.gcExpired
+}
+
+// ttlFunc is a function that returns the TTL to use for event expiration.
+type ttlFunc func(context.Context, kclient.Reader) (time.Duration, error)
+
+type handler struct {
+	ttl ttlFunc
```
cjellick, line 60: nit: calling this ttl confused me bc you also `ttl` for the duration variable. is there anything else you could name it to disambiguate it for clarity

pkg/controller/eventinstance/eventinstance.go
```diff
@@ -0,0 +1,91 @@
...
+	// last stores pre-parsed TTLs from the configuration.
+	var last struct {
+		sync.Mutex
+		raw string
+		ttl time.Duration
+	}
+
+	return handler{
+		ttl: func(ctx context.Context, getter kclient.Reader) (time.Duration, error) {
+			cfg, err := config.Get(ctx, getter)
```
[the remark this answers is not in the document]
StrongMonkey: Would it be better if we can initialize the value so we have to call this every time event is synced? Given the amount of the event it generates I think we will be calling this a lot

pkg/controller/eventinstance/eventinstance.go
cjellick: yeah, i think so. you also ultimately log the error, so 👍

pkg/controller/eventinstance/eventinstance.go
```diff
@@ -0,0 +1,91 @@
...
+
+const defaultTTL = 24 * time.Hour
+
+func GCExpired() router.HandlerFunc {
+	// last stores pre-parsed TTLs from the configuration.
+	var last struct {
+		sync.Mutex
+		raw string
+		ttl time.Duration
+	}
```
thedadams, line 24: The only real issue I have with this approach is that it makes the memory difficult to reason about.

For example, typically you should use a `*sync.Mutex` to ensure that things don't get copied. However, I _think_ that the closure takes the address of the `last` variable so it's OK here?

Is it worth just making `*sync.Mutex` a pointer to avoid confusion?

pkg/controller/eventinstance/eventinstance.go
StrongMonkey: I wonder if we can just apply a default TTL setting if the config is bad, just to protect the system from being overwhelmed

pkg/apis/api.acorn.io/v1/types.go
thedadams: I expected the default to be in the scale of multiple days. Something like a week or so.

Is that not advisable?

pkg/controller/eventinstance/eventinstance.go
njhale: 7 days it is. Done.

pkg/controller/eventinstance/eventinstance.go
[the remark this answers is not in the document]
thedadams: Yes, I have an idea on this.

What if the value is read when the handler is initialized. Then another handler function is created that watches the config. When the config changes, the handler reads the new TTL value only then.

Thoughts?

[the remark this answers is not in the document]
njhale: It is called a bunch, but the value _can_ change so we don't want to set and forget.

I think there's a better approach that involves keeping a shared atomic up to date with a separate config controller/watch/etc. WDYT?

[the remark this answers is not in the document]
thedadams: @njhale I don't think we need shared memory. I think the `handler` you created can have another `HandlerFunc` on it that watches the config and updates its TTL value.

pkg/controller/eventinstance/eventinstance.go
njhale: > I wonder if we can just apply a default TTL setting if the config is bad, just to protect the system from being overwhelmed

this was the other route I was going with, but the UX feels a bit odd to me when I:
1. Set a _valid_ non-default TTL
2. Set an invalid TTL (14d "not actually valid here")
3. Am ignorant of my mistake
4. Only ever see the latest 7 days of events.

docs/docs/100-reference/01-command-line/acorn_install.md
njhale: done

pkg/controller/eventinstance/eventinstance.go
njhale: sure, will update.

njhale: done

pkg/controller/eventinstance/eventinstance.go
```diff
@@ -0,0 +1,91 @@
...
+
+type handler struct {
+	getTTL ttlFunc
+}
+
+func (h handler) gcExpired(req router.Request, resp router.Response) error {
+	e := req.Object.(*v1.EventInstance)
+
+	// Get the currently configured TTL
+	ttl, err := h.getTTL(req.Ctx, req.Client)
```
g-linville, line 67: What advantage is there to have this separate handler object with the `getTTL` function on it, and the `last` struct that it maintains? Is it just to prevent us from having to parse the TTL every time this function is called? Because currently it is still going to fetch the Acorn config every time this function is called.

pkg/controller/eventinstance/eventinstance.go
njhale: Ah, great point.

`go vet` would have probably complained if there was a copy going on, and the struct is anonymous and tightly scoped to the function for that reason, but embedding a `sync.Mutex` is definitely not idiomatic.

that said, I haven't noticed `mu := new(sync.Mutex)` very often, so I'm wondering if that's also not idiomatic.

I went for a compromise and added a named field for `mu`, then made `last` a pointer.

pkg/controller/eventinstance/eventinstance.go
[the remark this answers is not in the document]
njhale: > Or I wonder if we have some mechanism like, for important setting X,Y,Z, if that changes, exit and restarting controller. Then we don't have to worry about dynamically change the values

We could. but that seems like an entirely novel concept for our controllers, and I thought there was already precedent for dynamically adjusting to the latest config in other handlers?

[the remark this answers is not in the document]
njhale: > I don't think we need shared memory. I think the handler you created can have another HandlerFunc on it that watches the config and updates its TTL value.

@thedadams If I'm understanding you correctly, this is where I was going with my statement, although I would still consider the `handler.ttl` field a critical region since we can have multiple goroutines reading and writing it.

pkg/controller/eventinstance/eventinstance.go
```diff
@@ -0,0 +1,91 @@
...
+
+type handler struct {
+	ttl ttlFunc
+}
+
+func (h handler) gcExpired(req router.Request, resp router.Response) error {
+	e := req.Object.(*v1.EventInstance)
+
+	// Get the currently configured TTL
+	ttl, err := h.ttl(req.Ctx, req.Client)
```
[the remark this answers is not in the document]
njhale: see https://github.com/acorn-io/acorn/pull/1719#discussion_r1220121319

pkg/controller/eventinstance/eventinstance.go
njhale: > Is it just to prevent us from having to parse the TTL every time this function is called? 

there's a few advantages to the approach:
- I can swap out the `getTTL` function in unit tests more easily
- Keeping the `handler` and `getTTL` unexported, prevents the "real" usage from getting to complicated
- don't need to parse TTL every time

> Because currently it is still going to fetch the Acorn config every time this function is called.

this is on purpose, because the config can change. but there's a discussion going on about this and I'm probably going to change this (see https://github.com/acorn-io/acorn/pull/1719#discussion_r1220052471)

cjellick:
https://github.com/acorn-io/acorn/issues/1708

pkg/controller/eventinstance/eventinstance.go
njhale: I've completely refactored this. No longer using mutexes directly.

njhale:
Requesting another round of reviews.

pkg/controller/eventinstance/eventinstance.go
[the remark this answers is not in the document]
njhale: OK -- so I've refactored a bunch of this so that we're no longer invoking the getter on every event reconciliation. PTAL @thedadams @StrongMonkey and LMK if it makes sense to you.

pkg/controller/eventinstance/eventinstance.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func NewHandler() *Handler {
+	return &Handler{
+		sem: semaphore.NewWeighted(1),
+	}
+}
+
+func (h *Handler) UpdateTTL(req router.Request, _ router.Response) error {
+	if err := h.sem.Acquire(req.Ctx, 1); err != nil {
```
njhale, line 31: I switched to a semaphore because it lets us set a deadline using a context (in this case, it shares a deadline with the request)

pkg/controller/eventinstance/eventinstance.go
```diff
@@ -0,0 +1,105 @@
...
+
+	v1 "github.com/acorn-io/acorn/pkg/apis/internal.acorn.io/v1"
+	"github.com/acorn-io/acorn/pkg/config"
+	"github.com/acorn-io/baaah/pkg/router"
+	"golang.org/x/sync/semaphore"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	kclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const defaultTTL = 168 * time.Hour // 7 days
```
StrongMonkey, line 16: Nit: 24 * 7 * time.Hour make it straightforward 😄

pkg/controller/eventinstance/eventinstance.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+}
+
+func (h *Handler) UpdateTTL(req router.Request, _ router.Response) error {
+	if err := h.sem.Acquire(req.Ctx, 1); err != nil {
+		return fmt.Errorf("failed to acquire TTL semaphore for update: %w", err)
+	}
+	defer h.sem.Release(1)
+
+	cfg, err := config.Get(req.Ctx, req.Client)
```
StrongMonkey, line 36: I think you can fetch the config from req.Object without having to do an extra api call

StrongMonkey (approved):

pkg/controller/eventinstance/eventinstance.go
cjellick: pretty sure this is canonical way to get a config. we do it all over the place

StrongMonkey: yeah but I guess then you will have to use another wrapper to convert con... probably not worth it

njhale: So I switched this back. This call hits the cache most of the time and also does some required merging to get the correct config values.

njhale:
ended up rolling back to something closer to the initial implementation. it works consistently for me.

pkg/controller/eventinstance/eventinstance.go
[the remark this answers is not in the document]
njhale: scratch that, rolled back those changes.

thedadams (approved):

Issue #1598, opened 2023-05-11
When autouprgade parameter is set for service images , auto upgrade is not triggered when you version is made available.


sangee2004:
acorn version v0.7.0-alpha1-43-ge7fb1587+e7fb1587

Steps to reproduce the problem:
Deploy app that consumes a service using service image.

```
containers: {
  nginx: {
    image: "nginx"
    env: {
     "SECRET": "@{secrets.mysec.output.key}"
     "ADDRESS": "@{services.mysec.address}"
     "PORT": "@{services.mysec.ports.3306}"
     "value1": "@{services.mysec.data.test1}"
     "value2": "@{services.mysec.data.test2}"
}
}
}
services: {
     mysec: {
	image: "<imgae>:v0.0.1"
        autoUpgrade: true
        autoUpgradeInterval: "15s"
        }
}
``` 

Make a version of service image made available .

Wait for > autoUpgradeInterval time.
There is no auto upgrade. 

If we do not support autouprgades for service images, then app deployment should have errored out with a parsing error.

cjellick:
@g-linville i am going to make an assumption that this is going to be the same issue and root cause as what @cloudnautique was hitting. so, im using this issue to represent that

g-linville:
Merged to main.

sangee2004:
Tested with acorn version - `v0.7.0-alpha1-152-g1706098f+1706098f`

Deploy app that consumes a service using service image.

```
containers: {
  nginx: {
    image: "nginx"
    env: {
     "NAME": "@{secrets.mysec.my-service-secret.username}"
     "PASSWORD": "@{secrets.mysec.my-service-secret.password}"
}
}
}
services: {
     mysec: {
        image: "<image>:v0.0.2"
        autoUpgrade: true
        autoUpgradeInterval: "15s"
        }
}
```

Make a new version of service image `<image>:v0.0.2` made available .

Wait for autoUpgradeInterval time.

Consuming app gets auto upgraded and reflects the new changes from the new version of the service image.

sangee2004:
Trying to deploys app with service images having autoupgrade format fails with `defined: could not parse reference: <image>:v#.#.# ` error.

Following is the Acornfile used for deploying app:

```
containers: {
  nginx: {
    image: "nginx"
    env: {
     "NAME": "@{secrets.mysec.my-service-secret.username}"
     "PASSWORD": "@{secrets.mysec.my-service-secret.password}"
}
}
}
services: {
     mysec: {
[...]

PR #1729, opened 2023-06-07
Support auto-upgrade patterns in nested acorns

g-linville:
for #1598

I verified that this works with both public and private container images.

### Checklist

### DIFF
--- a/pkg/controller/appdefinition/acorn.go
+++ b/pkg/controller/appdefinition/acorn.go
@@ -79,14 +79,14 @@ func scopeLinks(app *v1.AppInstance, bindings v1.ServiceBindings) (result v1.Ser
 
 func toAcorn(appInstance *v1.AppInstance, tag name.Reference, pullSecrets *PullSecrets, acornName string, acorn v1.Acorn) *v1.AppInstance {
 	var image string
-	_, isPattern := autoupgrade.AutoUpgradePattern(acorn.Image)
+	pattern, isPattern := autoupgrade.AutoUpgradePattern(acorn.Image)
 	isPattern = isPattern && acorn.AutoUpgrade != nil && *acorn.AutoUpgrade
 	if isPattern {
 		image = acorn.Image
 
 		// remove the autoupgrade pattern from the end of the image for resolving the pull secret
 		// the registry is all that really matters for the pull secret so this is safe to do
-		pullSecrets.ForAcorn(acornName, image[:strings.LastIndex(image, ":")])
+		pullSecrets.ForAcorn(acornName, strings.TrimSuffix(image, ":"+pattern))
 	} else {
 		image = images.ResolveTag(tag, acorn.Image)
 		if strings.HasPrefix(acorn.Image, "sha256:") {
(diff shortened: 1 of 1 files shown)


njhale (approved):

pkg/controller/appdefinition/acorn.go
```diff
@@ -77,19 +78,29 @@ func scopeLinks(app *v1.AppInstance, bindings v1.ServiceBindings) (result v1.Ser
 }
 
 func toAcorn(appInstance *v1.AppInstance, tag name.Reference, pullSecrets *PullSecrets, acornName string, acorn v1.Acorn) *v1.AppInstance {
-	image := images.ResolveTag(tag, acorn.Image)
-	if strings.HasPrefix(acorn.Image, "sha256:") {
-		image = strings.TrimPrefix(acorn.Image, "sha256:")
+	var image string
+	_, isPattern := autoupgrade.AutoUpgradePattern(acorn.Image)
+	isPattern = isPattern && acorn.AutoUpgrade != nil && *acorn.AutoUpgrade
```
thedadams, line 83: An auto-upgrade pattern is enough for the app to be considered an "auto upgraded app"

pkg/controller/appdefinition/acorn.go
```diff
@@ -77,19 +78,29 @@ func scopeLinks(app *v1.AppInstance, bindings v1.ServiceBindings) (result v1.Ser
...
+	if isPattern {
+		image = acorn.Image
+
+		// remove the autoupgrade pattern from the end of the image for resolving the pull secret
+		// the registry is all that really matters for the pull secret so this is safe to do
+		pullSecrets.ForAcorn(acornName, image[:strings.LastIndex(image, ":")])
```
thedadams, line 89: nit: if you don't ignore the first return of the call to `autoupgrade.AutoUpgradePattern` above (say you call it `pattern`), then this could be
```suggestion
		pullSecrets.ForAcorn(acornName, strings.TrimSuffix(image, pattern))
```

pkg/controller/appdefinition/acorn.go
g-linville: A user could still theoretically create a nested Acorn with a pattern in the image but have forgotten to set the auto upgrade parameter though, so I figured it would be best to explicitly check for that as well.

pkg/controller/appdefinition/acorn.go
g-linville: Good point, that is a lot cleaner. Updated.

pkg/controller/appdefinition/acorn.go
thedadams: I am saying that a user doesn't need to explicitly set `AutoUpgrade` on a nested acorn. Using an auto-upgrade pattern is enough to say that auto-upgrade is on.

The logic, as it is now, requires both a pattern and explicitly setting `AutoUpgrade`.

g-linville: Oh I did not realize that we infer that auto-upgrade is on based on finding the pattern. Fixed.

thedadams (approved):

PR #1731, opened 2023-06-07
Use subtests to parallelize image variation test cases

njhale:

### DIFF
--- a/integration/client/apps/apps_test.go
+++ b/integration/client/apps/apps_test.go
@@ -482,17 +482,17 @@ func TestAppRunImageVariations(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	imageNames := []string{
-		"foo/bar:baz",
-		imageID,
-		fmt.Sprintf("sha256:%s", imageID),
-		imageID[:8],
-	}
-
-	for _, imageName := range imageNames {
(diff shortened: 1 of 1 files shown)


iwilltry42 (approved):

integration/client/apps/apps_test.go
```diff
@@ -482,17 +482,17 @@ func TestAppRunImageVariations(t *testing.T) {
...
-		if err != nil {
-			t.Fatal(err)
-		}
+	for desc, imageName := range map[string]string{
+		"ref":    "foo/bar:baz",
+		"id":     imageID,
+		"sha256": fmt.Sprintf("sha256:%s", imageID),
+		"short":  imageID[:8],
+	} {
+		imageName := imageName
```
g-linville, line 491: What is the purpose of this line?

njhale: the `imageName` declared by the `for` loop lives outside the scope of a particular iteration; shadowing it within an iteration means the closure will bind that instead. If we didn't shadow, the value would change inside each closure as the loop progressed.

Here's a simplified example that might make things more clear: https://goplay.tools/snippet/_Tkb2zAiI_v

g-linville (approved):

Issue #1605, opened 2023-05-15
Acorn image references do not check local store


cloudnautique:
If you build an Acorn image `index.docker.io/cloudnautique/foo` and do not push to the registry, and then try to use it like so:

```
acorns: app: image: "index.docker.io/cloudnautique/foo"
```

It will fail because it can't find it in the remote registry. It never checks the local store.

Acorn should check the local store as well before failing.

cjellick:
Seems like a different manifestation of https://github.com/acorn-io/acorn/issues/1607

g-linville:
Merged to main.

sangee2004:
Tested with `acorn version v0.7.0-38-gdc95c02b+dc95c02b`
Able to deploy successfully when nested acorn image points to a local image not pushed to the registry yet.

Built an image `docker.io/sangeetha/myfirstnestedacorn:v0.0.1` (do not push)

Deploy app using following Acornfile:
```
acorns: {
   "my-app": {
       image: "docker.io/sangeetha/myfirstnestedacorn:v0.0.1"
   }
}
```

App deployment succeeds.

@g-linville When app is listed , we see the `imageid` being listed instead of the image tag. Why is this the case ?

```
acorn apps                          
NAME                     IMAGE                                                 HEALTHY   UP-TO-DATE   CREATED     ENDPOINTS                                                                                MESSAGE
localacorn.my-app        706fcef76cd5                                          1         1            5m29s ago   http://myweb1-localacorn-my-app-ec06a134-f9949ec5.local.oss-acorn.io => myweb1:80        OK

```
I would have expected the image to be `docker.io/sangeetha/myfirstnestedacorn:v0.0.1` in this case.

g-linville:
@sangee2004 yep, this is expected. The current behavior of nested acorns always sets the `.spec.image` field on the AppInstance to the ID instead of the tag (for images pull from a remote registry), so I kept that consistent for local images too.

sangee2004:
@g-linville For images pulled from a remote Registry, I see the image tag being used.
Acornfile for app:
```
acorns: {
    "my-acorn1": {
        image: "ghcr.io/acorn-io/library/hello-world:latest"
    }
    "my-acorn2": {
        image: "docker.io/sangeetha/myfirstacorn:v0.0.1"(pushed to registry)
        autoUpgrade: true
        autoUpgradeInterval: "15s"
    }
   "my-acorn3": {
[...]

PR #1735, opened 2023-06-07
Add support for local Acorn images in nested Acorns (#1605)

g-linville:
for #1605

There are some things that I don't love about this approach, but this is what I have for now. I would love any feedback or suggestions. I left comments on sections of the code for additional explanation or to express my doubts.

This change causes nested Acorns to prioritize local images over remote images. So if the image specified in the Acornfile matches the ID/Name or tag of an image in the same project as the build, the local image will be used. Otherwise, it will attempt to pull it from the remote registry if it is specified.

### Checklist

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"time"
 
+	apiv1 "github.com/acorn-io/acorn/pkg/apis/api.acorn.io/v1"
 	v1 "github.com/acorn-io/acorn/pkg/apis/internal.acorn.io/v1"
 	"github.com/acorn-io/acorn/pkg/appdefinition"
 	"github.com/acorn-io/acorn/pkg/autoupgrade"
@@ -361,14 +362,14 @@ func pullImage(ctx *buildContext, image string) (id string, err error) {
 	return pushTarget.Context().Digest(digest.String()).String(), nil
 }
 
-func resolveLocalImage(ctx *buildContext, image string) (string, error) {
+func resolveLocalImage(ctx *buildContext, imageName string) (string, error) {
 	c, err := k8sclient.Default()
 	if err != nil {
 		return "", err
 	}
 
 	// very important - make sure we only list images in the buildNamespace to avoid finding ones in other projects
-	imageList := v1.ImageInstanceList{}
+	imageList := apiv1.ImageList{}
 	err = c.List(ctx.ctx, &imageList, &kclient.ListOptions{
 		Namespace: ctx.buildNamespace,
 	})
@@ -376,32 +377,18 @@ func resolveLocalImage(ctx *buildContext, image string) (string, error) {
 		return "", err
 	}
 
-	var digest string
-imageLoop:
-	for _, imageInstance := range imageList.Items {
-		// check whether the image specified by the user matches the name (ID) of the image
-		// require at least the first three characters to match
-		if len(image) > 2 && strings.HasPrefix(imageInstance.Name, image) {
-			digest = imageInstance.Digest
-			break
-		}
-
-		// check to see if it matches any of the tags on the image
-		for _, tag := range imageInstance.Tags {
-			if tag == image {
-				digest = imageInstance.Digest
-				break imageLoop
-			}
-		}
+	image, _, err := images2.FindImageMatch(imageList, imageName)
+	if err != nil {
+		return "", err
 	}
 
-	if digest != "" {
+	if image.Digest != "" {
 		pushTarget, err := imagename.ParseReference(ctx.pushRepo)
 		if err != nil {
 			return "", err
 		}
 
-		return pushTarget.Context().Digest(digest).String(), nil
+		return pushTarget.Context().Digest(image.Digest).String(), nil
 	}
 	return "", fmt.Errorf("could not find local image %s", image)
 }
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -200,17 +200,14 @@ func buildAcorns(ctx *buildContext, acorns map[string]v1.AcornBuilderSpec) (map[
 				continue
 			}
 
-			var (
-				id  string
-				err error
-			)
-			if isImageRemote(ctx, acornImage.Image) {
-				id, err = pullImage(ctx, acornImage.Image)
-			} else {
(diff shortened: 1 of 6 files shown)


pkg/build/build.go
```diff
@@ -350,6 +364,69 @@ func pullImage(ctx *buildContext, image string) (id string, err error) {
...
+
+func resolveLocalImage(ctx *buildContext, image string) (string, error) {
+	// We need to be careful to make sure we only list ImageInstances in the namespace to which the user has access.
+	// We can determine the namespace based on the pushRepo, which is in the format <registry>/acorn/<namespace>.
+	// To be very safe, verify the structure of pushRepo before getting the namespace from it.
+	if len(strings.Split(ctx.pushRepo, "/")) != 3 || !strings.HasPrefix(ctx.pushRepo, "127.0.0.1:") || strings.Split(ctx.pushRepo, "/")[1] != "acorn" {
+		return "", fmt.Errorf("invalid pushRepo: %s", ctx.pushRepo)
+	}
+
+	namespace := strings.Split(ctx.pushRepo, "/")[2]
```
g-linville, line 389: This is the part of this that I dislike the most. I tried to put in enough safeguards such that, if in the future someone refactors how `pushRepo` works, it should be caught here and errors thrown instead of accidentally granting image access across projects.

One possible way around this would be to store another piece of information besides pushRepo in the Token, and that would simply be the namespace (project name) itself. That would be duplication of information for now, since that info is extractable from the pushRepo, but then it would at least be more explicit that that is something we are relying on. Thoughts?

pkg/build/build.go
```diff
@@ -350,6 +363,74 @@ func pullImage(ctx *buildContext, image string) (id string, err error) {
...
```
g-linville, line 388: This is the part of this that I dislike the most. I tried to put in enough safeguards such that, if in the future someone refactors how `pushRepo` works, it should be caught here and errors thrown instead of accidentally granting image access across projects.

One possible way around this would be to store another piece of information besides pushRepo in the Token, and that would simply be the namespace (project name) itself. That would be duplication of information for now, since that info is extractable from the pushRepo, but then it would at least be more explicit that that is something we are relying on. Thoughts?

pkg/build/build.go
```diff
@@ -350,6 +363,74 @@ func pullImage(ctx *buildContext, image string) (id string, err error) {
...
+
+	c, err := k8sclient.Default()
+	if err != nil {
+		return "", err
+	}
```
g-linville, line 393: Another thing. Initially I added the k8sclient to the `buildContext` struct and passed it down from the build server all the way to here. But it's pretty easy to just recreate it here. Not sure if there's an advantage to either approach, but I chose this one. Also curious to hear any thoughts on this.

pkg/build/build.go
```diff
@@ -350,6 +363,73 @@ func pullImage(ctx *buildContext, image string) (id string, err error) {
 	return pushTarget.Context().Digest(digest.String()).String(), nil
 }
 
+func isImageRemote(ctx *buildContext, image string) (bool, error) {
+	ref, err := images2.ParseReferenceNoDefault(image)
+	if err != nil {
+		return false, nil
```
g-linville, line 369: I don't return the error here because this error can happen if there is no registry supplied in the image string (i.e. if it's just `myimage:latest`). This is not a problem for local images.

pkg/build/build.go
thedadams: Can this [`token.Build.Namespace`](https://github.com/acorn-io/acorn/blob/dd4f1c7c8b6add7c9ecbfc2afe5e10bac0be5964/pkg/buildserver/server.go#L125) be added to the [`buildContext`](https://github.com/acorn-io/acorn/blob/959469c916c908502448dab058258701c99c506c/pkg/build/build.go#L38) to make this simpler?

g-linville: Oooh, very nice. I will update it to use that

pkg/build/build.go
thedadams: My opinion, for whatever it's worth, is that this is fine.

thedadams (approved):

pkg/build/build.go
```diff
@@ -350,6 +361,51 @@ func pullImage(ctx *buildContext, image string) (id string, err error) {
...
+	err = c.List(ctx.ctx, &imageList, &kclient.ListOptions{
+		Namespace: ctx.buildNamespace,
+	})
+	if err != nil {
+		return "", err
+	}
+
+	var digest string
+imageLoop:
+	for _, imageInstance := range imageList.Items {
```
iwilltry42, line 381: Could this be replaced by the `FindImageMatch` function that's also used by `client.ImageGet`?
https://github.com/acorn-io/acorn/blob/main/pkg/images/find.go#L39

iwilltry42 (approved):
Just one suggestion that may or may not be helpful. Apart from that LGTM

pkg/build/build.go
g-linville: I'll go try it

iwilltry42: I guess there's some easy way to translate from ImageInstance to Image, since they're basically the same struct-wise, but I'm not sure if it's worth the effort now.

g-linville: Well it turns out my assumption was wrong that I wouldn't be able to get the public Acorn API types. So I just updated it to use Images instead of ImageInstances and to use the FindImageMatch function. Thanks for the suggestion - glad we don't need to duplicate the logic of that function

pkg/build/build.go
```diff
@@ -350,6 +362,37 @@ func pullImage(ctx *buildContext, image string) (id string, err error) {
...
+
+	if image.Digest != "" {
+		pushTarget, err := imagename.ParseReference(ctx.pushRepo)
+		if err != nil {
+			return "", err
+		}
+
+		return pushTarget.Context().Digest(image.Digest).String(), nil
+	}
+	return "", fmt.Errorf("could not find local image %s", image)
```
iwilltry42, line 393:
```suggestion
	return "", fmt.Errorf("could not find local image %s", imageName)
```

iwilltry42 (approved):

thedadams (approved):

Issue #1715, opened 2023-06-05
ImageAllowRules - Change order of presenting the options for allow image.


sangee2004:
acorn version v0.7.0-alpha1-148-g7909ed41+7909ed41

Order in which options for allow image is presented to the user keeps changing . We need to stick to showing the same ordered list always.

Mostly I am presented with this list of options:
```
> NO
  yes (this tag only)
  registry (all images in this registry)
  repository (all images in this repository)
  all (all images out there)
```

Can the `repository (all images in this repository)` be moved up  `registry (all images in this registry)` so that the image allow becomes most restrictive to least restrictive?

This shows the order in which options for allow image is presented to the user keeps changing:

```
 % acorn run -n mytest1 sangeetha/myfirstservice:v0.0.1 
  •  WARNING:  This application would like to use the image 'sangeetha/myfirstservice:v0.0.1'.
                 This image is not trusted by any image allow rules in this project.
                 This could be VERY DANGEROUS to the cluster if you do not trust this
                 application. If you are unsure say no.

? Do you want to allow this app to use this (POTENTIALLY DANGEROUS) image?  [Use arrows to move, type to filter]
> NO
  yes (this tag only)
  registry (all images in this registry)
  repository (all images in this repository)
  all (all images out there)
  ✗  ERROR:  interrupt: image not found: sangeetha/myfirstservice:v0.0.1

% acorn run -n mytest1 sangeetha/myfirstservice:v0.0.1 
  •  WARNING:  This application would like to use the image 'sangeetha/myfirstservice:v0.0.1'.
                 This image is not trusted by any image allow rules in this project.
                 This could be VERY DANGEROUS to the cluster if you do not trust this
                 application. If you are unsure say no.

? Do you want to allow this app to use this (POTENTIALLY DANGEROUS) image?  [Use arrows to move, type to filter]
  repository (all images in this repository)
  all (all images out there)
> NO
  yes (this tag only)
  registry (all images in this registry)
  ✗  ERROR:  interrupt: image not found: sangeetha/myfirstservice:v0.0.1

% acorn run -n mytest1 sangeetha/myfirstservice:v0.0.1 
  •  WARNING:  This application would like to use the image 'sangeetha/myfirstservice:v0.0.1'.
                 This image is not trusted by any image allow rules in this project.
                 This could be VERY DANGEROUS to the cluster if you do not trust this
                 application. If you are unsure say no.

[...]

PR #1741, opened 2023-06-08
change: preserve the order of the choices in the prompt for imageallowrules scope

iwilltry42:
removed from the title: "incredibly ugly way to ..."

If you have any prettier idea to preserve the order as written in the maps, feel free to suggest :)
I couldn't find a simple way without writing some custom sorting function which seems overkill for this.

Ref #1715

### Checklist


g-linville (approved):
I think it's fine. My only suggestion is to add a comment in the code explaining that it is formatted this way in order to preserve the order.

g-linville (approved):
I think it's fine. My only suggestion is to add a comment in the code explaining that it is formatted this way in order to preserve the order.

thedadams (approved):
I think this is fine.

Another possible approach would be to add a `Display` function/method in the `imageallowrules` package. The only advantage of this would be that you could construct the `choices` slice as a `[]imageallowrules.SimpleImageScope` and do do something like
```
for _, choice := range choices {
    choiceMap[choice.Display()] = string(choice)
}
```

Not sure if it is worth the trouble, but I know you were commenting how you didn't like this code 😅

iwilltry42:
@g-linville I added a comment for clarity.
@thedadams I'll just close my eyes when scrolling past it in the future :)

Issue #1713, opened 2023-06-05
ImageAllowRule - Deploying apps using image-ids that were already allowed , results in user being prompted to allow images (resulting in duplicate imageallowrules created)


sangee2004:
acorn version v0.7.0-alpha1-148-g7909ed41+7909ed41

Steps to reproduce the problem:
Install acorn with --features image-allow-rules=true

Scenario1:
Deploy app from Acornfile.
When prompted for allowing images , select "yes" and allow for app to get deployed successfully.( Do you want to allow this app to use this (POTENTIALLY DANGEROUS) image? yes (this ID or SHA only))

```
acorn apps
mytest1            89f49874572c   1         1            13m ago     http://myweb1-mytest1-4984b863.local.oss-acorn.io => myweb1:80            OK
```
Deploy app using this image-id

```
acorn run 89f49874572c                    
  •  WARNING:  This application would like to use the image '89f49874572c03b67c57be1433e731ab9af42c9688d97610d6d1fe569a2ce69b'.
                 This image is not trusted by any image allow rules in this project.
                 This could be VERY DANGEROUS to the cluster if you do not trust this
                 application. If you are unsure say no.

? Do you want to allow this app to use this (POTENTIALLY DANGEROUS) image?  [Use arrows to move, type to filter]
> NO
  yes (this ID or SHA only)
```
Expected Behavior:
We should be allowed to deploy this app without prompting user since this is already an allowed image.

Scenario2:

If you choose "yes" , app deployment succeed . 
This result in duplicate imageAllowRules created for the same imageId.

```
sangeethahariharan@Sangeethas-MacBook-Pro example1 % kubectl describe ImageAllowRules -n new3  
Name:         bold-rain
Namespace:    new3
Labels:       <none>
Annotations:  <none>
API Version:  api.acorn.io/v1
Images:
  89f49874572c03b67c57be1433e731ab9af42c9688d97610d6d1fe569a2ce69b
Kind:  ImageAllowRule
Metadata:
  Creation Timestamp:  2023-06-05T19:44:57Z
  Generation:          1
  Resource Version:    1295125
  UID:                 d13a4ab2-ac05-4537-8820-092690ac825b-p
Signatures:
Events:  <none>


Name:         autumn-wave
Namespace:    new3
Labels:       <none>
Annotations:  <none>
API Version:  api.acorn.io/v1
Images:
  89f49874572c03b67c57be1433e731ab9af42c9688d97610d6d1fe569a2ce69b
[...]

PR #1742, opened 2023-06-08
change: try to get existing image before converting to app (#1713)

iwilltry42:
The problem is that we pass the unprocessed user input into the AppSpec, which will make IAR checks fail since they match against full IDs if IDs are used, so partial ID inputs will make it fail verification and thus double-prompt users to create new IARs.

In addition to that, this PR now also adds the Image ID to the list of allowed images for every IAR that is generated from prompt, as otherwise running an app by repo-name and later by ID would result in another prompt, while that exact image was allowed before (just that time referenced by name).

Ref #1713 

### Checklist

### DIFF
--- a/pkg/client/app.go
+++ b/pkg/client/app.go
@@ -72,7 +72,7 @@ func ToApp(namespace, image string, opts *AppRunOptions) *apiv1.App {
 }
 
 func (c *DefaultClient) AppRun(ctx context.Context, image string, opts *AppRunOptions) (*apiv1.App, error) {
-	img, tag, err := GetImageRef(ctx, c, image)
+	img, tag, err := FindImage(ctx, c, image)
 	if err != nil && !errors.As(err, &images.ErrImageNotFound{}) {
 		return nil, err
 	} else if err == nil && img != nil {
@@ -107,11 +107,14 @@ func ToAppUpdate(ctx context.Context, c Client, name string, opts *AppUpdateOpti
 	}
 
 	if opts.Image != "" {
-		img, err := c.ImageGet(ctx, opts.Image)
+		img, tag, err := FindImage(ctx, c, opts.Image)
 		if err != nil && !apierrors.IsNotFound(err) {
 			return nil, err
-		} else if img != nil {
+		} else if err == nil && img != nil {
 			opts.Image = img.Name
+			if tag != "" {
+				opts.Image = tag
+			}
 		}
 		app.Spec.Image = opts.Image
 	}
--- a/pkg/client/app.go
+++ b/pkg/client/app.go
@@ -70,11 +70,14 @@ func ToApp(namespace, image string, opts *AppRunOptions) *apiv1.App {
 }
 
 func (c *DefaultClient) AppRun(ctx context.Context, image string, opts *AppRunOptions) (*apiv1.App, error) {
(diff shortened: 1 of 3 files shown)


g-linville (approved):

iwilltry42:
@g-linville & @thedadams sorry, but while fixing the tests, I changed quite a bit and enhanced this PR a little.

pkg/client/image.go
```diff
@@ -229,3 +229,13 @@ func (c *DefaultClient) ImageList(ctx context.Context) ([]apiv1.Image, error) {
 
 	return result.Items, nil
 }
+
+// GetImageReference finds an image if exists and returns if it was found by tag
```
g-linville, line 233: nit:

```suggestion
// GetImageRef finds an image if exists and returns if it was found by tag
```

pkg/client/app.go
```diff
@@ -96,6 +107,12 @@ func ToAppUpdate(ctx context.Context, c Client, name string, opts *AppUpdateOpti
 	}
 
 	if opts.Image != "" {
+		img, err := c.ImageGet(ctx, opts.Image)
```
g-linville, line 110: Just curious, why not also use the `GetImageRef` function here?

pkg/rulerequest/handle.go
```diff
@@ -41,23 +41,25 @@ func PromptRun(ctx context.Context, c client.Client, dangerous bool, image strin
 		if err != nil {
 			return nil, err
 		}
-		var img *apiv1.Image
-		var tag string
-		img, tag, err = images.FindImageMatch(apiv1.ImageList{Items: il}, image)
+		var exImg *apiv1.Image
+		var exImgTag, exImgName string
```
g-linville, line 45: nit: if `ex` stands for `exact` here, I recommend just including that full word in the variable names. It was unclear to me what these were for until I saw the comment farther down.

g-linville (approved):

pkg/client/image.go
iwilltry42: Renamed that function actually because there is `GetImageReference` already in that package and it doesn't really reflect what it does.

pkg/client/app.go
iwilltry42: Because I forgot to apply it here :grimacing:

pkg/rulerequest/handle.go
iwilltry42: `existing` it is :)

g-linville (approved):

pkg/client/app.go
```diff
@@ -96,6 +107,15 @@ func ToAppUpdate(ctx context.Context, c Client, name string, opts *AppUpdateOpti
 	}
 
 	if opts.Image != "" {
+		img, tag, err := FindImage(ctx, c, opts.Image)
+		if err != nil && !apierrors.IsNotFound(err) {
```
thedadams, line 111: Here you are checking for `apierros.IsNotFound` and above you are checking `images.ErrImageNotFound{}`. Is this intentional?

thedadams (approved):
One question. Otherwise, looks good.

pkg/client/app.go
iwilltry42: Absolutely not intentional. Thanks for the hint!

PR #1755, opened 2023-06-09
events/order

njhale:
- Print acorn events in chronological order
- Add unit tests for event list query
- Expose table writer flush method
- Drop formatting for event table timestamps
- Enable acorn event streaming
- Update acorn event docs


pkg/server/registry/apigroups/acorn/events/testdata/query/TestQuery/full/expected.golden
```diff
@@ -0,0 +1,5 @@
+`0a6986e5dc34a7f211db5f8881d9160e AppCreate           "2023-06-08T13:32:36.345221311-04:00"
```
njhale, line 1: The timezone is causing a diff when running these tests in CI. I think this is a side-effect of the way I'm using jsonpath.

njhale (commented):
Left some notes.

pkg/cli/events.go
```diff
@@ -55,8 +61,22 @@ func (e *Events) Run(cmd *cobra.Command, args []string) error {
 	}
 
 	out := table.NewWriter(tables.Event, false, e.Output)
+	var flush bool
 	for event := range events {
-		out.Write(event)
+		if event != nil {
+			out.Write(event)
+		} else {
+			// A nil event means that it's time to flush
```
njhale, line 69: This feels super kludgy.

pkg/server/registry/apigroups/acorn/events/strategy.go
```diff
@@ -53,28 +67,107 @@ type query struct {
...
+
+	// Attempt to filter
+	obj := e.Object.(*apiv1.Event)
+	filtered := q.filter(*obj)
+	if len(filtered) < 1 {
+		// Drop the event, it's been filtered out
+		return nil
+	}
+
+	e.Object = filtered[0].DeepCopy()
```
njhale, line 91: Not sure if this `DeepCopy()` is necessary.

pkg/cli/events.go
ibuildthecloud: This does seem hacky. You might be abusing the table wrtier, but maybe it's fine.  I didn't know a tabwriter had a flush.  Fundamentally it has to buffer so that it can align columns so I don't understand how flush works.  Does it calculate column width historically for all rows, or just what is in the buffer.

Anyhow.  And alternate approach is that you just do a flush interval.  So one goroutine that a loop on an interval that calls flush. Like once a second.

pkg/server/registry/apigroups/acorn/events/strategy.go
```diff
@@ -53,28 +67,107 @@ type query struct {
...
+	if err != nil {
+		return
+	}
+
+	q.tail, stripped.Predicate.Limit = stripped.Predicate.Limit, 0
+
+	return
+}
+
+func forward(ctx context.Context, from <-chan watch.Event, to chan<- watch.Event, q query) error {
```
tylerslaton, line 146: Can we add header comments to this and the other functions?

pkg/cli/events.go
njhale: The behavior I've observed is that column width gets reset on flush, which makes for some pretty ugly tables.

Fortunately, it turns out that there's a drop-in replacement for `text/tabwriter` that can  remember column width https://pkg.go.dev/github.com/liggitt/tabwriter _and_ it's used by `kubectl` too.

pkg/server/registry/apigroups/acorn/events/strategy.go
njhale: done

pkg/cli/events.go
```diff
@@ -55,8 +61,22 @@ func (e *Events) Run(cmd *cobra.Command, args []string) error {
...
+			flush = true
+		}
+
+		if !flush {
+			continue
+		}
+
+		if err := out.Flush(); err != nil {
+			break
+		}
```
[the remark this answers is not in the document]
njhale: @g-linville Refactored a bit. PTAL

pkg/cli/events.go
```diff
@@ -55,8 +61,22 @@ func (e *Events) Run(cmd *cobra.Command, args []string) error {
 	}
 
 	out := table.NewWriter(tables.Event, false, e.Output)
+	var flush bool
```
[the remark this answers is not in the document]
njhale: https://github.com/acorn-io/acorn/pull/1755#discussion_r1227373208

pkg/cli/events.go
njhale: switched to that package and the formatting looks much better on repeated flush.

njhale:
I've addressed all feedback and refactored a bunch. PTAL.

g-linville (approved):
Lgtm, love the new channel.go file and the functions it provides.

PR #1770, opened 2023-06-12
Override acornfile env values with values from app.Spec.Environment

ibuildthecloud:

### DIFF
--- a/pkg/controller/appdefinition/deploy.go
+++ b/pkg/controller/appdefinition/deploy.go
@@ -31,6 +31,7 @@ import (
 	apierror "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/sets"
 	kclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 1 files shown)


cjellick:
For testing purposes:

The old behavior would have been that the pod would have had env vars that looks something like
```
env:
- name: FOO
   value: {value from acornfile}
- name: FOO
   value: {value from command line}
``` 
and apparently that meant the actually running container got the first value. The new behavior means 
```
- name: FOO
   value: {value from acornfile}
```
will be omitted entirely and you'll just get the command line value, which will make it work as expected.

sangee2004:
Able to reproduce the issue in 0.7.0.

Deploy app using  `acorn run -n mytest1 -e NAME=value2 .`
Acornfile:
```
containers: {
  nginx: {
    image: "nginx"
    env: {
     "NAME": "value1"
}
}
}
```
Pod spec has 2 env entries for "NAME":

```
           "spec": {
                "containers": [
                    {
                        "env": [
                            {
                                "name": "NAME",
                                "value": "value1"
                            },
                            {
                                "name": "NAME",
                                "value": "value2"
                            }
                        ],
```

Tested the same scenario with acorn version `v0.7.0-45-g12ffac87+12ffac87`

Pod spec has only 1 env entries for "NAME" corresponding to the value from command line.

```
            "spec": {
                "containers": [
                    {
                        "env": [
                            {
                                "name": "NAME",
                                "value": "value2"
                            }
                        ],
```

PR #1788, opened 2023-06-14
Fix: Allow execing into jobs

cjellick:
When execing, you need to specify which container in the pod you're
execing into. We were relying only on the
containerReplica.spec.containerName field for this, but in the case of a
job, that field is blank and jobName is set. This changes accounts for
that scenario.

### DIFF
--- a/pkg/server/registry/apigroups/acorn/containers/exec.go
+++ b/pkg/server/registry/apigroups/acorn/containers/exec.go
@@ -108,6 +108,9 @@ func (c *ContainerExec) Connect(ctx context.Context, id string, options runtime.
 	}
 
 	containerName := container.Spec.ContainerName
+	if containerName == "" {
+		containerName = container.Spec.JobName
+	}
 	if container.Spec.SidecarName != "" {
 		containerName = container.Spec.SidecarName
 	}


thedadams (approved):

njhale (approved):

sangee2004:
Tested with acorn version v0.7.0-64-g0c26c87a+0c26c87a

Deploy App that has job definition

```
containers: app: {
  image: "alpine"
  entrypoint: ["sleep"]
  command: ["3600"]
  dependsOn: ["migrate"]
}

jobs: migrate: {
  image: "alpine"
  entrypoint: ["sleep"]
  command: ["3600"]
}
```

Able to exec into job as expected.

```
acorn containers 
NAME                         APP            IMAGE     STATE     RESTARTCOUNT   CREATED     MESSAGE
mytestdepend.migrate-5hwfq   mytestdepend   alpine    running   0              2m19s ago   

 % acorn exec mytestdepend.migrate-5hwfq 
/ # ls
bin    dev    etc    home   lib    media  mnt    opt    proc   root   run    sbin   srv    sys    tmp    usr    var
/ # 

acorn exec mytestdepend 
/ # ls
bin    dev    etc    home   lib    media  mnt    opt    proc   root   run    sbin   srv    sys    tmp    usr    var
/ # 

```

Issue #1738, opened 2023-06-07
Acorn update on apps deployed with remote images picks the latest version of image available locally.


sangee2004:
**acorn version** v0.7.0-alpha1-168-g5dc719d0+5dc719d0

### Steps to reproduce the problem:

**Scenario 1:**

1. Deploy app with any remote image - ` acorn run -n mytest1 sangeetha/myfirstacorn:v0.0.1`
```
 acorn images
REPOSITORY                               TAG       IMAGE-ID
index.docker.io/sangeetha/myfirstacorn   v0.0.1    9cb84f3628cc
```
2. Build new version of image with same image tag used in step1
```
acorn build -t index.docker.io/sangeetha/myfirstacorn:v0.0.1 .
```
After this I see only 1 image entry which corresponds to the locally built version.
```
acorn images
REPOSITORY                               TAG       IMAGE-ID
index.docker.io/sangeetha/myfirstacorn   v0.0.1    76bad608b788
```

3. Update app - `acorn update -l test1=value1 mytest1` , I see the app getting deployed with the locally built version.

**Scenario 2:**

1. Deploy app with any remote image - `acorn run -n mytest1 sangeetha/myfirstacorn:latest`
```
 acorn images
REPOSITORY                               TAG       IMAGE-ID
index.docker.io/sangeetha/myfirstacorn   latest    331bdb2ed4ee
```
2. Make a new version of the `sangeetha/myfirstacorn:latest` made available ( build and push from a different project).
3. `acorn pull index.docker.io/sangeetha/myfirstacorn:latest`
```
acorn images                                            
REPOSITORY                               TAG       IMAGE-ID
index.docker.io/sangeetha/myfirstacorn   latest    fb246b5524ea
```
4. Update app - `acorn update -l test1=value1 mytest1` 
This results in app getting deployed with the latest version of the image built in step2.

**Scenario 3:**

1. Deploy app with any remote image - `acorn run -n mytest1 sangeetha/myfirstacorn:latest`
```
acorn images
REPOSITORY                               TAG       IMAGE-ID
index.docker.io/sangeetha/myfirstacorn   latest    fb246b5524ea
```
2. Make a new version of the `sangeetha/myfirstacorn:latest` made available ( build and push from a different project).
3. Deploy another app using same image -`acorn run -n mytest1 sangeetha/myfirstacorn:latest`
```
 acorn images 
REPOSITORY                               TAG       IMAGE-ID
[...]

PR #1789, opened 2023-06-14
Fix `acorn update` so that it won't unexpectedly change the app image (#1738)

g-linville:
for #1738

This PR fixes all three scenarios in that issue. In general, this is the behavior that I observed while testing these changes:

- Running `acorn update` without `--image` or `--pull` will not change the app's image.
- Running `acorn update --image <same image as already specified when the app was first run>` will also not change the app's image
- Running `acorn update --image <some different image>` will change the app's image to the new one that was specified
- Running `acorn update --pull` will take the `app.spec.image` (which matches exactly what the user specified as the image when they originally ran `acorn run`) and re-resolve it, prioritizing local images first. If a new image is found, the app is updated to the new image.

I put some comments in the PR to explain some of the changes.

### Checklist

### DIFF
--- a/pkg/cli/run.go
+++ b/pkg/cli/run.go
@@ -313,20 +313,18 @@ func (s *Run) update(ctx context.Context, c client.Client, imageSource imagesour
 		return nil, false, err
 	}
 
-	if !imageSource.IsImageSet() {
-		// If there is no image set, then lookup the existing app and use the image of the current app
-		imageSource = imageSource.WithImage(app.Status.AppImage.ID)
-	}
-
-	image, deployArgs, err := imageSource.GetImageAndDeployArgs(ctx, c)
-	if err != nil {
-		return nil, false, err
-	}
-
 	updateOpts := opts.ToUpdate()
 	updateOpts.Replace = s.Replace
-	updateOpts.Image = image
-	updateOpts.DeployArgs = deployArgs
+
+	if imageSource.IsImageSet() {
+		image, deployArgs, err := imageSource.GetImageAndDeployArgs(ctx, c)
+		if err != nil {
+			return nil, false, err
+		}
+		updateOpts.Image = image
+		updateOpts.DeployArgs = deployArgs
+	}
+
 	app, err = rulerequest.PromptUpdate(ctx, c, s.Dangerous, app.Name, updateOpts)
 	if err != nil {
 		return nil, false, err
--- a/pkg/client/app.go
+++ b/pkg/client/app.go
@@ -106,7 +106,8 @@ func ToAppUpdate(ctx context.Context, c Client, name string, opts *AppUpdateOpti
 		return app, nil
 	}
 
-	if opts.Image != "" {
+	// only change app.Spec.Image if the user specified a different image
+	if opts.Image != "" && opts.Image != app.Spec.Image {
 		img, tag, err := FindImage(ctx, c, opts.Image)
 		if err != nil && !errors.As(err, &images.ErrImageNotFound{}) {
 			return nil, err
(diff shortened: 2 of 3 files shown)


pkg/cli/run.go
```diff
@@ -313,20 +313,18 @@ func (s *Run) update(ctx context.Context, c client.Client, imageSource imagesour
 		return nil, false, err
 	}
 
-	if !imageSource.IsImageSet() {
-		// If there is no image set, then lookup the existing app and use the image of the current app
-		imageSource = imageSource.WithImage(app.Status.AppImage.ID)
-	}
```
g-linville, line 319: This was the main source of the bug. We were taking `app.Status.AppImage.ID` (which could be a number of different things), setting that as the `--image` parameter quietly for `acorn update`, and then moving along.

For Docker Hub in particular this caused issues, because the ID would have `index.docker.io` as the prefix, even if the image Name didn't even specify the registry. This would cause the controller handler to re-resolve the image and potentially locate a new one even though the user never asked for the image to change.

pkg/client/app.go
```diff
@@ -106,7 +106,8 @@ func ToAppUpdate(ctx context.Context, c Client, name string, opts *AppUpdateOpti
 		return app, nil
 	}
 
-	if opts.Image != "" {
+	// only change app.Spec.Image if the user specified a different image
+	if opts.Image != "" && opts.Image != app.Spec.Image {
```
g-linville, line 110: This if statement contains the logic that finds a new image if the user specified an image for the update. In this case, we only want to do this if an image is defined (`opts.Image != ""`) and the image defined is not the same as the one currently in the app spec (`opts.Image != app.Spec.Image`). Only then do we go looking for the new image and change the app spec based on it.

pkg/controller/appdefinition/pullappimage.go
```diff
@@ -58,12 +58,16 @@ func pullAppImage(transport http.RoundTripper, client pullClient) router.Handler
...
 		)
 		if !appInstance.Status.AvailableAppImageRemote {
-			resolved, _, err = client.resolve(req.Ctx, req.Client, appInstance.Namespace, target)
+			resolved, isLocal, err = client.resolve(req.Ctx, req.Client, appInstance.Namespace, target)
 			if err != nil {
 				cond.Error(err)
 				return nil
+			} else if !isLocal {
+				// force pull from remote, since the only local image we found was marked remote, and there might be a newer version
+				resolved = target
```
g-linville, line 70: Here, `target` is the name of the image we are trying to pull. If we happen to find an Image API object that is marked with `Remote: true`, then instead of taking that image, we want to set `resolved` back to `target` so that we pull again from the remote registry, in case the image has been updated.

pkg/controller/appdefinition/pullappimage.go
```diff
@@ -58,12 +58,16 @@ func pullAppImage(transport http.RoundTripper, client pullClient) router.Handler
...
+			isLocal  bool
 		)
 		if !appInstance.Status.AvailableAppImageRemote {
-			resolved, _, err = client.resolve(req.Ctx, req.Client, appInstance.Namespace, target)
+			resolved, isLocal, err = client.resolve(req.Ctx, req.Client, appInstance.Namespace, target)
 			if err != nil {
 				cond.Error(err)
 				return nil
+			} else if !isLocal {
+				// force pull from remote, since the only local image we found was marked remote, and there might be a newer version
```
njhale, line 69: nit: I haven't been able to track down explicit documentation about this, but I'm fairly sure the rule-of-thumb for Go commentary -- except above declarations -- is to capitalize the first letter; at least, every major Go project I've seen seems to follow that convention (e.g. [k8s](https://github.com/kubernetes/kubernetes/blob/b63700630231803967f85e83ff35d719db511baa/pkg/controller/deployment/sync.go#L55))

```suggestion
				// Force pull from remote, since the only local image we found was marked remote, and there might be a newer version
```

iwilltry42 (approved):

PR #1791, opened 2023-06-14
Put project region information on status

thedadams:
Previously, the default region on status was only set if the user didn't specify a default region for the project. As soon as the user specified a default region, the field on the project's status would be cleared.

This behavior meant that the controllers and the UI would have to check the spec and the status to determine default and supported regions. Not ideal.

After this change, the project's status will be populated with the default and supported regions regardless of whether the user set them. If the user set them on the spec, then these values will simply be copied to the status. The goal is that the controllers and the UI now have one place to check for a default and supported regions: the status of the project.


njhale (approved):

g-linville (approved):

njhale (approved):
```
#3 ERROR: failed to copy: httpReadSeeker: failed open: unexpected status code https://registry-1.docker.io/v2/library/nginx/manifests/sha256:a11239a63fc962c2ef5b8d528c54b91c28ac566b7899aa626bb085e58a70202c: 429 Too Many Requests - Server message: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating
```

unrelated test failure, lgtm

tylerslaton (approved):

njhale (approved):

PR #1796, opened 2023-06-15
add: Sign and Verify Images from the CLI

iwilltry42:
This PR changed quite a lot due to lots of discussions around UX, etc.

Here's the current state (20.07.2023):

## The CLI-side

- new commands: `acorn image sign` & `acorn image verify` to sign and verify local/remote images
- new flags:
  - `acorn image pull --verify` to only pull an image if verification passed (extra flags `--key` and `--annotation`)
  - `acorn image push --sign` to sign an image and push it together with the new signature (extra flags `--key` and `--annotation`)

## The Server Side

- `ImageDetails` now contain the signature digest if we could find a signature for an image
- new `image/sign` and `image/verify` subresources, where we never pass in the private key, only public keys
- Verification (from CLI or as part of IARs) now support `gh://` and `ac://` identifiers to pull public keys from GitHub and Acorn Manager user accounts for verification
- `ImagePush` and `ImagePull` now take care of the signature as well if there is one
- `ImageCopy` and `RepoCopy` now both consider signatures as well
- `ImageDelete` cleans up signatures from the registry

### Answered

- Q: Do we want to store/cache one of the following in-cluster?
  - a) Public Keys, e.g. fetched from GitHub - to lower required API requests (e.g. cache keys that matched before, as they're more likely to be used again and only fetch new ones if that one didn't match)
    - A/a: 21.06.2023 (Craig): later
  - b) Signatures as created by `acorn image sign` (would allow for local-only signatures, e.g. I pull an image from DockerHub, where I cannot push my signature back, but I just sign it directly in-cluster using `acorn image sign` to allow it to be run)
    - A/b: 21.06.2023 (Craig): for now, we'll stick with the standard of pushing to remote registries only
  - We are **not** going to manage private keys in Manager
  - Potentially, we can implement "keyless" signing using ephemeral keypairs, where we use the private key once for signing and only record the public key, thus connecting the Hub identity to the signature
    - In that case, users must be enabled to manage their history of signatures/public keys
    - This does not require a CA or TL (like Rekor), meaning less management burden
    - For now we'll focus on client-side signing only with existing user keys

### Checklist


g-linville:
> What about acorn pull --verify?, which would only pull if verification passed? In this case we could combine it with checking IARs in the target project as well.

If it's not too much work to implement, I think this would be nice for the user. I assume (haven't tested it yet) that `acorn verify` works with images in remote registries, but this would be a handy way to verify and pull with just one command.

pkg/cosign/manager.go
```diff
@@ -0,0 +1,77 @@
...
+
+func getAcornPublicKeys(username string) ([]AcornPublicKey, error) {
+	logrus.Debugf("Getting public keys for Acorn user %s", username)
+
+	parts := strings.Split(username, "@")
+	if len(parts) > 2 {
+		return nil, fmt.Errorf("invalid username: %s", username)
+	}
+
+	host := "https://hub.acornlabs.com"
```
g-linville, line 27: nit: I think this should be

```suggestion
	host := "https://acorn.io"
```

pkg/cli/images_sign.go
```diff
@@ -0,0 +1,165 @@
...
+		ValidArgsFunction: newCompletion(c.ClientFactory, imagesCompletion(true)).complete,
+		Args:              cobra.ExactArgs(1),
+	})
+	_ = cmd.MarkFlagFilename("key")
+	return cmd
+}
+
+type ImageSign struct {
+	client      ClientFactory
+	Key         string            `usage:"Key to use for signing" short:"k" local:"true" default:"./cosign.key"`
```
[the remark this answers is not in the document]
iwilltry42: It defines whether a flag is "local" to this command or a persistent flag (see https://github.com/acorn-io/runtime/blob/main/pkg/cli/builder/builder.go#L122-L125)

pkg/cli/images_verify.go
```diff
@@ -0,0 +1,107 @@
...
+
+	target := ref.Context().Digest(targetDigest)
+
+	if a.Key == "" {
+		return fmt.Errorf("key is required")
+	}
+	pterm.Info.Printf("Verifying Image %s (digest: %s) using key %s\n", targetName, targetDigest, a.Key)
+
+	annotationRules := internalv1.SignatureAnnotations{
+		Match: a.Annotations,
```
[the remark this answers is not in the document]
iwilltry42: Yup, no problem there

pkg/cosign/cosign.go
```diff
@@ -334,20 +383,92 @@ var algorithms = map[string]crypto.Hash{
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to load public key from SSH: %w", err)
+		}
+		verifiers = append(verifiers, v)
+	} else if strings.HasPrefix(keyRef, "-----BEGIN") {
+		key, err := ParseSSHPublicKey(keyRef)
+		if err != nil {
+			return nil, err
+		}
+		logrus.Infof("Parsed key: %#v", key)
```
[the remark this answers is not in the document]
iwilltry42: Yeah, that makes sense.. and also adjust log level to debug in most places.

pkg/cosign/manager.go
iwilltry42: Absolutely!

pkg/cli/images_sign.go
```diff
@@ -0,0 +1,165 @@
...
+
+	pf := func(_ bool) ([]byte, error) {
+		return pass, nil
+	}
+
+	// Get a sigSigner-verifier from a private key and if the key type is not supported, try to import it first
+	var sigSigner sigsig.SignerVerifier
+	sigSigner, err = signature.SignerVerifierFromKeyRef(cmd.Context(), a.Key, pf) // TODO(iwilltry42): use our own style password prompt
+	if err != nil {
+		if !strings.Contains(err.Error(), "unsupported pem type") {
```
g-linville, line 99: Is there any way we can cast the error to a specific error type and check that way instead of checking the error string? Ordinarily I am not against this approach if it is in our own code, but since this is coming from a library, we don't know when they might change the wording of the error message for some reason.

g-linville (approved):
One other question - what does `acorn image sign` do if the user does not specify the push option? I assume it is not pushing signatures into the internal registry, since the CLI has no direct access to the internal registry. It is it basically a dry run?

pkg/cli/images_sign.go
```diff
@@ -0,0 +1,165 @@
...
+		return err
+	}
+
+	pf := func(_ bool) ([]byte, error) {
+		return pass, nil
+	}
+
+	// Get a sigSigner-verifier from a private key and if the key type is not supported, try to import it first
+	var sigSigner sigsig.SignerVerifier
+	sigSigner, err = signature.SignerVerifierFromKeyRef(cmd.Context(), a.Key, pf) // TODO(iwilltry42): use our own style password prompt
```
thedadams, line 97: Can we pass `generate.GetPass` directly here instead of this closure?

pkg/cli/images_sign.go
```diff
@@ -0,0 +1,165 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	dupeDetector := remote.NewDupeDetector(sigSigner)
+
+	var annotations map[string]interface{}
+	if a.Annotations != nil {
+		annotations = make(map[string]interface{})
```
thedadams, line 117:
```suggestion
		annotations = make(map[string]interface{}, len(a.Annotations))
```

pkg/cli/images_sign.go
```diff
@@ -0,0 +1,165 @@
...
+		for k, v := range a.Annotations {
+			annotations[k] = v
+		}
+	}
+
+	payload, signature, err := sigsig.SignImage(sigSigner, target, annotations)
```
thedadams, line 123: Is it no OK to just pass `a.Annotations` here directly?

pkg/cli/images_sign.go
```diff
@@ -0,0 +1,165 @@
...
+	"github.com/sigstore/cosign/v2/pkg/oci/static"
+	"github.com/sigstore/cosign/v2/pkg/signature"
+	sigsig "github.com/sigstore/sigstore/pkg/signature"
+	"github.com/spf13/cobra"
+)
+
+func NewImageSign(c CommandContext) *cobra.Command {
+	cmd := cli.Command(&ImageSign{client: c.ClientFactory}, cobra.Command{
+		Use:               "sign IMAGE_NAME [flags]",
+		Example:           `acorn image sign my-image --key ./my-key`,
```
thedadams, line 29: Is it possible to pass a GitHub key here? If so, can we add another example to illustrate this?

pkg/cli/images_verify.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/acorn-io/runtime/pkg/images"
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/pterm/pterm"
+	"github.com/spf13/cobra"
+)
+
+func NewImageVerify(c CommandContext) *cobra.Command {
+	cmd := cli.Command(&ImageVerify{client: c.ClientFactory}, cobra.Command{
+		Use:               "verify IMAGE_NAME [flags]",
+		Example:           `acorn image verify my-image --key ./my-key.pub`,
```
thedadams, line 20: Same question here about passing GitHub keys.

pkg/cosign/manager.go
```diff
@@ -0,0 +1,77 @@
...
+	// token := "g4xc25zkb6b2dkld5fjwvqf6p9lns6lp2h9rbctrqtqf7cb6fr75rd"
+	// req.Header.Set("Authorization", "Bearer "+token)
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode > 400 {
```
thedadams, line 50: Just in case?
```suggestion
	if resp.StatusCode >= 400 {
```

pkg/cli/images_sign.go
iwilltry42: gh://, ac://, etc. are only supported for verification, where we need public keys

pkg/cli/images_sign.go
```diff
@@ -0,0 +1,165 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	target := ref.Context().Digest(targetDigest)
+
+	if a.Key == "" {
+		return fmt.Errorf("key is required")
+	}
```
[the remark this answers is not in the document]
iwilltry42: Fair point.

pkg/cli/images_sign.go
iwilltry42: We're using `pf`, which just returns the "cached" password entered once to avoid potentially asking for the password twice if we need to to the intermediate import first (a few lines below)

pkg/cli/images_sign.go
iwilltry42: Unfortunately they don't have a specific error type for this: https://github.com/sigstore/cosign/blob/main/pkg/cosign/keys.go#L216

pkg/cli/images_sign.go
iwilltry42: Unforatunately not, because down the road we need `map[string]interface{}`, not `map[string]string`.
Alternatively I could make the flag param `map[string]any`, but that would also require new logic in the CLI builder to support that type, so I thought this might be easier.

pkg/cli/images_verify.go
iwilltry42: True here. Will add some examples.

pkg/cosign/cosign.go
```diff
@@ -173,15 +195,42 @@ func VerifySignature(ctx context.Context, opts VerifyOpts) error {
...
+	}
+
+	verified := false
+	var errs []error
+	for _, v := range opts.Verifiers {
+		cosignOpts.SigVerifier = v
+		err := verifySignature(ctx, sigs, imgDigestHash, opts, cosignOpts)
+		if err == nil {
+			verified = true
+			break
```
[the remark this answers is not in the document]
iwilltry42: That's true, I'll clean that up.

pkg/cosign/cosign.go
```diff
@@ -173,15 +195,42 @@ func VerifySignature(ctx context.Context, opts VerifyOpts) error {
 		IgnoreTlog:         true,
 	}
 
+	if opts.Verifiers == nil {
+		opts.Verifiers = []signature.Verifier{}
+	}
+
```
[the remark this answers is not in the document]
iwilltry42: I just can't get that out of my head :D

pkg/cosign/keys.go
```diff
@@ -0,0 +1,92 @@
...
+	"encoding/hex"
+	"encoding/pem"
+	"fmt"
+	"regexp"
+
+	"golang.org/x/crypto/ssh"
+)
+
+var (
+	supportedKeyTypes = map[string]interface{}{
```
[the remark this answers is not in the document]
iwilltry42: Optimize!

pkg/cosign/manager.go
iwilltry42: Oh Absolutely!

pkg/cli/images_sign.go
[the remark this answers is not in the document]
iwilltry42: It feels weird to have them as PersistentFlags, even though it doesn't change anything here :grimacing:

iwilltry42:
> One other question - what does `acorn image sign` do if the user does not specify the push option? I assume it is not pushing signatures into the internal registry, since the CLI has no direct access to the internal registry. It is it basically a dry run?

@g-linville Yeah it's basically a dry-run. We could think about outputting the signature artifact e.g. to the filesystem.

pkg/cli/images_sign.go
njhale: > Alternatively I could make the flag param map[string]any, but that would also require new logic in the CLI builder to support that type

Hmm, I wouldn't have expected new logic because IIUC `any` is just a type alias for `interface{}`.

pkg/cosign/cosign.go
```diff
@@ -175,13 +197,30 @@ func VerifySignature(ctx context.Context, opts VerifyOpts) error {
...
+		err := verifySignature(ctx, sigs, imgDigestHash, opts, cosignOpts)
+		if err == nil {
+			return nil
+		}
+		errs = append(errs, err)
 	}
 
+	err = cosign.NewVerificationError("failed to find valid signature for %s matching given identity and annotations using %d loaded verifiers/keys", opts.ImageRef.String(), len(opts.Verifiers))
+	err.(*cosign.VerificationError).SetErrorType(cosign.ErrNoMatchingSignaturesType)
+	logrus.Debugf("%s: %v", err, merr.NewErrors(errs...))
```
njhale, line 219: I _think_ you can use the standard library for this now; see https://pkg.go.dev/errors#Join

pkg/cli/images_sign.go
njhale: https://goplay.tools/snippet/rrynQWaJ1JV -- but maybe I'm misunderstanding what @iwilltry42 meant by

> I could make the flag param map[string]any

iwilltry42: @njhale Currently `a.Annotations` is `map[string]string` and I'm doing exactly what you're doing in the playground snippet (converting to `map[string]any` in a loop).

`sigsig.SignImage` accepts `map[string]any`, that's the reason for the conversion.
To use `a.Annotations` directly there it'd need to be `map[string]any` right away, which would require another implementation of `assignMaps` (https://github.com/acorn-io/runtime/blob/main/pkg/cli/builder/builder.go#L288)

pkg/cosign/cosign.go
iwilltry42: Absolutely!

iwilltry42:
> > What about acorn pull --verify?, which would only pull if verification passed? In this case we could combine it with checking IARs in the target project as well.
> 
> If it's not too much work to implement, I think this would be nice for the user. I assume (haven't tested it yet) that `acorn verify` works with images in remote registries, but this would be a handy way to verify and pull with just one command.
> 
> I also have a PR out for `acorn copy`. Once this stuff is in, I might go add some signature verification options to that as well.

@g-linville  here you go:

```bash
$ acorn pull --verify --key gh://iwilltry42 'my.reg.io/public/acorn/hello-world:v1.0.0'
  •  Verifying Image my.reg.io/public/acorn/hello-world:v1.0.0 (digest: sha256:1a6c64d2ccd0bb035f9c8196d3bfe72a7fdbddc4530dfcb3ab2a0ab8afb57eeb) using key gh://iwilltry42
  ✔  Signature verified
 [112231571/112231571] ███████████████████████████████████████████████ 100% | 2s

$ acorn pull --verify --key gh://g-linville 'my.reg.io/public/acorn/hello-world:v1.0.0'
  •  Verifying Image my.reg.io/public/acorn/hello-world:v1.0.0 (digest: sha256:1a6c64d2ccd0bb035f9c8196d3bfe72a7fdbddc4530dfcb3ab2a0ab8afb57eeb) using key gh://g-linville
  ✗  ERROR:  NOT pulling image: failed to find valid signature for my.reg.io/public/acorn/hello-world@sha256:1a6c64d2ccd0bb035f9c8196d3bfe72a7fdbddc4530dfcb3ab2a0ab8afb57eeb matching given identity and annotations using 1 loaded verifiers/keys
```

This is **not** checking against IARs right now

g-linville:
@iwilltry42 I have another question. Does `acorn image sign --push` ever push signatures to the local registry, or can it only push to the remote registry? If it can only push to remote registries, is there any point to running `acorn image verify` on a local image?

iwilltry42:
@g-linville if I understood @ibuildthecloud correctly, we will only ever want signatures in a remote registry, not in the internal one.

g-linville:
@iwilltry42 in that case I think we should never check the local registry for `acorn image verify` since there will never be any signatures in there to verify anyway.

pkg/cli/images_verify.go
```diff
@@ -0,0 +1,116 @@
...
+
+	targetName := args[0]
+	targetDigest := ""
+
+	c, err := a.client.CreateDefault()
+	if err != nil {
+		return err
+	}
+
+	img, tag, err := client.FindImage(cmd.Context(), c, args[0])
```
thedadams, line 57: nit:
```suggestion
	img, tag, err := client.FindImage(cmd.Context(), c, targetName)
```

thedadams (approved):

iwilltry42:
Another change incoming as per yesterday's discussion:

## Image Signatures in Acorn - Verification

- Piggyback on `ImageDetails`
- Pull signature together with image (do not auto-refresh later - keep only signature from pull-time)
- `acorn image verify` will use the local image and signature if exists, else remote only

## Image Signing

- client-side: only create signature payload, so we're not sending the private key
- mutate the OCI entity server-side
- keep signature in internal registry until the image is pushed (then push signature with it)

g-linville:
@iwilltry42 I am not seeing these changes. Are there more commits that need to be pushed?

pkg/server/registry/apigroups/acorn/images/sign.go
```diff
@@ -0,0 +1,129 @@
...
+
+	if isig.Name == "" {
+		ri, ok := request.RequestInfoFrom(ctx)
+		if ok {
+			isig.Name = ri.Name
+		}
+	}
+	ns, _ := request.NamespaceFrom(ctx)
+	opts := []remote.Option{t.transportOpt}
+	if isig.Auth != nil {
```
g-linville, line 56: Does `Auth` ever get set on this type? I don't think I saw that happen anywhere.

g-linville: Also I don't think we normally pass registry credentials from the client to the apiserver like this. If you look at most of our image operations, we use functions like `images.GetAuthenticationRemoteOptionsWithLocalAuth` to get the credentials.

pkg/server/registry/apigroups/acorn/images/sign.go
```diff
@@ -0,0 +1,129 @@
...
+		verifier, err := acornsign.DecodePEM(signature.PublicKey, crypto.SHA256)
+		if err != nil {
+			return "", err
+		}
+		dupeDetector := cremote.NewDupeDetector(verifier)
+
+		mutateOpts = append(mutateOpts, mutate.WithDupeDetector(dupeDetector))
+	}
+
+	ref, err := images.GetImageReference(ctx, t.client, namespace, targetName)
```
g-linville, line 98: `name.ParseReference` will probably suffice here. `images.GetReference` is supposed to look for local, nameless images by their digest hash, which is probably not what we want happening here.

pkg/server/registry/apigroups/acorn/images/sign.go
iwilltry42: Yeah, it happens on line 60 :grimacing: 
Yeah I used to use `images.GetAuthenticationRemoteOptionsWithLocalAuth` elsewehere, but somehow I ended up copying this over from ImageDetails.
I may want to dive into the differences of those variations. :thinking:

pkg/server/registry/apigroups/acorn/images/sign.go
iwilltry42: I think `images.GetReference` is the best option here, since it calls `GetInternalRepoForNamespaceAndID` for local images, so I get the full registry/repo location for either external or local images.
That's part of the change that we push signatures for local images to the local registry.

pkg/cli/push.go
```diff
@@ -57,5 +61,22 @@ func (s *Push) Run(cmd *cobra.Command, args []string) error {
...
+			client:      s.client,
+			Key:         s.Key,
+			Annotations: s.SignatureAnnotations,
+		}
+		if err := sign.Run(cmd, args); err != nil {
+			return err
+		}
+
+		pterm.Success.Printf("Pushed and signed %s\n", args[0])
+	}
```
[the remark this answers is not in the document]
iwilltry42: Oh good catch. I forgot to revamp the image push logic :facepalm:

pkg/server/registry/apigroups/acorn/images/sign.go
iwilltry42: Nevermind, found my mistake :man_facepalming:

pkg/cli/push.go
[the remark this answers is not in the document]
iwilltry42: :heavy_check_mark:

iwilltry42:
@g-linville & @thedadams & @njhale sorry for the hassle around this damn PR.
It went back and forth quite often.
I think I have all the requested changes and adjustments in now and pushed some tests.

I hope this was the final round of changes and I also updated the PR description to match the latest changes.

Thanks for your previous reviews :pray:

pkg/cosign/manager.go
```diff
@@ -0,0 +1,77 @@
...
+
+	url := fmt.Sprintf("%s/v1/manager.acorn.io.publickeys/%s/", host, username)
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: maybe use token if user is logged in, so that we could rate-limit anonymous requests?
+	// token := "g4xc25zkb6b2dkld5fjwvqf6p9lns6lp2h9rbctrqtqf7cb6fr75rd"
+	// req.Header.Set("Authorization", "Bearer "+token)
```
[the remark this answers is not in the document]
iwilltry42: We decided not to do this for now.

pkg/cli/images_verify.go
```diff
@@ -0,0 +1,121 @@
...
+	imageName := args[0]
+
+	c, err := a.client.CreateDefault()
+	if err != nil {
+		return err
+	}
+
+	var auth *apiv1.RegistryAuth
+	ref, err := name.ParseReference(imageName)
+	if err == nil { // not failing here, since it could be a local image
```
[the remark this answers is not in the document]
iwilltry42: Good catch!
Actually with the new logic of having the signature in the internal registry, both should work for tagless local images.
@ibuildthecloud voted for overriding the signature when pushing in general, so here, we would merge the new signature into the existing one in the internal registry and upon pushing the image, we would override the externally stored signature.

I'll adjust the code accordingly.

pkg/cosign/keys.go
```diff
@@ -0,0 +1,110 @@
...
+		pub, err := x509.ParsePKIXPublicKey(block.Bytes)
+		if err != nil {
+			return nil, err
+		}
+		switch key := pub.(type) {
+		case *rsa.PublicKey:
+			return key, nil
+		case *ecdsa.PublicKey:
+			return key, nil
+		case ed25519.PublicKey:
```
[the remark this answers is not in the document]
iwilltry42: :heavy_check_mark:

pkg/cosign/registry.go
```diff
@@ -66,3 +68,27 @@ func makeOptions(opts ...crane.Option) crane.Options {
 	}
 	return opt
 }
+
+func FindSignature(imageDigest name.Digest, opts ...remote.Option) (name.Tag, ggcrv1.Hash, error) {
+	ociremoteOpts := []ociremote.Option{ociremote.WithRemoteOptions(opts...)}
+
+	var tag name.Tag
+	var hash ggcrv1.Hash
+	var err error
```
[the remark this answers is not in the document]
iwilltry42: :heavy_check_mark:

pkg/cli/images_verify.go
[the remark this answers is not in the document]
iwilltry42: Updated :heavy_check_mark:

g-linville (approved):
Tested out the new changes and it works great!

pkg/cli/push.go
```diff
@@ -21,7 +22,10 @@ func NewPush(c CommandContext) *cobra.Command {
 }
 
 type Push struct {
-	client ClientFactory
+	client               ClientFactory
+	Sign                 bool              `usage:"Sign the image before pushing" short:"s" local:"true" default:"false"`
+	Key                  string            `usage:"Key to use for signing" short:"k" local:"true" default:"./cosign.key"`
+	SignatureAnnotations map[string]string `usage:"Annotations to add to the signature" short:"a" local:"true"`
```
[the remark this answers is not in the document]
iwilltry42:
```suggestion
	SignatureAnnotations map[string]string `usage:"Annotations to add to the signature" short:"a" local:"true" name:"signature-annotation"`
```

integration/client/images/images_test.go
```diff
@@ -336,6 +343,136 @@ func TestImageDetails(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	pem, _, err := acornsign.PemEncodeCryptoPublicKey(pubkey)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if pubkey != nil {
```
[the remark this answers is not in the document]
iwilltry42: It doesn't actually matter here, as it's optional and used for "clean" signature deduplication when pushing the signature to the registry.

pkg/apis/api.acorn.io/v1/types.go
```diff
@@ -223,10 +223,11 @@ type ImageDetails struct {
...
 
 	// Output Params
-	AppImage   v1.AppImage   `json:"appImage,omitempty"`
-	AppSpec    *v1.AppSpec   `json:"appSpec,omitempty"`
-	Params     *v1.ParamSpec `json:"params,omitempty"`
-	ParseError string        `json:"parseError,omitempty"`
+	AppImage        v1.AppImage   `json:"appImage,omitempty"`
+	AppSpec         *v1.AppSpec   `json:"appSpec,omitempty"`
+	Params          *v1.ParamSpec `json:"params,omitempty"`
+	SignatureDigest string        `json:"signatureDigest,omitempty"`
```
tylerslaton, line 229: Should this be apart of the `AppImage` struct such that I can reference `imageDetails.AppImage.SignatureDigest`?

pkg/cli/images_detail.go
```diff
@@ -66,8 +67,16 @@ func (a *ImageDetails) Run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	type ImageDetailsOutput struct {
+		internalv1.AppImage
+		SignatureDigest string `json:"signature,omitempty" yaml:"signature,omitempty"`
```
tylerslaton, line 72: If we do end up putting the `SignatureDigest` into the `AppImage` struct this change will likely become easier

pkg/cli/images_rm.go
```diff
@@ -43,7 +42,7 @@ func (a *ImageDelete) Run(cmd *cobra.Command, args []string) error {
 		}
 
 		if img == nil && len(removedTags) == 0 {
-			logrus.Infof("No image found for %s", image)
+			fmt.Printf("No image found for %s", image)
```
tylerslaton, line 45: Why the switch from logrus?

pkg/cli/pull.go
```diff
@@ -48,6 +53,17 @@ func (s *Pull) Run(cmd *cobra.Command, args []string) error {
...
 	}
 
+	if s.Verify {
+		v := ImageVerify{
+			client:      s.client,
+			Key:         s.Key,
+			Annotations: s.Annotations,
+		}
+		if err := v.Run(cmd, args); err != nil {
+			return fmt.Errorf("NOT pulling image: %w", err)
```
tylerslaton, line 63: nit:
```suggestion
			return fmt.Errorf("NOT pulling image due to verification issue: %w", err)
```

pkg/apis/api.acorn.io/v1/types.go
iwilltry42: I think we aligned on treating signatures as separate entities completely, which are just referencing AppImages by Digest. At least @ibuildthecloud wanted to have them only coupled on the ImageDetails level.

integration/client/images/images_test.go
```diff
@@ -13,6 +13,8 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/stretchr/testify/assert"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+
+	_ "embed"
```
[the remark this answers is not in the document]
iwilltry42: Nope, not required. Leftover from when all tests were in one file. Good catch!

pkg/apis/api.acorn.io/v1/types.go
tylerslaton: Ah okay. That does make sense. The only benefit that I saw is that some parts of the code become a bit simpler. However, if we want to treat them differently then I don't think that this is something we want to change.

pkg/cli/images_rm.go
iwilltry42: fmt all around here, since it's CLI output

tylerslaton: Ahhh I see!

tylerslaton (approved):

pkg/server/registry/apigroups/acorn/images/push.go
```diff
@@ -162,12 +156,73 @@ func (i *ImagePush) ImagePush(ctx context.Context, image *apiv1.Image, tagName s
...
+					errString = update.Error.Error()
+				}
+				progress <- ImageProgress{
+					Total:       update.Total,
+					Complete:    update.Complete,
+					Error:       errString,
+					CurrentTask: fmt.Sprintf("Pushing %s %s ", c.sourceName, c.destTag.String()),
+				}
+			}
+		}
```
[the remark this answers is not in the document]
iwilltry42: Done: [`c939dab` (#1796)](https://github.com/acorn-io/runtime/pull/1796/commits/c939dab7f4bcb34c30f5e7f2a956b75b47b81f57)

PR #1797, opened 2023-06-15
Clarify project update validation message

njhale:
- Clarify which resources need to be cleaned up before validation will pass
when regions are removed from a project
- Fall back on scheme to set resource kind in error message

e.g. of message after this change

```
Project.api.acorn.io "demo" is invalid: spec.supportedRegions: Invalid value: []string{"my-us-east-2"}: cannot remove regions [acorn-us-east-2] while in use by the following apps: [my-db.cdk-context, my-db]
```

### DIFF
--- a/pkg/server/registry/apigroups/acorn/projects/validator.go
+++ b/pkg/server/registry/apigroups/acorn/projects/validator.go
@@ -42,14 +42,14 @@ func (v *Validator) Validate(_ context.Context, obj runtime.Object) field.ErrorL
 	return nil
 }
 
-func (v *Validator) ValidateUpdate(ctx context.Context, new, old runtime.Object) field.ErrorList {
+func (v *Validator) ValidateUpdate(ctx context.Context, obj, old runtime.Object) field.ErrorList {
 	// Ensure that default region and supported regions are valid.
-	if err := v.Validate(ctx, new); err != nil {
+	if err := v.Validate(ctx, obj); err != nil {
 		return err
 	}
 
 	// If the user is removing a supported region, ensure that there are no apps in that region.
-	oldProject, newProject := old.(*apiv1.Project), new.(*apiv1.Project)
+	oldProject, newProject := old.(*apiv1.Project), obj.(*apiv1.Project)
 	var removedRegions []string
 	for _, region := range append(oldProject.Spec.SupportedRegions, oldProject.Status.DefaultRegion) {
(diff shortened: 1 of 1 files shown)


pkg/server/registry/apigroups/acorn/projects/validator.go
```diff
@@ -87,14 +87,28 @@ func (v *Validator) ensureNoObjectsExistInRegions(ctx context.Context, namespace
 				field.NewPath("spec", "supportedRegions"),
 				regions,
 				fmt.Sprintf(
-					"cannot remove regions %v that have %s: %v",
+					"cannot remove regions %v while in use by the following %ss: %v",
 					regions,
-					strings.TrimSuffix(strings.ToLower(obj.GetObjectKind().GroupVersionKind().Kind), "list"),
```
njhale, line 92: If the `obj` passed here has no GVK set on it's typemeta -- which is what we're doing -- then this evaluates to an empty string.

pkg/server/registry/apigroups/acorn/projects/validator.go
```diff
@@ -42,14 +42,14 @@ func (v *Validator) Validate(_ context.Context, obj runtime.Object) field.ErrorL
 	return nil
 }
 
-func (v *Validator) ValidateUpdate(ctx context.Context, new, old runtime.Object) field.ErrorList {
```
njhale, line 45: I think we should avoid shadowing builtins

pkg/server/registry/apigroups/acorn/projects/validator.go
```diff
@@ -58,43 +58,57 @@ func (v *Validator) ValidateUpdate(ctx context.Context, new, old runtime.Object)
...
 			return nil
 		}); err != nil {
 			return field.ErrorList{field.InternalError(field.NewPath("spec", "supportedRegions"), err)}
 		}
 
-		if len(objectsInRemoveRegions) > 0 {
+		if len(inRemovedRegion) > 0 {
 			result = append(result, field.Invalid(
 				field.NewPath("spec", "supportedRegions"),
 				regions,
```
njhale, line 88: `field.Invalid` takes the _new_ field value as an argument here; before we were feeding it the removed regions.

thedadams (approved):

g-linville (approved):

PR #1799, opened 2023-06-15
Refactor Resources struct and its helper functions

tylerslaton:
Donnie brought up a good point about why these resources aren't represented in the runtime `Resources` struct. I am adding those and also making some UX improvements for them.
### Checklist

### DIFF
--- a/pkg/apis/internal.admin.acorn.io/v1/resources.go
+++ b/pkg/apis/internal.admin.acorn.io/v1/resources.go
@@ -43,19 +43,28 @@ func (current *Resources) Add(incoming Resources) {
 
 // Remove will remove the resources of another Resources struct from the current one.
 func (current *Resources) Remove(incoming Resources, all bool) {
-	current.Apps -= incoming.Apps
-	current.Containers -= incoming.Containers
-	current.Jobs -= incoming.Jobs
-	current.Volumes -= incoming.Volumes
-	current.Images -= incoming.Images
-	current.Projects -= incoming.Projects
+	// Do not allow resources to go below 0
+	nonNegativeSubtract := func(currentVal, incomingVal int) int {
+		difference := currentVal - incomingVal
+		if difference < 0 {
+			difference = 0
+		}
+		return difference
+	}
+
+	current.Apps = nonNegativeSubtract(current.Apps, incoming.Apps)
+	current.Containers = nonNegativeSubtract(current.Containers, incoming.Containers)
+	current.Jobs = nonNegativeSubtract(current.Jobs, incoming.Jobs)
+	current.Volumes = nonNegativeSubtract(current.Volumes, incoming.Volumes)
+	current.Images = nonNegativeSubtract(current.Images, incoming.Images)
+	current.Projects = nonNegativeSubtract(current.Projects, incoming.Projects)
 
 	current.Memory.Sub(incoming.Memory)
 	current.CPU.Sub(incoming.CPU)
 
 	// Only remove persistent resources if all is true.
 	if all {
-		current.Secrets -= incoming.Secrets
+		current.Secrets = nonNegativeSubtract(current.Secrets, incoming.Secrets)
 		current.VolumeStorage.Sub(incoming.VolumeStorage)
 	}
 }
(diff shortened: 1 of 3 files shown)


pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -60,87 +65,85 @@ func (current *Resources) Remove(incoming Resources, all bool) {
...
+	checkResource("Containers", current.Containers, incoming.Containers)
+	checkResource("Jobs", current.Jobs, incoming.Jobs)
+	checkResource("Volumes", current.Volumes, incoming.Volumes)
+	checkResource("Secrets", current.Secrets, incoming.Secrets)
+	checkResource("Images", current.Images, incoming.Images)
+	checkResource("Projects", current.Projects, incoming.Projects)
+
+	checkQuantityResource("VolumeStorage", current.VolumeStorage, incoming.VolumeStorage)
+	checkQuantityResource("Memory", current.Memory, incoming.Memory)
+	checkQuantityResource("Cpu", current.CPU, incoming.CPU)
```
njhale, line 93: Mentioned in slack -- here's an alternative (take it or leave it):

```go
func (current *Resources) Fits(incoming Resources) error {
    // ...
	for _, cmp := range []struct{
	    resource string
	    limit, incoming any
	}{
	  {"Apps", current.Apps, incoming.Apps},
	  // ...
	  {"Cpu", current.CPU, incoming.CPU},
	}{
	   if gt(cmp.incoming, cmp.limit) {
	       exceeded = append(exceeded, cmp.resource)
	   }
	}
	//...
}

// gt returns true IFF a is greater than b.
func [T any]gt(a, b T) bool {
   switch v := a.(type) {
   case *resource.Quantity:
     return v.Cmp(b.(resource.Quantity)) > 0
   case int:
     return v > b.(int)
   }

   return false
}
```

njhale: you could also do something like `runtime.Scheme`, where -- using reflection -- you register comparator functions for different types, then you wouldn't need a type switch

anyway, this is a nice exercise, but I don't feel that any of these approaches are _more_ understandable or practical than the approach being replaced by this PR.

njhale (approved):

thedadams (approved):

pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -60,87 +65,85 @@ func (current *Resources) Remove(incoming Resources, all bool) {
...
-	if current.Volumes > 0 {
-		resources = append(resources, fmt.Sprintf("Volumes:%v", current.Volumes))
-	}
-	if current.Secrets > 0 {
-		resources = append(resources, fmt.Sprintf("Secrets:%v", current.Secrets))
-	}
-	if current.Images > 0 {
-		resources = append(resources, fmt.Sprintf("Images:%v", current.Images))
-	}
+func (current *Resources) NonEmptyString() string {
```
g-linville, line 104: nit: function comment needs to be updated

pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -60,87 +65,85 @@ func (current *Resources) Remove(incoming Resources, all bool) {
...
+		if value > 0 {
+			resources = append(resources, resource)
+		}
 	}
-	if !current.CPU.IsZero() {
-		resources = append(resources, fmt.Sprintf("CPU:%v", current.CPU))
+
+	// Define function for checking quantity resources to keep code DRY
+	checkQuantityResource := func(resource string, currentVal resource.Quantity) {
+		if currentVal.IsZero() {
```
g-linville, line 116: I think this is supposed to be inverted:

```suggestion
		if !currentVal.IsZero() {
```

pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -31,25 +34,27 @@ func (current *Resources) Add(incoming Resources) {
...
+	current.Projects += incoming.Projects
 
 	current.VolumeStorage.Add(incoming.VolumeStorage)
 	current.Memory.Add(incoming.Memory)
 	current.CPU.Add(incoming.CPU)
 }
 
 // Remove will remove the resources of another Resources struct from the current one.
-func (current *Resources) Remove(incoming Resources, all bool) {
+func (current *Resources) Remove(incoming Resources, removeAll bool) {
```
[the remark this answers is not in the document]
tylerslaton: Fair enough, probably just me over adjusting.

pkg/apis/internal.admin.acorn.io/v1/resources.go
tylerslaton: Yep!

pkg/apis/internal.admin.acorn.io/v1/resources.go
tylerslaton: This is super interesting. I do think that what you said is right in that it is cool but isn't moving the needle of understanding. If you're okay with, let's get the PR in as is?

PR #1800, opened 2023-06-16
Drop built-in http01 challenge, use cert-manager to issue certs for custom domain

StrongMonkey:
### Context

We want to drop existing built-in http01 handler, instead, use exteral cert-manager and annoation to provision letsencrypt certs for custom domain. Detailed discussion here: https://acorn-io.slack.com/archives/C03FU91U4SX/p1686771966865679

This PR has the following changes:
1. Drop existing handler and code that deal with custom domain cert provisioning.
2. Setup proper cert-manager annotation to clusterIssuer to delegate certs provisioning to cert-manager.
3. Add an install option to allow specify the name of the cluster issuer.

https://github.com/acorn-io/hub/issues/315

### Checklist


pkg/apis/api.acorn.io/v1/types.go
```diff
@@ -424,6 +424,7 @@ type Config struct {
 	AWSIdentityProviderARN         *string         `json:"awsIdentityProviderArn" name:"aws-identity-provider-arn" usage:"ARN of cluster's OpenID Connect provider registered in AWS"`
 	EventTTL                       *string         `json:"eventTTL" name:"event-ttl" usage:"Amount of time an Acorn event will be stored before being deleted (default '168h' - 7 days)"`
 	Features                       map[string]bool `json:"features" name:"features" boolmap:"true" usage:"Enable or disable features. (example foo=true,bar=false)"`
+	CertManagerIssuer              *string         `json:"certManagerIssuer" name:"cert-manager-issuer" usage:"The name of the cert-manager cluster issuer to use for TLS certificates on custom domains (default letsencrypt-prod)" default:"letsencrypt-prod"`
```
iwilltry42, line 427:
```suggestion
	CertManagerIssuer              *string         `json:"certManagerIssuer" name:"cert-manager-issuer" usage:"The name of the cert-manager cluster issuer to use for TLS certificates on custom domains" default:"letsencrypt-prod"`
```

Dropping this here so we don't have duplicate defaults in the readme.

pkg/publish/ingress.go
```diff
@@ -176,74 +186,97 @@ func Ingress(req router.Request, svc *v1.ServiceInstance) (result []kclient.Obje
...
+	} {
+		if len(rules.rules) == 0 {
+			continue
+		}
+		var targets map[string]Target
+		if rules.name == clusterDomain {
+			targets = clusterDomainTargets
+		} else {
+			targets = customDomainTargets
+		}
```
iwilltry42, line 217: Couldn't you put targets in the struct you're looping over and drop this condition and use `rules.targets`?

pkg/publish/ingress.go
```diff
@@ -176,74 +186,97 @@ func Ingress(req router.Request, svc *v1.ServiceInstance) (result []kclient.Obje
...
-	result = append(result, secrets...)
+		result = append(result, secrets...)
+	}
 
 	return
 }
 
-func setupCertManager(serviceName string, annotations map[string]string, rules []networkingv1.IngressRule, tls []networkingv1.IngressTLS) []networkingv1.IngressTLS {
-	if annotations["cert-manager.io/cluster-issuer"] == "" && annotations["cert-manager.io/issuer"] == "" {
+func setupCertManager(serviceName string, annotations map[string]string, rules []networkingv1.IngressRule, tls []networkingv1.IngressTLS, useCertManager bool) []networkingv1.IngressTLS {
```
iwilltry42, line 278: Feels weird passing in `useCertManager` to `setupCertManager`.. like if you don't want to use cert manager, then don't call this function :sweat_smile: 
It's fine though.

iwilltry42 (changes requested):
Small change and some question... and a lot of sadness from my side for removing this  :cry:

pkg/publish/ingress.go
StrongMonkey: yep good idea!

pkg/controller/service/testdata/ingress/basic/expected.golden
```diff
@@ -70,14 +70,14 @@ apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   annotations:
-    acorn.io/targets: '{"ci1.acorn.not":{"port":81,"service":"oneimage"},"localhost":{"port":81,"service":"oneimage"},"oneimage-app-name-a5b0aade.local.oss-acorn.io":{"port":81,"service":"oneimage"}}'
+    acorn.io/targets: '{"oneimage-app-name-a5b0aade.local.oss-acorn.io":{"port":81,"service":"oneimage"}}'
```
StrongMonkey, line 73: The change is expected. This PR breaks down ingress into custom domain and cluster domain so that we don't apply cert-manager annotation to existing cluster domain that is using certs issued by acorn's letsencrypt integration

pkg/controller/service/testdata/ingress/basic/expected.golden
```diff
@@ -70,14 +70,14 @@ apiVersion: networking.k8s.io/v1
...
   creationTimestamp: null
   labels:
     acorn.io/app-name: app-name
     acorn.io/app-namespace: app-namespace
     acorn.io/container-name: oneimage
     acorn.io/managed: "true"
-  name: oneimage
+  name: oneimage-cluster-domain
```
StrongMonkey, line 80: I am suffixing `cluster-domain` and `custom-domain` to seperate ingress resources.

pkg/controller/service/testdata/ingress/basic/expected.golden
```diff
@@ -179,7 +202,7 @@ status:
     type: defined
   endpoints:
   - address: oneimage-app-name-a5b0aade.local.oss-acorn.io
-    publishProtocol: https
```
StrongMonkey, line 182: I think this change is also expected. Previously, it will set `publishProtocol` to https if one ingress TLS is found, which is not correct. After this change, if custom domain is setting ingress but cluster domain doesn't have cert provisioned, the cluster domain remain to be http.

https://github.com/acorn-io/acorn/blob/main/pkg/publish/ingress.go#L206-L209

pkg/controller/service/service_test.go
```diff
@@ -109,3 +109,7 @@ func TestRouter(t *testing.T) {
 func TestSecret(t *testing.T) {
 	tester.DefaultTest(t, scheme.Scheme, "testdata/secret", RenderServices)
 }
+
+func TestCertManager(t *testing.T) {
```
StrongMonkey, line 113: Added a new test for testing cert-manager setting

docs/docs/50-running/03-certificates.md
```diff
@@ -53,3 +53,32 @@ acorn run -p my-app.example.com:web [MY_APP_IMAGE]
 
 Acorn will automatically inspect each certificate in the Acorn namespace for one that can be used with `my-app.example.com`.
 If no TLS secret is found with that FQDN, it will be exposed on HTTP only.
+
+### Issuing custom domain certs
+
+Acorn Let's Encrypt integration does not issue certificates for custom domain. Instead, you will rely on external cert-manager to issue certificates. To do so, you will need to create a cluster issuer first. For more information on how to configure cert-manager, see [cert-manager docs](https://cert-manager.io/docs/).
+
+```yaml
```
StrongMonkey, line 61: Docs has been called out on how to use it. Those setting will be managed in hub. https://github.com/acorn-io/infra-provisioning-cdk/pull/23

StrongMonkey:
https://github.com/acorn-io/infra-provisioning-cdk/pull/23 Configures the managed runtime setting in hub.

docs/docs/50-running/03-certificates.md
```diff
@@ -53,3 +53,32 @@ acorn run -p my-app.example.com:web [MY_APP_IMAGE]
 
 Acorn will automatically inspect each certificate in the Acorn namespace for one that can be used with `my-app.example.com`.
 If no TLS secret is found with that FQDN, it will be exposed on HTTP only.
+
+### Issuing custom domain certs
+
+Acorn Let's Encrypt integration does not issue certificates for custom domain. Instead, you will rely on external cert-manager to issue certificates. To do so, you will need to create a cluster issuer first. For more information on how to configure cert-manager, see [cert-manager docs](https://cert-manager.io/docs/).
```
iwilltry42, line 59:
```suggestion
Acorn's Let's Encrypt integration does not issue certificates for custom domains. Instead, you will rely on external cert-manager to issue certificates. To do so, you will need to create a cluster-issuer first. For more information on how to install and configure cert-manager, see [cert-manager docs](https://cert-manager.io/docs/).
```

pkg/apis/api.acorn.io/v1/types.go
```diff
@@ -424,7 +424,7 @@ type Config struct {
 	AWSIdentityProviderARN         *string         `json:"awsIdentityProviderArn" name:"aws-identity-provider-arn" usage:"ARN of cluster's OpenID Connect provider registered in AWS"`
 	EventTTL                       *string         `json:"eventTTL" name:"event-ttl" usage:"Amount of time an Acorn event will be stored before being deleted (default '168h' - 7 days)"`
 	Features                       map[string]bool `json:"features" name:"features" boolmap:"true" usage:"Enable or disable features. (example foo=true,bar=false)"`
-	CertManagerIssuer              *string         `json:"certManagerIssuer" name:"cert-manager-issuer" usage:"The name of the cert-manager cluster issuer to use for TLS certificates on custom domains (default letsencrypt-prod)" default:"letsencrypt-prod"`
+	CertManagerIssuer              *string         `json:"certManagerIssuer" name:"cert-manager-issuer" usage:"The name of the cert-manager cluster issuer to use for TLS certificates on custom domains" default:""`
```
iwilltry42, line 427: I'm actually wondering, if we could create some sort of "acorn-protected" issuer.
I.e. you don't use cluster-issuer, but a namespaced issuer that is installed into the acorn-system namespace.
Challenges will happen completely in acorn-system namespace and only the resulting TLS secret is copied to the app namespace :thinking: 
Might be over-engineered :sweat_smile:

iwilltry42 (approved):
Small nit in the docs and a question, but nothing blocking.

pkg/apis/api.acorn.io/v1/types.go
StrongMonkey: Yeah using issuer doesn't make any sense unless we do some magic like copying certs around like you said

iwilltry42: I think we actually do that already here: https://github.com/acorn-io/acorn/blob/main/pkg/publish/cert.go#L131-L142

pkg/publish/ingress.go
g-linville: Agreed

g-linville (commented):
Documentation nits

pkg/publish/ingress.go
```diff
@@ -176,74 +186,92 @@ func Ingress(req router.Request, svc *v1.ServiceInstance) (result []kclient.Obje
...
+				}),
+			},
+			Spec: networkingv1.IngressSpec{
+				IngressClassName: ingressClassName,
+				Rules:            rules.rules,
+				TLS:              ingressTLS,
+			},
+		}
+		// if custom domains don't use cert-manager, use the default acorn cert-manager setting
+		if useCertManager && (ingress.Annotations["cert-manager.io/cluster-issuer"] == "" && ingress.Annotations["cert-manager.io/issuer"] == "") {
```
StrongMonkey, line 259: Only setting cert-manager annotation if there is no cert-manager annotation being set... This is so that if user sets the annotation outside in acorn, we should respect that.

g-linville (approved):

pkg/publish/ingress.go
StrongMonkey: Makes sense. I have moved out the condition to the toplevel.

Issue #347, opened 2022-07-27
Add ability to create functional basic secrets from cli


cjellick:
```
acorn secrets create --type basic my-secret
```
should create a functional basic secret with username and password set, but instead its an empty secret with no keys.

In reality, I was trying to figure out the k8s yaml for generating a secret so that I could generate this secret in fleet and use it in two different (gitops created) acorn apps.

So, I also want `kubectl create -f secret.yaml` to work where secret.yaml is
```
apiVersion: api.acorn.io/v1
kind: Secret
metadata:
  name: staging-db-user-credentials2
  namespace: acorn
type: basic
```

pratikjagrut:
Hi, @cjellick Is the issue still relevant? If yes, I would like to work on it.

sangee2004:
Tested with acorn version - `v0.8.0-alpha7-114-gbf4c72ea+bf4c72ea`

```
acorn secrets create --type basic my-secret
my-secret
sangeethahariharan@Sangeethas-MacBook-Pro acorn-starter-nextjs % acorn secret reveal my-secret
NAME        TYPE      KEY        VALUE
my-secret   basic     password   t#@%s27SwXqxGY8&
my-secret   basic     username   =pWfeMMz
```

We are also able to create secret using kubectl using the following yaml 

kubectl create -f ./secret.yml -n acorn

```apiVersion: api.acorn.io/v1
kind: Secret
metadata:
  name: cred1
  namespace: acorn
type: basic
```

```
acorn secret reveal cred1
NAME      TYPE      KEY        VALUE
cred1     basic     password   oe-F^j=-5Lk#RXb!
cred1     basic     username   !+f*%%#y
sangeethahariharan@Sangeethas-MacBook-Pro ~ % 
```

PR #1808, opened 2023-06-17
feat: Set defaults for empty basic secrets (#347)

pratikjagrut:
If no file or data is passed while creating a basic secret then default username=username and password=password will be set.
This PR addresses #347 issue.
### Checklist

### DIFF
--- a/pkg/secrets/secret.go
+++ b/pkg/secrets/secret.go
@@ -251,7 +251,7 @@ func generateBasic(req router.Request, appInstance *v1.AppInstance, secretName s
 
 	for i, key := range []string{corev1.BasicAuthUsernameKey, corev1.BasicAuthPasswordKey} {
 		if len(secret.Data[key]) == 0 {
-			v, err := GenerateRandomSecret(16)
+			v, err := GenerateRandomSecret(54)
 			v = v[:(i+1)*8]
 			if err != nil {
 				return nil, err
--- a/pkg/secrets/secret.go
+++ b/pkg/secrets/secret.go
@@ -22,7 +22,6 @@ import (
 	"github.com/acorn-io/runtime/pkg/system"
 	"github.com/rancher/wrangler/pkg/data/convert"
 	"github.com/rancher/wrangler/pkg/merr"
-	"github.com/rancher/wrangler/pkg/randomtoken"
 	"golang.org/x/exp/maps"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
@@ -252,8 +251,7 @@ func generateBasic(req router.Request, appInstance *v1.AppInstance, secretName s
 
 	for i, key := range []string{corev1.BasicAuthUsernameKey, corev1.BasicAuthPasswordKey} {
 		if len(secret.Data[key]) == 0 {
-			// TODO: Improve with more characters (special, upper/lowercase, etc)
-			v, err := randomtoken.Generate()
+			v, err := GenerateRandomSecret(16)
 			v = v[:(i+1)*8]
 			if err != nil {
 				return nil, err
@@ -261,6 +259,7 @@ func generateBasic(req router.Request, appInstance *v1.AppInstance, secretName s
 			secret.Data[key] = []byte(v)
 		}
 	}
+
 	return updateOrCreate(req, existing, secret)
 }
 
--- a/pkg/server/registry/apigroups/acorn/secrets/defaultCreate.go
+++ b/pkg/server/registry/apigroups/acorn/secrets/defaultCreate.go
@@ -1,0 +1,43 @@
+package secrets
+
+import (
+	"context"
+
+	"github.com/acorn-io/mink/pkg/strategy"
+	"github.com/acorn-io/mink/pkg/types"
+	apiv1 "github.com/acorn-io/runtime/pkg/apis/api.acorn.io/v1"
+	sec "github.com/acorn-io/runtime/pkg/secrets"
+)
+
+type defaultSecretGenerateStrategy struct {
+	strategy strategy.CompleteStrategy
+}
+
+func (d *defaultSecretGenerateStrategy) Create(ctx context.Context, object types.Object) (types.Object, error) {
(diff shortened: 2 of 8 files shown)


tylerslaton (approved):

pratikjagrut:
Hey @cjellick, could you review this PR? It's related to your issue. Thanks!

cjellick:
My hope/expectation in that issue was that random values would be generated for username and password, as they are when you create a basic secret similarly in an acornfile, see: https://docs.acorn.io/authoring/secrets#basic-secrets

pratikjagrut:
@cjellick Thank you for the feedback.

I have incorporated your suggested changes in the updated PR. Additionally, I noticed the comment in the code section you mentioned [here](https://github.com/acorn-io/runtime/blob/main/pkg/secrets/secret.go#L255), which appears to be related to generating a random username and secret from the Acornfile, as you mentioned. In response, I have implemented a new secret generation function based on that comment. Please review the changes and let me know if they meet your approval.

g-linville:
@pratikjagrut thank you for your interest in contributing to Acorn! These changes work well for generating default secret values in the CLI, but the issue also includes this:

> So, I also want kubectl create -f secret.yaml to work where secret.yaml is
> ```
> apiVersion: api.acorn.io/v1
> kind: Secret
> metadata:
>   name: staging-db-user-credentials2
>   namespace: acorn
> type: basic
> ```

In order for this to work, there needs to be some logic in the api-server to generate the values there.

The api-server code is in the folder `pkg/server`. In particular we are interested in `pkg/server/registry/apigroups/acorn/secrets`. Here you will see a file called `storage.go` which sets up the strategy for storing Acorn secrets. We use a translation strategy, meaning, that they will be stored as CRDs internally in Kubernetes, but will go through some translation logic before being presented to the user.

It is possible to override the secret creation logic with some custom code to generate these values. If you take a look at the folder `pkg/server/registry/apigroups/acorn/builds`, you can see in its `storage.go` file that it is also using a translation strategy, but it creates another strategy to wrap around the translation strategy specifically for creation. You can see the code for that wrapper in the `strategy.go` file's `Create` function. You can use a pattern like this to generate the default values in the api-server.

Thanks again for your interest in contributing and let me know if you have any questions!

pkg/secrets/secret.go
```diff
@@ -252,15 +251,15 @@ func generateBasic(req router.Request, appInstance *v1.AppInstance, secretName s
 
 	for i, key := range []string{corev1.BasicAuthUsernameKey, corev1.BasicAuthPasswordKey} {
 		if len(secret.Data[key]) == 0 {
-			// TODO: Improve with more characters (special, upper/lowercase, etc)
-			v, err := randomtoken.Generate()
+			v, err := GenerateRandomSecret(16)
```
g-linville, line 254: nit: the function that this replaces generated a string of length 54, so let's do that here too

```suggestion
			v, err := GenerateRandomSecret(54)
```

pkg/server/registry/apigroups/acorn/secrets/defaultCreate.go
```diff
@@ -0,0 +1,43 @@
...
+		}
+
+		password, err := sec.GenerateRandomSecret(16)
+		if err != nil {
+			return nil, err
+		}
+
+		if secret.Data == nil {
+			secret.Data = map[string][]byte{}
+		}
```
g-linville, line 33: nit: this was already checked in the outer if statement

```suggestion
		secret.Data = map[string][]byte{}
```

pkg/server/registry/apigroups/acorn/secrets/defaultCreate.go
g-linville, line 1: nit: I would like this file to be renamed to `strategy.go`, since that is the convention used with other API types.

g-linville (changes requested):
Looks good, @pratikjagrut! I left a few small comments. Also, I would like for the username/password generation in the CLI code to be removed so that the apiserver can always do it instead.

pkg/secrets/secret.go
pratikjagrut: Even if I make the suggested changes, they won't have any impact since the subsequent code truncates the length to 8 characters for the username and 16 characters for the password.

pkg/server/registry/apigroups/acorn/secrets/defaultCreate.go
pratikjagrut: Good catch. Thanks. :)

pkg/server/registry/apigroups/acorn/secrets/defaultCreate.go
pratikjagrut: okay.

pkg/server/registry/apigroups/acorn/secrets/strategy .go
```diff
@@ -0,0 +1,41 @@
...
+
+		password, err := sec.GenerateRandomSecret(16)
+		if err != nil {
+			return nil, err
+		}
+
+		secret.Data = map[string][]byte{}
+
+		secret.Data["username"] = []byte(username)
+		secret.Data["password"] = []byte(password)
```
g-linville, line 34: nit: one small cleanup

```suggestion
		secret.Data = map[string][]byte{
			"username": []byte(username),
			"password": []byte(password),
		}
```

pkg/server/registry/apigroups/acorn/secrets/strategy .go
g-linville, line 1: nit: the name of this file has a space before the extension

g-linville (commented):
@pratikjagrut couple more small things, but looks good! I'll have one of the other Acorn Labs guys look over it too.

tylerslaton (approved):
Looks great! Thank you for the contribution @pratikjagrut!

pkg/server/registry/apigroups/acorn/secrets/strategy .go
g-linville: @pratikjagrut I would still like for this to be changed, if you are willing.

g-linville (approved):

PR #1811, opened 2023-06-19
Add CPU/Memory request/limits and PriorityClasses for system components

tylerslaton:
This PR ensures that the critical Acorn workloads have the same priority class as critical Kubernetes workloads. In addition, it uses internal usage metrics to determine logical values for their request/limit of memory and CPU. 

### Checklist

### DIFF
--- a/pkg/system/resources.go
+++ b/pkg/system/resources.go
@@ -1,0 +1,75 @@
+package system
+
+import (
+	"os"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+// Values will likely need to be tweaked as we get more usage data. They are currently based
+// on metrics we have collected from internal use. You can override these values by setting
+// the corresponding environment variable.
+var (
+	mi = int64(1_048_576) // 1 MiB in bytes
+
+	registryMemoryRequest = *resource.NewQuantity(40*mi, resource.BinarySI)    // REGISTRY_MEMORY_REQUEST
+	registryMemoryLimit   = *resource.NewQuantity(80*mi, resource.BinarySI)    // REGISTRY_MEMORY_LIMIT
+	registryCPURequest    = *resource.NewMilliQuantity(50, resource.DecimalSI) // REGISTRY_CPU_REQUEST
+
+	buildkitdMemoryRequest = *resource.NewQuantity(100*mi, resource.BinarySI)   // BUILDKITD_MEMORY_REQUEST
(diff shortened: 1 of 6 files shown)


pkg/system/resources.go
```diff
@@ -0,0 +1,69 @@
...
+		},
+	}
+}
+
+func envOrDefault(env, def string) resource.Quantity {
+	amount := def
+	if env != "" {
+		amount = env
+	}
+	return resource.MustParse(amount)
```
thedadams, line 68: Typically, this is frowned upon because, if the environment variable cannot be parsed, then this will panic. That would be OK if this was parsed on startup only, but I don't think that is the case here.

Are you opposed to checking an error and reverting to the default if an error occurs?

pkg/system/resources.go
```diff
@@ -0,0 +1,69 @@
...
+			corev1.ResourceMemory: envOrDefault(os.Getenv("BUILDKITD_SERVICE_MEMORY_REQUEST"), buildkitdServiceMemoryRequest),
+			corev1.ResourceCPU:    envOrDefault(os.Getenv("BUILDKITD_SERVICE_CPU_REQUEST"), buildkitdServiceCPURequest),
+		},
+		Limits: corev1.ResourceList{
+			corev1.ResourceMemory: envOrDefault(os.Getenv("BUILDKITD_SERVICE_MEMORY_LIMIT"), buildkitdServiceMemoryLimit),
+		},
+	}
+}
+
+func envOrDefault(env, def string) resource.Quantity {
```
StrongMonkey, line 63: Nit: you can do `os.GetEnv()` here so you don't have to call it outside the function

pkg/system/resources.go
tylerslaton: Nope, not opposed to that at all. I'll get to implementing that.

pkg/system/resources.go
tylerslaton: Good call 👍🏻

pkg/system/resources.go
```diff
@@ -0,0 +1,75 @@
...
+		},
+	}
+}
+
+func envOrDefault(env string, def resource.Quantity) resource.Quantity {
+	if os.Getenv(env) == "" {
+		return def
+	}
+
+	quantity, err := resource.ParseQuantity(env)
```
g-linville, line 70:
```suggestion
	quantity, err := resource.ParseQuantity(os.Getenv(env))
```

tylerslaton: Just made a change to this. What do you think?

g-linville: Works for me

g-linville (approved):

StrongMonkey (approved):

PR #1815, opened 2023-06-20
Add PriorityClass support to ComputeClasses

tylerslaton:
With this PR, `ComputeClasses` now come with a new field, `PriorityClassName`. This allows users create `ComputeClasses` to define a `PriorityClass` that will get set on all Pods using the `ComputeClass`.

### Checklist


pkg/controller/scheduling/scheduling.go
```diff
@@ -90,22 +93,36 @@ func addScheduling(req router.Request, appInstance *v1.AppInstance, workloads ma
...
 	return nil
 }
 
-// Add edits the provided PodTemplateSpec to have the applied configuration for the ComputeClass and Memory values
-func Nodes(req router.Request, name string, container v1.Container, app *v1.AppInstance, computeClass *adminv1.ProjectComputeClassInstance) (*corev1.Affinity, []corev1.Toleration, error) {
-	if computeClass != nil {
-		// Return any custom affinities and tolerations from the ComputeClass
-		return computeClass.Affinity, computeClass.Tolerations, nil
+// Nodes returns the Affinity and Tolerations from a ComputeClass if they exist
+func Nodes(req router.Request, computeClass *adminv1.ProjectComputeClassInstance) (*corev1.Affinity, []corev1.Toleration) {
```
tylerslaton, line 106: Also tidied up this function a bit.

pkg/install/role.yaml
```diff
@@ -78,6 +78,9 @@ rules:
   - verbs: ["get", "list", "watch"]
     apiGroups: ["storage.k8s.io"]
     resources: ["storageclasses"]
+  - verbs: ["*"]
```
thedadams, line 81: Can this just be `["get", "list", "watch"]`?

pkg/server/registry/apigroups/admin/computeclass/validator.go
```diff
@@ -39,6 +40,13 @@ func (s *ProjectValidator) Validate(ctx context.Context, obj runtime.Object) (re
 		}
 	}
 
+	// If a PriorityClass is specified, validate it exists
+	if cc.PriorityClassName != "" {
+		if err := s.client.Get(ctx, kclient.ObjectKey{Name: cc.PriorityClassName, Namespace: ""}, &schedulingv1.PriorityClass{}); err != nil {
+			return append(result, field.Invalid(field.NewPath("spec", "priorityClassName"), cc.PriorityClassName, err.Error()))
+		}
+	}
```
thedadams, line 48: Do we want to validate this on create? There are other instances where we don't validate such things. For example, we don't validate that a storage class exists when a volume class is created.

tylerslaton: Yeah so I had a similar thought. My thought process went slightly different so please let me know what you think.

If the PriorityClass isn't validated here, it is validated on App creation (like memory, cpu, etc is now). What felt different about that situation is that it felt off for an App to fail creation because a ComputeClass references a non-existent/misconfigured PriorityClass. What does feel weird about this is if a PriorityClass gets deleted after a ComputeClass gets created. In that event I added some controller logic on the App to verify that PriorityClass exists.

pkg/install/role.yaml
tylerslaton: Yes, it can.

pkg/server/registry/apigroups/admin/computeclass/validator.go
thedadams: Volume classes are similar. A volume class can reference any storage class it wants; we don't validate that on creation of the volume class.

If the storage class doesn't exist, then the app will fail when it is created because Kubernetes won't be able to find the storage class. It may not return an error, but it will hang forever because the volume won't be created.

tylerslaton: Okay, I agree with you here then. I've removed the API Server validation. The controller already had logic to check if the PriorityClass exists, do you think that should be removed as well in favor of just letting the Pods create?

thedadams (approved):

g-linville (approved):

PR #1820, opened 2023-06-21
Split make test target

njhale:
Split make test target into separate unit and integration targets.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -30,7 +30,8 @@ jobs:
       - run: docker buildx install
       - run: make setup-ci-image
       - run: ./bin/acorn install --image acorn:v-ci --skip-checks --acorn-dns=disabled --network-policies=true
-      - run: TEST_ACORN_CONTROLLER=external TEST_FLAGS="--junitfile test-summary.xml" make test
+      - run: TEST_FLAGS="--junitfile test-summary.xml" make unit
+      - run: TEST_ACORN_CONTROLLER=external TEST_FLAGS="--junitfile test-summary.xml" make integration
       - name: Build test summary
         uses: test-summary/action@v1
         if: success() || failure()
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yaml
```diff
@@ -30,7 +30,8 @@ jobs:
       - run: docker buildx install
       - run: make setup-ci-image
       - run: ./bin/acorn install --image acorn:v-ci --skip-checks --acorn-dns=disabled --network-policies=true
-      - run: TEST_ACORN_CONTROLLER=external TEST_FLAGS="--junitfile test-summary.xml" make test
+      - run: TEST_FLAGS="--junitfile test-summary.xml" make unit
+      - run: TEST_ACORN_CONTROLLER=external TEST_FLAGS="--junitfile test-summary.xml" make integration
```
tylerslaton, line 34: What would your thoughts be on breaking out these two steps into their own jobs? I see a few benefits here:

1. Unit tests shouldn't (at least in theory) need an Acorn installation or image built. So the separated Unit job can ignore those parts.
2. The Unit and Integration tests can run in parallel which would speed up the feedback loop.
3. Each test type gets their own summary.

The main con is that we increase the amount of action minutes we use but not by that much.

tylerslaton (approved):
Had a comment about breaking these tests out into their own jobs for a few reasons. It isn't blocking though and I could see this going in as is.

.github/workflows/test.yaml
njhale: Totally agree that we should break this out into two distinct jobs. I chose this route for now because we effectively still get two separate `go test` runs with separate timeouts, but without the hassle of rewiring the test output/slack reporting.

I was thinking about doing that as a follow up. WDYT?

tylerslaton: Yep! Sounds good to me. Totally fine with this being a follow-up.

g-linville (approved):

PR #1822, opened 2023-06-21
Generalize event TTL logic

njhale:
Relax GCExpired constraints to allow any type implementing the GCable
interface to be cleaned up.

### DIFF
--- a/pkg/controller/eventinstance/eventinstance.go
+++ b/pkg/controller/eventinstance/eventinstance.go
@@ -25,7 +25,10 @@ func GCExpired() router.HandlerFunc {
 	}
 
 	return handler{
-		getTTL: func(ctx context.Context, getter kclient.Reader) (time.Duration, error) {
+		getTTL: func(
+			ctx context.Context,
+			getter kclient.Reader,
+		) (time.Duration, error) {
 			cfg, err := config.Get(ctx, getter)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/controller/eventinstance/eventinstance.go
```diff
@@ -71,22 +86,26 @@ func (h handler) gcExpired(req router.Request, resp router.Response) error {
 	}
 
 	// Check expiration
-	if now, expiration := time.Now(), e.Observed.Add(ttl); now.Before(expiration) {
+	if now, expiration := time.Now(), e.(GCable).GetObserved().Add(ttl); now.Before(expiration) {
```
thedadams, line 89: This will panic if we accidentally use this handler with something that isn't `GCable`. Can we put a type cast near the top of this method and log a warning if the cast fails?

njhale: I can, but let me first note my reasoning behind not doing that to begin with: we will already panic in almost every handler when performing the initial type assertion if the type isn't what we expected; e.g. `e := req.Object.(*v1.EventInstance)`. To me, this is very similar in that regard.

The situation where this would panic is a programming error, for which I think a panic is most apt.

thedadams (approved):

g-linville (approved):

Issue #1427, opened 2023-04-03
Not able to deploy apps using dockerhub images in autoupgrade format without using "docker.io" in image name.


sangee2004:
acorn version v0.6.0-106-g09f636a2+09f636a2 

This behavior is also seen in older versions.

Steps to reproduce the problem:

```
1.  acorn run --interval 15s -n check1 docker.io/sangeetha/myfirstacorn:v#.#.# ==> works

2.  acorn run --interval 15s -n mytestautonew1 sangeetha/myfirstacorn:v#.#.# ==> does not work 

3.  acorn run  sangeetha/myfirstacorn:v6.0.0 ==> works (No autoupgrade format) 
```

Do we expect `#2` work ? `#1` working and `#3` working but `#2` not working is confusing behavior.

g-linville:
I have a draft PR for this here: https://github.com/acorn-io/acorn/pull/1817

However, this has a major consequence that I **really** do not like.

For auto-upgrade, the rule currently is that we will always prioritize/prefer remote images over local ones when resolving. If we allow the implicit use of Docker Hub, then we open up users to a dependency confusion attack, which could play out like this:

1. I build a local image and tag it with `myproject/myimage:v1.0.0`
2. I run the local image in an auto-upgrade app with `acorn run -n myapp myproject/myimage:v#.#.#`
3. A threat actor uploads a public Docker Hub image called `index.docker.io/myproject/myimage:v2.0.0`
4. Because Acorn prefers remote images over local images for auto-upgrade, the auto-upgrade daemon finds the image on Docker Hub and updates to it

Imo we should not support having an implicit Docker Hub registry for auto-upgrade. It opens up too many risks.

Curious to hear everyone's thoughts.

tylerslaton:
> @g-linville However, this has a major consequence that I really do not like.

I have two thoughts here:

1. Whatever we do, should we make it consistent? As Sangeetha points out well here - there's a difference in behavior between the two modes. As a user I personally would expect both situations to act similarly and if they don't its clearly communicated.
2. Just from what you've outline here, I'd say that requiring a fully-qualified (for lack of a better term) image during auto-upgrades to prevent the situation you're describing. Could we add (or does one already exist?) an API server error that rejects any auto-upgrade images that are ambiguous?

g-linville:
[...]

PR #1823, opened 2023-06-21
Prevent Acorn from assuming Docker Hub for auto-upgrade apps with no specified registry (#1427)

g-linville:
for #1427

This change ensures that apps with auto-upgrade enabled will never implicitly pull images from Docker Hub.

This is important in order to avoid dependency confusion. Since auto-upgrade apps, unlike normal apps, will give priority to remote images when resolving, it would be possible for an attacker to create a public image that matches the name of a private, local-only image, and Acorn would pick it up and auto-upgrade to that. This PR prevents that from being possible.

Unfortunately, there were three separate areas where I had to implement this logic:
- Get image details
- App validation in the apiserver
- Pull app image

I wrote tests (two unit, one integration) to check for this logic in all three places.

Currently, prior to this PR, it is possible for auto-upgrade apps to implicitly use Docker Hub as long as they aren't using a tag pattern (e.g. `acorn run -n nginx grantlinville/nginx:latest` works and uses my image from Docker Hub). This will no longer work as soon as this is merged. For running apps that are like this, they will continue to run, but there will be an error in their status until their image is updated using `acorn update` (or until a local image that matches the auto-upgrade is built).

### Checklist

### DIFF
--- a/pkg/controller/appdefinition/pullappimage.go
+++ b/pkg/controller/appdefinition/pullappimage.go
@@ -78,7 +78,7 @@ func pullAppImage(transport http.RoundTripper, client pullClient) router.Handler
 					}
 					if ref.Context().RegistryStr() == defaultNoReg {
 						// Prevent this from being resolved remotely, as we should never assume Docker Hub for auto-upgrade apps
-						return fmt.Errorf("no local image found for %v\nif you are trying to use Docker Hub, use docker.io/%v", target, target)
+						return fmt.Errorf("no local image found for %v - if you are trying to use Docker Hub, use docker.io/%v", target, target)
 					}
 				}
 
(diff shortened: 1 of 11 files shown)


pkg/controller/appdefinition/pullappimage.go
```diff
@@ -13,12 +13,15 @@ import (
...
 	"github.com/acorn-io/acorn/pkg/tags"
 	"github.com/acorn-io/baaah/pkg/router"
+	imagename "github.com/google/go-containerregistry/pkg/name"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/sirupsen/logrus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const defaultNoReg = "xxx-no-reg"
```
thedadams, line 23: There is an exported `const` in the `pkg/images` package called `DefaultRegistry`. Can this exported constant be changed (I think `xxx-no-reg` makes more sense) and be used everywhere we need it?

cjellick (approved):

thedadams (approved):

PR #1824, opened 2023-06-21
Rename to acorn-io/runtime

thedadams:


thedadams:
This will merge after the repository is renamed.

g-linville (approved):

tylerslaton:
Should this be `runtime`?

https://github.com/acorn-io/acorn/blob/e790cf834c6727999e58de3cbf757eaa93c292ee/.goreleaser.yml#L80C6-L80C6

thedadams:
@tylerslaton Yes! Great catch. Both should be `runtime`.

docs/versioned_docs/version-0.3/10-home.md
```diff
@@ -5,7 +5,7 @@ slug: /
 
 :::caution
 
-Acorn is a work in progress.  Features will evolve over time and there may be breaking changes between releases.  Please give us your feedback in [Slack](https://slack.acorn.io), [Discussions](https://github.com/acorn-io/acorn/discussions), or [Issues](https://github.com/acorn-io/acorn/issues)!
+Acorn is a work in progress.  Features will evolve over time and there may be breaking changes between releases.  Please give us your feedback in [Slack](https://slack.acorn.io), [Discussions](https://github.com/acorn-io/runtime/discussions), or [Issues](https://github.com/acorn-io/runtime/issues)!
```
njhale, line 8: Do we want to update docs for older releases? With the redirects, they shouldn't break.

tylerslaton (approved):

main.go
```diff
@@ -1,7 +1,7 @@
 package main
 
 import (
-	acorn "github.com/acorn-io/acorn/pkg/cli"
+	acorn "github.com/acorn-io/runtime/pkg/cli"
```
njhale, line 4:
```suggestion
	runtime "github.com/acorn-io/runtime/pkg/cli"
```

pkg/apis/internal.admin.acorn.io/v1/scheme.go
```diff
@@ -1,7 +1,7 @@
 package v1
 
 import (
-	internal_admin_acorn_io "github.com/acorn-io/acorn/pkg/apis/internal.admin.acorn.io"
+	internal_admin_acorn_io "github.com/acorn-io/runtime/pkg/apis/internal.admin.acorn.io"
```
njhale, line 4: nit(not for this PR): interesting snake_case

njhale (approved):
LGTM -- let's see what breaks.

docs/versioned_docs/version-0.3/10-home.md
thedadams: I am open to not doing it. I also thought that the updating it wouldn't break anything either.

Do you have a preference for not doing this?

main.go
thedadams: This seems to be referring to the actual `acorn` CLI command, so I would be inclined to keep it as is. Would you still feel better about changing it?

docs/versioned_docs/version-0.3/10-home.md
njhale: I did, but on second thought, we'll eventually remove these, and I don't expect users to be poking around in such old docs anyway.

So, I'm fine with merging as-is, old doc changes included.

main.go
njhale: good point. I'm fine with it as is.

Issue #1812, opened 2023-06-19
acorn secret reveal -o json has incorrect values.


sangee2004:
**acorn version** v0.7.1-15-g167b9f5e+167b9f5e

**Steps to reproduce the problem:**

1. Create a secret 
```
 acorn secret create --data test1=value1 --data test2=value2 mytestsec
```
2. Use `acorn secret reveal` to reveal the secret.

```
 % acorn secret reveal mytestsec        
NAME        TYPE      KEY       VALUE
mytestsec   opaque    test1     value1
mytestsec   opaque    test2     value2

```
3. Use `acorn secret reveal -o json ` to reveal the secret. Notice that there are 2 entries returned . Each entries has details of both keys test1 and test2. The values are incorrect in one of them and empty in the other entry.

```
acorn secret reveal mytestsec -o json
{
    "items": [
        {
            "kind": "Secret",
            "apiVersion": "api.acorn.io/v1",
            "metadata": {
                "name": "mytestsec",
                "namespace": "acorn",
                "uid": "1a2dea61-83d9-4d55-93c2-9c37698f0835-s-p",
                "resourceVersion": "1517564",
                "creationTimestamp": "2023-06-19T22:53:14Z"
            },
            "type": "opaque",
            "data": {
                "test1": "dmFsdWUx",
                "test2": "dmFsdWUy"
            },
            "keys": [
                "test1",
                "test2"
            ]
        },
        {
            "kind": "Secret",
            "apiVersion": "api.acorn.io/v1",
            "metadata": {
                "name": "mytestsec",
                "namespace": "acorn",
                "uid": "1a2dea61-83d9-4d55-93c2-9c37698f0835-s-p",
                "resourceVersion": "1517564",
                "creationTimestamp": "2023-06-19T22:53:14Z"
            },
            "type": "opaque",
            "data": {
                "test1": "dmFsdWUx",
                "test2": "dmFsdWUy"
            },
            "keys": [
                "test1",
                "test2"
            ]
        }
    ]
}
```

iwilltry42:
Regression introduced by https://github.com/acorn-io/acorn/pull/1806

1. "Wrong" values because it's now printing the plain Secret Objects, with values not decoded
2. Duplicate entry, because it's printing on a loop over the included keys

Question is if we actually want to have the raw object output here.

sangee2004:
Tested with acorn version `v0.8.0-alpha7-98-gbb130866+bb130866`

[...]

PR #1833, opened 2023-06-23
fix: cleanup `acorn secrets reveal` and general non-table output (#1812)

iwilltry42:
We had 3 problems here:

1. Secret was printed twice because we were putting the source object into buffer once for every secret data key
2. [General] Due to the `map[string]any` type output, the object fields (e.g. managed fields) were not cleaned anymore
3. Secret data values were not encoded as it printed the plain objects, where the secret data values fields are []byte

Ref #1812

### Checklist

### DIFF
--- a/pkg/cli/builder/table/funcs.go
+++ b/pkg/cli/builder/table/funcs.go
@@ -126,7 +126,7 @@ func toKObject(obj any) (kclient.Object, bool) {
 }
 
 func cleanFields(obj any) any {
-	if ol, ok := obj.(ObjectList); ok {
+	if ol, ok := obj.(objectList); ok {
 		for i, o := range ol.Items {
 			ol.Items[i] = cleanFields(o)
 		}
--- a/pkg/cli/builder/table/funcs.go
+++ b/pkg/cli/builder/table/funcs.go
@@ -154,9 +154,36 @@ func cleanFields(obj any) any {
 			}
 		}
 		ro.SetAnnotations(annotations)
+
+		// decode secret values
+		if sec, ok := ro.(*apiv1.Secret); ok {
+			return decodeSecret(sec)
+		}
+
 		return ro
 	}
 	return obj
+}
+
+func decodeSecret(sec *apiv1.Secret) any {
+	decodedSecret := struct {
+		metav1.TypeMeta   `json:",inline"`
+		metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+
+		Type string            `json:"type,omitempty"`
+		Data map[string]string `json:"data,omitempty"`
+		Keys []string          `json:"keys,omitempty"`
+	}{
+		TypeMeta:   sec.TypeMeta,
+		ObjectMeta: sec.ObjectMeta,
+		Type:       sec.Type,
+		Data:       map[string]string{},
+		Keys:       sec.Keys,
+	}
+	for k, v := range sec.Data {
+		decodedSecret.Data[k] = string(v)
+	}
+	return decodedSecret
 }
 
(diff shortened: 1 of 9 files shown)


iwilltry42:
Now some more tests are failing, because the list output is cleaned (again, as it looks like it was done before) from Acorn labels and some more stuff... question is if we really want that to happen, since some [acorn.io](http://acorn.io/) labels may be valuable to the user, right?

pkg/cli/builder/table/writer.go
```diff
@@ -230,6 +230,10 @@ func (t *writer) writeDataObject(objs ...any) error {
 	return errors.Join(t.errs...)
 }
 
+type ObjectList struct {
```
g-linville, line 233: nit: can this be unexported if it is only going to be used in this file?

pkg/cli/builder/table/funcs.go
```diff
@@ -147,11 +154,38 @@ func cleanFields(obj any) any {
 			}
 		}
 		ro.SetAnnotations(annotations)
+
+		// decode secret values
+		if sec, ok := ro.(*apiv1.Secret); ok {
+			return decodeSecret(sec)
+		}
```
njhale, line 161: We touched on this lightly offline, but I think we could also invert the dependency so that the type implementer can tell us how cleanup should work for their type; For example:

```go
package table
// ...

// Cleanable describes types that implement custom cleanup logic that should be called before writing them to a table buffer.
type Cleanable interface {
  // Clean returns a cleaned up copy of the underlying Cleanable.
  Clean() any
}
//...


func cleanFields(obj any) any {
  //...
  ro.SetAnnotations(annotations)

  // Finally, run any custom cleaning logic
  if c, ok := ro.(Cleanable); ok {
    return c.Clean()
  } 
  //...
}
```

Meanwhile, over at the type definition:

```go
package v1
//...

func (s *Secret) Clean() any {
  decodedSecret := struct {
    metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
  
    Type string            `json:"type,omitempty"`
    Data map[string]string `json:"data,omitempty"`
	Keys []string          `json:"keys,omitempty"`
  }{
    TypeMeta:   s.TypeMeta,
    ObjectMeta: s.ObjectMeta,
	Type:       s.Type,
	Data:       map[string]string{},
	Keys:       s.Keys,
  }
  for k, v := range sec.Data {
    decodedSecret.Data[k] = string(v)
  }
  return decodedSecret
}

```

The benefit is that we don't need to perform surgery on the table package whenever we need to extend cleanup logic.

njhale (commented):
@iwilltry42 -- looks good. Just have one suggestion, LMK what you think.

pkg/cli/builder/table/funcs.go
iwilltry42: I like that and now I understand what you wanted to tell me there :grin: 
Is it good to have that logic, which we only need for CLI output, in the API code though?
I guess it doesn't do any harm there, just wondering.

pkg/cli/builder/table/writer.go
iwilltry42: Yup. Done!

pkg/cli/builder/table/funcs.go
njhale: > Is it good to have that logic, which we only need for CLI output, in the API code though?

Ah, that's a great point. Now I'm on the fence about the value of the suggestion.

I don't want to hold this fix up when we can always refactor later if we think of something nicer.

njhale (approved):

g-linville (approved):

Issue #1712, opened 2023-06-05
ImageAllowRules - Not allowed to update images (not prompted for allowing images)


sangee2004:
acorn version v0.7.0-alpha1-148-g7909ed41+7909ed41

Steps to reproduce the problem:
Install acorn with `--features image-allow-rules=true`

Scenario1:
Deploy app from Acornfile.
When prompted for allowing images , select "yes" and allow for app to get deployed successfully.( Do you want to allow this app to use this (POTENTIALLY DANGEROUS) image? yes (this ID or SHA only))
Make changes to Acornfile and update app.  There is no prompt for allowing the new image sha and update fails
```
 acorn run --update -n mytest1 . 
....
 => => pushing manifest for 127.0.0.1:5000/acorn/new2:latest@sha256:d9350c3bbf3b1be57140aa42166ce0da845f6775305ec22094a1f1128cd65ea0                                                                                        0.0s
  ✗  ERROR:  image <> is not allowed by any ImageAllowRule in this project

```
Scenario2:
Deploy app from image  and choose to allow only this image.
Update app with a new image version. There is no prompt for allowing the new image  and update fails
```
acorn run --update -n mytestimage sangeetha/myfirstacorn:v0.0.2
  ✗  ERROR:  image <> is not allowed by any ImageAllowRule in this project
```

Expected Behavior:
There should be a prompt to allow for images on app updates.

cjellick:
I don't want the imageAllowRules feature to be blockers. Moving out of of v0.7

iwilltry42:
Waiting for #1742 to be merged before.
Afterwards it's just a copy-paste to add the same code to `AppUpdate` that we have in `AppRun` already.

sangee2004:
Blocked on testing this workflow because of regression introduced in - https://github.com/acorn-io/runtime/issues/1698#issuecomment-1639005670

PR #1835, opened 2023-06-23
fix: handle "image not allowed error" for both run and update (#1712)

iwilltry42:
Ref #1712

### Checklist

### DIFF
--- a/pkg/rulerequest/handle.go
+++ b/pkg/rulerequest/handle.go
@@ -18,6 +18,51 @@ import (
 	"github.com/pterm/pterm"
 )
 
+// handleNotAllowedError handles the case where an image is not allowed by IARs and prompts the user to create a basic one (image pattern only, no signatures, etc.)
+func handleNotAllowedError(ctx context.Context, c client.Client, dangerous bool, image string, err error, app *apiv1.App, opts any) (*apiv1.App, error) {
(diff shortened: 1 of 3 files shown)


pkg/rulerequest/handle.go
```diff
@@ -18,6 +18,51 @@ import (
...
+	} else if choice != "NO" {
+		iarErr := createImageAllowRule(ctx, c, image, choice, existingImgName) // existingImgName to ensure that this exact image ID is allowed in addition to whatever pattern we're allowing
+		if iarErr != nil {
+			return nil, iarErr
+		}
+		switch opts := opts.(type) {
+		case *client.AppRunOptions:
+			return c.AppRun(ctx, image, opts)
+		case *client.AppUpdateOptions:
+			return c.AppUpdate(ctx, app.Name, opts)
```
g-linville, line 57: nit: I would add something to the godoc comment for this function explaining how the `opts` parameter is used to make this determination. I think it's a clean solution but if I were looking at the function signature only, I would be pretty thrown off by it.

g-linville (approved):

pkg/rulerequest/handle.go
iwilltry42: Good point! Added another line to the comment.

thedadams (approved):

Issue #1802, opened 2023-06-16
Filter acorn events by source


njhale:
Add a way to filter the set of events returned by the `acorn events` command by the event source field such that a user can list events sourced from:
- resources of a certain type
- a specific object

To accomplish this, the `NAME` argument could be extended to accept a prefix that denotes a source type, followed optionally by the name of resource.

Here's what a snippet of the help text would look like for that approach:

```
# Filtering by Event Source 
  # The FILTER argument comes in the form '<source-kind>/<source-name>' and, when given, prunes the result set to contain events related to Acorn resources with the respective kind and name.
  # List events related to the 'hello' app in the current project
  acorn events app/hello
  
  # The '/<source-name>' suffix is optional and, when omitted, matches events related to any resource of the given kind.
  # List events related to any volume in the current project
  acorn events vol

# Filtering by Event Name
  # If FILTER doesn't have a '/name' suffix and isn't a known Acorn kind, it's value is interpreted as an event name.
```

sangee2004:
Tested with acorn version - `v0.8.0-alpha7-98-gbb130866+bb130866`

Able to filter events by app name and event name. Prefix matching is used to filter events.

```
acorn events app/mytest
SOURCE                                   NAME                               TYPE                  ACTOR                  OBSERVED                                  DESCRIPTION
app/mytest1                              b70d76ad6986fcf9af5b0a3572d5d713   AppCreate             acorn:internal-admin   2023-07-17 10:28:51.017037503 -0700 PDT   App testnew1/mytest1 created
app/mytest1                              323a3a45522cd136d5089eb5944a5ba4   AppImagePullSuccess   acorn-system           2023-07-17 10:28:51.351317887 -0700 PDT   Pulled test1:v1
app/mytest1                              0df27d96f8dd68282229f717c9c3de47   AppSpecUpdate         acorn:internal-admin   2023-07-17 10:32:05.65547202 -0700 PDT    Spec field updated for App testnew1/mytest1
app/mytest1                              65451fe76cd0bd86eea0103d46a911a5   AppImagePullSuccess   acorn-system           2023-07-17 10:32:05.87267034 -0700 PDT    Pulled test1:v2
app/mytestservice                        fb3bb1be90f4eeb247c77f10c5e8cd55   AppCreate             acorn:internal-admin   2023-07-17 10:40:47.90206473 -0700 PDT    App testnew1/mytestservice created
```

```
 acorn events 7
[...]

PR #1846, opened 2023-06-26
Add support for filtering events by source (#1802)

njhale:
Add support for filtering events by source kind and name to the `acorn events` command.

To this end, the `NAME` argument is renamed `PREFIX` and extended to accept both fully and partially qualified references to event sources.

What does that mean?

A _partially qualified reference_ matches events sourced by any resource of a given kind.

e.g. List events sourced by any app:

```sh
$ acorn events app
```

A _fully qualified reference_ matches events sourced by a specific resource.

e.g. List events sourced by an app named "hello":

```sh
$ acorn events app/hello
``` 

In addition to source references, the `PREFIX` argument also supports filtering events by name prefix.

e.g. List events with names that begin with "4b2b":

```sh
$ acorn events 4b2b
```

Addresses: #1802

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -327,14 +327,14 @@ type EventStreamOptions struct {
 	Tail            int    `json:"tail,omitempty"`
 	Follow          bool   `json:"follow,omitempty"`
 	Details         bool   `json:"details,omitempty"`
-	Name            string `json:"name,omitempty"`
+	Prefix          string `json:"prefix,omitempty"`
 	ResourceVersion string `json:"resourceVersion,omitempty"`
 }
 
 func (o EventStreamOptions) ListOptions() *kclient.ListOptions {
 	fieldSet := make(fields.Set)
-	if o.Name != "" {
-		fieldSet["metadata.name"] = o.Name
+	if o.Prefix != "" {
+		fieldSet["prefix"] = o.Prefix
 	}
 	if o.Details {
 		fieldSet["details"] = strconv.FormatBool(o.Details)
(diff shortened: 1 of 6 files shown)


pkg/client/client.go
```diff
@@ -327,14 +327,14 @@ type EventStreamOptions struct {
...
+	Prefix          string `json:"prefix,omitempty"`
 	ResourceVersion string `json:"resourceVersion,omitempty"`
 }
 
 func (o EventStreamOptions) ListOptions() *kclient.ListOptions {
 	fieldSet := make(fields.Set)
-	if o.Name != "" {
-		fieldSet["metadata.name"] = o.Name
+	if o.Prefix != "" {
+		fieldSet["prefix"] = o.Prefix
```
njhale, line 337: This is a backwards incompatible change. It means a new (i.e. w/ change) client talking to an old (i.e. w/o change) API won't be able to get events by name.

pkg/server/registry/apigroups/acorn/events/strategy.go
```diff
@@ -127,22 +137,33 @@ func (q query) filter(events ...apiv1.Event) []apiv1.Event {
...
+		if !q.prefix.matches(event) {
+			// Exclude from results
+			continue
+		}
+
+		if !q.details {
+			event.Details = nil
+		}
+
+		results = append(results, event)
```
njhale, line 156: Looking for some guidance here.

I think the _correct_ way to do this would be by indexing at the storage layer and querying that index, but I'm not sure if that's something we've exposed in mink yet; especially for remote backed strategies. I scanned through a few times but didn't find anywhere else we're doing something like this.

g-linville: I think this current approach is fine. I'm also not aware of anything in mink that would let us do something like that.

g-linville (approved):

thedadams (approved):

tylerslaton (approved):

Issue #1844, opened 2023-06-24
Go panic when running app in dev mode


lucj:
Hello, 
I'm using Acorn client 0.7.1 on MacOS against a k3s cluster running in a amd64 Linux VM.
When I try to run the app in dev mode, I get a go panic.
Below is the whole output I get.

```
% acorn dev   
INFO[0003] Waiting for builder to start                 
[+] Building 3.2s (5/5) FINISHED                                                                                                                             
 => [internal] load .dockerignore                                                                                                                       0.1s
 => => transferring context: 2B                                                                                                                         0.1s
 => [internal] load build definition from Dockerfile                                                                                                    0.1s
 => => transferring dockerfile: 67B                                                                                                                     0.1s
 => [internal] load metadata for docker.io/library/postgres:15.3-alpine3.18                                                                             1.5s
 => [1/1] FROM docker.io/library/postgres:15.3-alpine3.18@sha256:48d8422c6ae570a5bda52f07548b8e65dd055ac0b661f25b44b20e8cff2f75f0                       0.7s
 => => resolve docker.io/library/postgres:15.3-alpine3.18@sha256:48d8422c6ae570a5bda52f07548b8e65dd055ac0b661f25b44b20e8cff2f75f0                       0.0s
 => => sha256:c47c060e762a35307ec4c09878a5b445789b7006f999ac186d122f99b8f9174e 4.79kB / 4.79kB                                                          0.1s
 => => sha256:31e352740f534f9ad170f75378a84fe453d6156e40700b882d737a8f4a6988a3 3.40MB / 3.40MB                                                          0.2s
 => => sha256:a120e2610875f46aec6c91f34f46574d4cb6a07f9ab469247fe1affccc08a50d 89.93MB / 89.93MB                                                        0.7s
 => => sha256:36cb57831f5201e337de1dd5b1fb077701af0afe39fd7fa8143f9e552eb6ae20 149B / 149B                                                              0.2s
 => => sha256:64f9e9ad23bdf2ce610762af4d1041d79f7f48a7c3fa5a7d822cdae9295bb9b4 9.45kB / 9.45kB                                                          0.1s
 => => sha256:dd2a4281faaa4ada354cc4f1a91ed4f96436ae3412597727da0abeecdb0007f5 162B / 162B                                                              0.1s
 => => sha256:daef310ca2c66fb0048375e8b9bc35d91b315dd9903474749e5246b84bf313b2 203B / 203B                                                              0.1s
[...]

PR #1850, opened 2023-06-26
Fix GenericMap.DeepCopyInto (#1844)

njhale:
Convert GenericMap to and from JSON on DeepCopyInto. This circumvents
the use of runtime.DeepCopyJSON, which panics when int values
exist in the map.

Addresses #1844

### DIFF
--- a/pkg/apis/internal.acorn.io/v1/map.go
+++ b/pkg/apis/internal.acorn.io/v1/map.go
@@ -4,7 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 
-	"k8s.io/apimachinery/pkg/runtime"
+	"github.com/sirupsen/logrus"
 )
 
 type GenericMap map[string]interface{}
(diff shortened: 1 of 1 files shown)


pkg/apis/internal.acorn.io/v1/map.go
```diff
@@ -49,9 +47,8 @@ func (in *GenericMap) UnmarshalJSON(data []byte) error {
 
 func (in *GenericMap) DeepCopyInto(out *GenericMap) {
 	*out = map[string]interface{}{}
-	for k, v := range runtime.DeepCopyJSON(*in) {
-		(*out)[k] = v
-	}
+	data, _ := in.MarshalJSON()
+	_ = out.UnmarshalJSON(data)
```
njhale, line 51: This is probably not the best way to go about this, but it does fix the problem.

thedadams:
As I understand, DeepCopyJSON doesn't support `int` because that type differs based on the platform, but it does support `int64`.

I understand that changing our use of `int` to `int64` would be a much bigger change, but would that give us an advantage here? Should our `ObjectMeta` types just always use `int64`?

Just want to get your thoughts on this.

njhale:
@thedadams thanks for taking a look at this!

> As I understand, DeepCopyJSON doesn't support int because that type differs based on the platform, but it does support int64.

yup, that's my read too

> I understand that changing our use of int to int64 would be a much bigger change, but would that give us an advantage here? Should our ObjectMeta types just always use int64?

I think that's reasonable. After all, there are quite a few places where we're already forced into explicitly using `int64`; for example, the limit field of ListOptions is `int64`.

but for this particular situation -- run args in a GenericMap -- I'm not sure at the moment how we could apply that constraint globally.

ibuildthecloud:
The GenericMap has also assumed that the data is only native json types.  So for example you can put a struct in there. I'm fine with not allowing `int`.  If we really really want it, we can just copy the deepcopy function and add int.  That exact logic I've written tons of time and you will probably find it at least twice somewhere in acorn/baaah already.

I believe the reason `int` isn't supported is because json.Unmarshal will never produce an int.  It only creates int64, float64, json.Number.

thedadams (approved):

tylerslaton (approved):

PR #1852, opened 2023-06-27
Update mac-main-release to use BuildJet

tylerslaton:
Updating the `notarize` script to use cross-platform OSS libraries. This makes it so that the `mac-main-release` action can run on BuildJet instead of requiring a Mac.

### Checklist


tools/notarize
```diff
@@ -27,10 +27,16 @@ echo "SIGN=${SIGN} NOTARIZE=${NOTARIZE} BUNDLE=${AC_BUNDLE} BINARY=${BINARY} DMG
 if [[ "${SIGN}" == 1 ]]; then
   echo "Signing"
 
-  echo "${AC_P12}" | base64 --decode > signing.p12
-  security import signing.p12 -P "${AC_P12_PASSWORD}" -A -t cert -f pkcs12 -k $KEYCHAIN || true
+  # Install the necessary tools, specifically cargo, to install rcodesign, a Rust 
+  # implementation of codesign.
+  which curl || (apt-get update && apt-get install curl)
+  which cargo || curl https://sh.rustup.rs -sSf | sh -s -- -y && bash
+  which gcc || apt-get install gcc
```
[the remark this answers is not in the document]
tylerslaton: I see your point, yes `apt-get update` should run no matter what. I'll change that.

tools/notarize
```diff
@@ -39,22 +47,22 @@ fi
...
-  cat <<EOF > "gon.hcl"
-notarize {
-  path = "${DMG}"
-  bundle_id = "${AC_BUNDLE}"
-  staple = true
-}
-EOF
+  # Build the DMG
+  cp LICENSE README.md "${DIR}/"
+  mkfile -n 1g "${DMG}"
```
vincent99, line 54: The size here should be based on the (sum of the) files you're going to write into it, + a bit of buffer for filesystem.

~Use `seek=(size needed) count=` instead of actually reading/writing the whole count.~ ok mkfile works the same.

The previous `hdiutil` command is making a `UDZO` = compressed read-only image, so 1GB is a mostly meaningless max size and any space you don't use compresses to ≈ nothing.  I don't know of any tools to create this kind in linux, so you'd be writing a full gig image every time here right now.

tools/notarize
```diff
@@ -13,48 +13,52 @@ DIR="releases/mac_darwin_all"
...
-EOF
+  # Build the DMG
+  cp LICENSE README.md "${DIR}/"
+  SIZE="$(du -s ${DIR} | awk '{print $1 + 1024*1024}')" # The size of the directory + 1MB for any overhead
+  mkfile -n "${SIZE}" "${DMG}"
+  mkfs.hfsplus -v "Acorn" "${DMG}"
+  mkdir -p /tmp/acorn_mount
+  sudo mount -t hfsplus -o loop "${DMG}" /tmp/acorn_mount
+  cp -R "${DIR}"/* /tmp/acorn_mount
+  sudo umount /tmp/acorn_mount
```
vincent99, line 55: Presumably the `sudo`s are unnecessary here since you're doing apt-get installs and such above.

vincent99 (approved):

thedadams (approved):

tools/notarize
tylerslaton: Good point. We'll have to see what permissions BuildJet's default user is but you're right that these should be consistent.

PR #1857, opened 2023-06-27
Temporarily disable signing and notarization for main releases

tylerslaton:


thedadams (approved):

tylerslaton:
This was done to unblock some other critical path work. I'll be fixing it on the side to not be broken.

Issue #752, opened 2022-10-06
Reduce log level of unhelpful logs and allow for dynamic control of acorn log level log level of acorn


cjellick:
This enhancement is a two parter:

- First, provide a way to dynamically set the log level of the acorn controller or api server. As an example, rancher did this via: https://github.com/rancher/loglevel/tree/master and https://github.com/rancher/log/blob/master/server/server.go. This allows an operator to run a command against the pod that will change log level with out redploying
- Second, turn the following logs down from info to debug:
```
time="2023-06-12T22:46:09Z" level=info msg="Handling [acorn-image-system/registry] [/v1, Kind=Service]"
time="2023-06-12T22:46:09Z" level=info msg="Triggering [acorn-system/acorn-config] [/v1, Kind=ConfigMap] from [acorn-image-system/registry] [/v1, Kind=Service]"
```
These logs live in mink, I believe.

cjellick:
So, the requirement is to be able to change the log level of the acorn api server or controller without restarting the pod

g-linville:
It should now be possible to shell into the Runtime apiserver and controller and run `loglevel --set <level>` to change the log level without restarting the pod.

sangee2004:
Tested with acorn version - `v0.8.0-alpha7-131-g4bf05b4d+4bf05b4d`

1. Able to change log levels by executing the following command within the controller and api-server pods

`kubectl exec acorn-controller-7cc79ddc88-2f5td  -n acorn-system -- loglevel --set debug`
`kubectl exec acorn-api-55494b75d5-62msg -n acorn-system -- loglevel --set debug`



2. Following logs are now logged in "debug" level:
```
time="2023-07-20T21:00:48Z" level=debug msg="Handling trigger [acorn-image-system/registry] [apps/v1, Kind=Deployment]"
time="2023-07-20T21:00:48Z" level=debug msg="Triggering [acorn-system/acorn-config] [/v1, Kind=ConfigMap] from [lima-rancher-desktop] [/v1, Kind=Node]"
```

PR #1861, opened 2023-06-28
Use loglevel to dynamically change logrus' log level during runtime (#752)

g-linville:
for #752

This adds a little server to the controller and the api-server that listens on a local socket for requests to change the log level, and sets the logrus log level accordingly.

This also adds the `loglevel` binary to `/usr/local/bin` in the container image, which can be used like `loglevel --set info` to set the log level, or just `loglevel` to print the current log level.

This also bumps the baaah dependency in order to get the new changes that reduce the log severity on some of the common logs, as also described in #752.

### Checklist

### DIFF
--- a/pkg/cli/apiserver.go
+++ b/pkg/cli/apiserver.go
@@ -43,7 +43,9 @@ func (a *APIServer) Run(cmd *cobra.Command, args []string) error {
 	}
 
 	logrus.SetLevel(logrus.InfoLevel)
-	logserver.StartServerWithDefaults()
+	if err := logserver.StartServerWithDefaults(); err != nil {
+		logrus.Warnf("failed to start log server: %v", err)
+	}
 
 	<-cmd.Context().Done()
 	return cmd.Context().Err()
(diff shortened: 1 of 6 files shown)


iwilltry42 (approved):

pkg/cli/apiserver.go
```diff
@@ -40,6 +42,9 @@ func (a *APIServer) Run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	logrus.SetLevel(logrus.InfoLevel)
```
thedadams, line 45: Can this be added to the `StartServerWithDefaults()`?

g-linville:
Putting this into draft until acorn-io/loglevel#1 is merged, since this is now dependent on that since we are switching to an abstract namespace unix socket.

tylerslaton (approved):
Looks great! Can you help me understand how the log level gets changed? I see that we have the code infrastructure in place to set it and am looking for how I trigger that outside of the endpoint.

g-linville:
> Looks great! Can you help me understand how the log level gets changed? I see that we have the code infrastructure in place to set it and am looking for how I trigger that outside of the endpoint.

@tylerslaton the only way to change it is to exec into the pod and run the `loglevel` command. From the issue description:

> First, provide a way to dynamically set the log level of the acorn controller or api server. As an example, rancher did this via: https://github.com/rancher/loglevel/tree/master and https://github.com/rancher/log/blob/master/server/server.go. This allows an operator to run a command against the pod that will change log level with out redploying

thedadams (approved):

PR #1863, opened 2023-06-28
Add --profile and component resources flags

tylerslaton:
This PR adds a few new install flags: 
* `--profile`.
* `--controller-memory`
* `--controller-cpu`
* `--api-server-memory`
* `--api-server-cpu`
* `--buildkitd-memory`
* `--buildkitd-cpu`
* `--buildkitd-service-cpu`
* `--buildkitd-service-memory`
* `--registry-memory`
* `--registry-cpu`

### Specifying resources for runtime components
With this update we can now set component resources via their respective install flag. Each component follows this format:

```console
acorn install --<component>-<resource> <req>:<limit>`
```

Alternatively, you can just pass the `req` like so.

```console
acorn install --<component>-<resource> <req>
```

### Profile-based defaults
Introducing the ability to set resource flags also encourages having logical defaults for production environments while keeping the default install free of install assumptions (like resource capacity). With this PR, there are two profiles: `production` and `default`.

You can specify a profile with the following command.

```console
acorn install --profile production
```

### Checklist


pkg/system/resources.go
```diff
@@ -1,76 +1,73 @@
...
-		return quantity
+		parsedReq, err := resource.ParseQuantity(reqString)
+		if err == nil {
+			requirements.Requests[resourceType] = parsedReq
+		}
+
+		parsedLimit, err := resource.ParseQuantity(limitString)
+		if err == nil {
+			requirements.Limits[resourceType] = parsedLimit
+		}
```
g-linville, line 69: Just want to make sure that we are intentionally ignoring all errors here.

pkg/system/resources.go
```diff
@@ -1,76 +1,73 @@
...
-	registryMemoryRequest = *resource.NewQuantity(128*mi, resource.BinarySI)    // REGISTRY_MEMORY_REQUEST
-	registryMemoryLimit   = *resource.NewQuantity(512*mi, resource.BinarySI)    // REGISTRY_MEMORY_LIMIT
-	registryCPURequest    = *resource.NewMilliQuantity(200, resource.DecimalSI) // REGISTRY_CPU_REQUEST
+var ErrInvalidResourceSpecification = fmt.Errorf("invalid resource specification")
 
-	buildkitdMemoryRequest = *resource.NewQuantity(256*mi, resource.BinarySI)    // BUILDKITD_MEMORY_REQUEST
-	buildkitdMemoryLimit   = *resource.NewQuantity(10*gi, resource.BinarySI)     // BUILDKITD_MEMORY_LIMIT
-	buildkitdCPURequest    = *resource.NewMilliQuantity(800, resource.DecimalSI) // BUILDKITD_CPU_REQUEST
+func ValidateResources(resources ...string) error {
+	// TODO - fix
```
g-linville, line 15: Does something still need to be changed or was this left behind?

g-linville (approved):
Lgtm, just a few small things.

Also, a question. With how this works, can I specify a profile and then continue to override configuration options on top of it? Say I want to use the production profile but my ingress controller lives in `nginx` instead of `traefik`. Could I do `acorn install --profile production --ingress-controller-namespace nginx`?

pkg/system/resources.go
tylerslaton: This was left behind.

pkg/imagesystem/registrytemplate.go
```diff
@@ -38,7 +39,7 @@ func registryService(namespace string) []client.Object {
 	}
 }
 
-func registryDeployment(namespace, registryImage string) []client.Object {
+func registryDeployment(namespace, registryImage string, cfg *apiv1.Config) []client.Object {
```
njhale, line 42: nit: This function only really depends on a small portion of `Config`; a more concise  signature would look something like this:

```go
func registryDeployment(namespace, registryImage string, requirements corev1.ResourceRequirements) []client.Object
```

(see also https://en.wikipedia.org/wiki/Interface_segregation_principle)

njhale (approved):
Left one suggestion, but not blocking

tylerslaton:
> @g-linville Also, a question. With how this works, can I specify a profile and then continue to override configuration options on top of it? Say I want to use the production profile but my ingress controller lives in nginx instead of traefik. Could I do acorn install --profile production --ingress-controller-namespace nginx?

Yes! That's the intent here. Basically the way that it is coded is that any default values will be replaced but any user specified ones will be respected. So if a user has not specified anything, all of the `profile` values will be used. But If you specify `-ingress-controller-namespace nginx` that will always be respected (even if you change profiles).

pkg/system/resources.go
tylerslaton: It's a good thing to call out. My thought process was that when this is called its likely in areas that cannot/should not error. For example, this area of code.

https://github.com/tylerslaton/runtime/blob/ed7fae64670fc4907a7fd006fabe3c96955cb704/pkg/imagesystem/buildertemplate.go#L58-L59

This makes it so that we don't `panic` in these situations but I would like your opinion here. The main scenario this presents as an annoyance is if the user manually edits or creates an `apiv1.Config` and provides an invalid value. In this situation they would receive an `0` amount.

pkg/imagesystem/registrytemplate.go
tylerslaton: I believe this is addressed. What do you think?

g-linville:
Also @tylerslaton I think it would be good to add something to the docs explaining how `--profile` works.

pkg/system/resources.go
g-linville: Hmm...I think it is okay to leave it as is :+1:

tylerslaton:
@g-linville Added some documentation. Let me know what you think!

pkg/system/resources.go
```diff
@@ -1,76 +1,76 @@
...
-	buildkitdServiceMemoryLimit   = *resource.NewQuantity(256*mi, resource.BinarySI)    // BUILDKITD_SERVICE_MEMORY_LIMIT
-	buildkitdServiceCPURequest    = *resource.NewMilliQuantity(200, resource.DecimalSI) // BUILDKITD_SERVICE_CPU_REQUEST
-)
+		req, limit, both := strings.Cut(rs, ":")
+		if !both {
+			_, err := resource.ParseQuantity(rs)
+			if err != nil {
+				return err
+			}
+			return nil
```
g-linville, line 28: nit:

```suggestion
			return err
```

docs/docs/30-installation/02-options.md
```diff
@@ -134,4 +134,11 @@ For strings array flags, you can reset the slice to empty by pass empty string "
...
 acorn install --propagate-project-annotation ""
-```
\ No newline at end of file
+```
+
+## Install Profiles
+When you are installing Acorn, you can specify a profile to use. A profile is a set of installation flag defaults that are pre-defined. You can see the list of available profiles by running `acorn install --help`. The default profile is `default`, and it is used if no profile is specified.
+
+:::note
+Once a profile is set, this will set new default values based on the profile. Any default values previously used will be switched to the new profile defaults. However, any install flags that were specified, or have specified, will still be respected.
```
g-linville, line 143:
```suggestion
Once a profile is set, this will set new default values based on the profile. Any default values previously used will be switched to the new profile defaults. However, any install flags that were or are specified will still be respected.
```

pkg/profiles/default.go
```diff
@@ -0,0 +1,69 @@
...
+	InternalClusterDomainDefault = "svc.cluster.local"
+
+	AcornDNSEndpointDefault = "https://oss-dns.acrn.io/v1"
+	AcornDNSStateDefault    = "auto"
+
+	// LetsEncryptOptionDefault is the default state for the Let's Encrypt integration
+	LetsEncryptOptionDefault = "disabled"
+
+	// DefaultImageCheckIntervalDefault is the default value for the DefaultImageCheckInterval field
+	DefaultImageCheckIntervalDefault = "5m"
```
g-linville, line 17: nit: this seems like a weird name to me

```suggestion
	// AutoUpgradeIntervalDefault is the default value for the AutoUpgradeInterval field
	AutoUpgradeIntervalDefault = "5m"
```

(would also need to update this in other places in this file)

pkg/profiles/default.go
```diff
@@ -0,0 +1,69 @@
...
+
+	// Default HttpEndpointPattern set to enable Let's Encrypt
```
g-linville, line 19: nit: this comment seems incorrect

g-linville (approved):
Just a few little nits

pkg/system/resources.go
```diff
@@ -1,76 +1,73 @@
...
-			corev1.ResourceMemory: envOrDefault("REGISTRY_MEMORY_LIMIT", registryMemoryLimit),
-		},
+		_, err := resource.ParseQuantity(req)
+		if err != nil {
+			return err
+		}
+		_, err = resource.ParseQuantity(limit)
+		if err != nil {
+			return err
+		}
```
thedadams, line 35: Is it important to validate that limit is greater than request?

pkg/system/resources.go
```diff
@@ -1,76 +1,73 @@
...
-func BuildkitdServiceResources() corev1.ResourceRequirements {
-	return corev1.ResourceRequirements{
-		Requests: corev1.ResourceList{
-			corev1.ResourceMemory: envOrDefault("BUILDKITD_SERVICE_MEMORY_REQUEST", buildkitdServiceMemoryRequest),
-			corev1.ResourceCPU:    envOrDefault("BUILDKITD_SERVICE_CPU_REQUEST", buildkitdServiceCPURequest),
-		},
-		Limits: corev1.ResourceList{
-			corev1.ResourceMemory: envOrDefault("BUILDKITD_SERVICE_MEMORY_LIMIT", buildkitdServiceMemoryLimit),
-		},
+	var resources = map[corev1.ResourceName]string{
```
thedadams, line 45: nit:
```suggestion
	resources := map[corev1.ResourceName]string{
```

thedadams (approved):

pkg/system/resources.go
tylerslaton: I agree. I updated this to now check that the request is not greater than the limit.

pkg/system/resources.go
tylerslaton: Slight deviation but I believe I captured the essence of what you were looking for here!

thedadams (approved):

PR #1880, opened 2023-06-30
Switch to use a zip instead of a DMG

tylerslaton:
This PR turns off notarization and stapling. The notarized DMG or Zip files are not necessary for our current installation methods of WGET or Homebrew. I've left some comments that explains further.

In addition, this always switches the notarization logic to create a `zip` file instead of a `dmg`. By using a `zip` file we are able to notarize on Linux without the complexity of creating a `dmg` file on the OS. Instead we can get the 90% of the way there by using and notarizing a `zip` file. The 10% we are missing out on is stapling. This is a feature that allows for binaries to be distributed and verified by MacOS offline. Since you need the internet to download the zip file this seems like an unnecessary complexity addition. 

### Checklist

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -37,7 +37,7 @@ universal_binaries:
         - cmd: ./tools/notarize "{{ .Path }}" "acorn-v{{ .Version }}-macOS-universal"
           output: true
           env:
-            - NOTARIZE={{ if index .Env "NOTARIZE" }}{{ .Env.NOTARIZE }}{{ end }}
+            - NOTARIZE=0 # Always disable Notarization since it is not currently necessary.
             - AC_IDENTITY={{ if index .Env "AC_IDENTITY" }}{{ .Env.AC_IDENTITY }}{{ end }}
             - AC_PROVIDER={{ if index .Env "AC_PROVIDER" }}{{ .Env.AC_PROVIDER }}{{ end }}
             - AC_USERNAME={{ if index .Env "AC_USERNAME" }}{{ .Env.AC_USERNAME }}{{ end }}
--- a/tools/notarize
+++ b/tools/notarize
@@ -43,6 +43,22 @@ rcodesign sign \
   "${BINARY}"
 echo "Signed ${BINARY}!"
 
+# Currently, NOTARIZE is always set to 0. In effect, this means that the
+# binary is signed, but not zipped up, notarized or stapled. Since all of the 
+# official installation processes are through brew or wget, the notarized DMG we were
+# creating was not necessary. Where this does become a problem is if a user 
+# wants to install from a browser. In this situation, there will need to be 
+# a notarization process otherwise MacOS will mark our binary as unsafe and 
+# the user will need to go through hoops to install it. In this instance,
+# setting NOTARIZE to 1 will create a ZIP file (since it was difficult to create
+# a DMG while not on MacOS) and Notarize it.
+#
+# Note - If you want to staple that ZIP, you will need to staple each individual
+#        item in the zip file.
+#
+# References: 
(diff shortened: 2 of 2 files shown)


tools/notarize
```diff
@@ -44,21 +44,15 @@ rcodesign sign \
...
-  sudo mount -t hfsplus -o loop "${DMG}" /tmp/acorn_mount
-  sudo cp -R "${DIR}"/* /tmp/acorn_mount
-  sudo umount /tmp/acorn_mount
-  echo "Built ${DMG}!"
-
-  # Notarize and staple the DMG
+  zip -r "${ZIP}" "${DIR}"
+  echo "Built ${ZIP}!"
+
+  # Notarize and staple the ZIP
```
njhale, line 55:
```suggestion
  # Notarize the ZIP
```

tools/notarize
```diff
@@ -44,21 +44,15 @@ rcodesign sign \
 echo "Signed ${BINARY}!"
 
 if [[ "${NOTARIZE}" == "1" ]]; then
-  which mkfs.hfsplus || sudo apt-get install hfsprogs -y
+  which zip || sudo apt-get zip -y
```
g-linville, line 47: nit:

```suggestion
  which zip || sudo apt-get install zip -y
```

g-linville (approved):

.goreleaser.yml
```diff
@@ -37,7 +37,7 @@ universal_binaries:
         - cmd: ./tools/notarize "{{ .Path }}" "acorn-v{{ .Version }}-macOS-universal"
           output: true
           env:
-            - NOTARIZE={{ if index .Env "NOTARIZE" }}{{ .Env.NOTARIZE }}{{ end }}
+            - NOTARIZE={{ if index .Env "NOTARIZE" }}0{{ end }} # Always disable Notarization since it is not currently necessary.
```
g-linville, line 40: not 100% sure but I think just this would work:

```suggestion
            - NOTARIZE=0 # Always disable Notarization since it is not currently necessary.
```

g-linville (approved):

.goreleaser.yml
tylerslaton: Yeah, that'll work. The main difference here is that you can't pass and environment variable to the notarize script but in this context that's fine.

tools/notarize
```diff
@@ -43,22 +43,26 @@ rcodesign sign \
...
+# the official installation processes are through brew or wget, this will not
+# be a problem.
+#
+# Where this does become a problem is if a user wants to install from a browser.
+# In this situation, unless there will need to be a notarization process otherwise 
+# MacOS will mark our binary as unsafe and the user will need to go through hoops
+# to install it.
 if [[ "${NOTARIZE}" == "1" ]]; then
-  which mkfs.hfsplus || sudo apt-get install hfsprogs -y
+  which zip || sudo apt-get install zip -y
```
thedadams, line 57: The comment above indicates that it is not possible to notarize the `zip` file, but then the changes indicate that it may be possible?

Would it make sense to leave the `DMG` stuff in place? It seems more likely that there would be a way to notarize the `DMG` before Apple accepts notarization of a `zip`.

tylerslaton: My comment is misleading in this sense. [You can notarize a `zip` file](https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution/customizing_the_notarization_workflow#:~:text=While%20you%20can%20notarize%20a%20ZIP%20archive%2C%20you%20can%E2%80%99t%20staple%20to%20it%20directly.), I'll make that comment clearer.

That being said, the old DMG logic wasn't working due to us not "properly" creating a DMG on Linux. So if we ever came back to use that code the DMG creation would not work. Thoughts?

thedadams (approved):

Issue #1459, opened 2023-04-07
auto-upgrade not outputted as part of acorn run -o yaml


cjellick:

tylerslaton:
Gave this a look and this isn't a direct result of adding `-o`, `auto-upgrade` actually something that isn't set on the spec unless explicitly specified by the user. The daemon for `auto-upgrade` then makes assumptions based on what subset of those values are present.

For example, if you run an app without the `-o`, it will also have the `auto-upgrade` spec value not set but you will still get auto-upgrades. So the options I see are:
1. Move the daemon's "assumption" logic into `run.go` so that all apps that will be ran with `auto-upgrade` will have that field set regardless if it is explicitly set by the user.
2. Leave it the same and have consumers of `-o` yaml do that "assumption" logic.

For reference what I'm calling the "assumption logic" can be found [here](https://github.com/acorn-io/acorn/blob/487206457a1157a90ae4ac95224cdcdbe1b9adb1/pkg/autoupgrade/daemon.go#L330-L331).

Thoughts @cjellick?

cjellick:
Pretty sure I want to change this behavior. I think this also affects the UI. Like, @vincent99 may not know to mark an app as autoupgrade if that flag isnt set.

I think the behavior I want is that the autoUpgrade field *must* be set, but the CLI helps - if the tag is detected to be a pattern, set the autoUpgrade field for the user. This should be done client side, not server side.

tylerslaton:
The backend code for this has been set in place. Now the UI just needs to read from `AppInstance.Status.ObservedAutoUpgrade` in order to properly display when auto-upgrade is enabled.

PR #1881, opened 2023-07-02
Always set AutoUpgrade if it is implied (#1459)

tylerslaton:
for #1459 

Prior to this commit, `AutoUpgrade` would not get set for Apps that had implied `AutoUpgrade` behavior. For example, if an `App` came through with `AutoUpgradeInterval` set or a pattern set for its `Image` then `AutoUpgrade` would be false despite it actually being enabled.

### Checklist

### DIFF
--- a/pkg/cli/run.go
+++ b/pkg/cli/run.go
@@ -10,6 +10,7 @@ import (
 
 	apiv1 "github.com/acorn-io/runtime/pkg/apis/api.acorn.io/v1"
 	v1 "github.com/acorn-io/runtime/pkg/apis/internal.acorn.io/v1"
+	"github.com/acorn-io/runtime/pkg/autoupgrade"
 	cli "github.com/acorn-io/runtime/pkg/cli/builder"
 	"github.com/acorn-io/runtime/pkg/client"
 	"github.com/acorn-io/runtime/pkg/dev"
@@ -237,6 +238,11 @@ func (s *Run) Run(cmd *cobra.Command, args []string) (err error) {
 	opts, err := s.ToOpts()
 	if err != nil {
 		return err
+	}
+
+	// Ensure opts.AutoUpgrade is set if is implied by opts.AutoUpgradeInterval or imageSource.Image's pattern.
+	if opts.AutoUpgrade == nil || !*opts.AutoUpgrade {
+		opts.AutoUpgrade = &[]bool{autoupgrade.ImpliedAutoUpgrade(opts.AutoUpgradeInterval, imageSource.Image)}[0]
 	}
 
 	// Force install prompt if needed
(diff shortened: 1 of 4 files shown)


g-linville (approved):

pkg/cli/run.go
```diff
@@ -239,6 +240,11 @@ func (s *Run) Run(cmd *cobra.Command, args []string) (err error) {
 		return err
 	}
 
+	// Ensure opts.AutoUpgrade is set if is implied by opts.AutoUpgradeInterval or imageSource.Image's pattern.
+	if opts.AutoUpgrade == nil || !*opts.AutoUpgrade {
+		opts.AutoUpgrade = &[]bool{autoupgrade.ImpliedAutoUpgrade(opts.AutoUpgradeInterval, imageSource.Image)}[0]
+	}
```
thedadams, line 246: Quote from @cjellick:
> I think the behavior I want is that the autoUpgrade field must be set, but the CLI helps - if the tag is detected to be a pattern, set the autoUpgrade field for the user. This should be done client side, not server side.

This looks like the "CLI helps" portion, but there is nothing here to require that this field is set.

thedadams: I interpreted the statement that "the autoUpgrade field must be set" to mean that we should require the `autoUpgrade` field be set if the user is trying to use an autoupgrade pattern. Additionally, "but the CLI helps" to mean that if the user hasn't set this field, then the CLI will help them out. I understand that you have done the second part here.

Do we want to require that the `autoUpgrade` field is set whenever an autoupgrade tag is used? For example, as I understand it, any app created outside of the CLI may not satisfy this requirement.

tylerslaton: > Do we want to require that the autoUpgrade field is set whenever an autoupgrade tag is used?
Yes! That is definitely something we want. I'll code some validation for this into the API Server that will enforce Apps to have `AutoUpgrade` set when its use is assumed.

tylerslaton: @cjellick Just for context on this discussion.

g-linville (approved):

pkg/server/registry/apigroups/acorn/apps/validator.go
```diff
@@ -96,6 +96,12 @@ func (s *Validator) Validate(ctx context.Context, obj runtime.Object) (result fi
 		return
 	}
 
+	au := app.Spec.AutoUpgrade
+	if au != nil && !*au && autoupgrade.ImpliedAutoUpgrade(app.Spec.AutoUpgradeInterval, app.Spec.Image) {
```
thedadams, line 100: Should this be:
```suggestion
	if (au == nil || !*au) && autoupgrade.ImpliedAutoUpgrade(app.Spec.AutoUpgradeInterval, app.Spec.Image) {
```

tylerslaton: Yes! It should be.

tylerslaton: Added this and updated a test that failed as a result of this being added.

thedadams (approved):

Issue #1809, opened 2023-06-19
acorn copy command


cjellick:
We want the equivalent of `crane copy`, but for acorn images, built right into the acorn cli

sangee2004:
Tested with acorn version - ` v0.8.0-alpha7-114-gbf4c72ea+bf4c72ea`
Able to use `acorn copy` command for the following scenarios:
1. copy an image from the internal registry to a remote registry 
2. copy images between remote registries
3. copy entire repositories between remote registries (with the -a option)

@g-linville We are not allowed to copy all tags of an image from  internal registry to a remote registry using "-a" option. 
Do we expect for this to succeed ? If not , can the error message for this be improved ?

```
acorn images mytestssh
REPOSITORY   TAG       IMAGE-ID
mytestssh    v0.0.1    dd805b801e98
mytestssh    v0.0.3    c5d1f2026afc
mytestssh    v0.0.2    48b4edb63751

 % acorn copy -a mytestssh ghcr.io/<accountname>/sshtestp
  ✗  ERROR:  websocket: close 1011 (internal server error): [ags/list?n=1000: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:library/mytestssh Type:repository]]]
```

g-linville:
@sangee2004 yeah, `--all-tags` is only valid with remote registries and not the local one. I'm hesitant to look for this and catch the error in case the user has a local image that matches the name of a remote registry and it might cause them issues.

For example, say I have a remote Acorn image at `docker.io/grantlinville/myacorn:v1`. I build and push `docker.io/grantlinville/myacorn:latest`. Now I have a local image called `docker.io/grantlinville/myacorn:latest`. If I were to then attempt to copy both tags from my Docker Hub to my GHCR by running `acorn copy --all-tags docker.io/grantlinville/myacorn ghcr.io/g-linville/myacorn`, the apiserver would get confused and think that I am wrongly referring to a local image, when I am actually trying to refer to a valid remote repository. So I would prefer to leave this as is.

`acorn copy --help` indicates that `--all-tags` does not work with local images.

sangee2004:
Tested with acorn version `v0.8.0-rc4-58-gbe49d0bd+be49d0bd`

1. Acorn copy works only with local credentials now.Need to make error messages relating to this more obvious to users tracked in https://github.com/acorn-io/runtime/issues/2119

2. acorn copy with source being a tag does not work. I am not sure how this is suppose to work.

```
[...]

PR #1883, opened 2023-07-05
Add acorn copy command (#1809)

g-linville:
for #1809

This implements the `acorn copy` (alias `acorn image copy`) command, which behaves similarly to `crane copy`. It can be used to:

- copy an image from the internal registry to a remote registry (i.e. `acorn push`)
- copy images between remote registries
- copy entire repositories between remote registries (with the `--all-tags` option)

This command cannot do the following:

- copy an image from a remote registry to the internal registry (i.e. `acorn pull`)
- copy images between projects in the internal registry
- re-tag images in the internal registry (i.e. `acorn tag`)
- copy entire "repositories" from the internal registry to somewhere else
  - since our "tags" on local images really only exist in the API-level and not in the registry itself, there is not a clean way to be able to copy a whole "repo" of images sourced from the internal registry

By default, this command will avoid overwriting remote tags that already exist unless `--force` is specified.

I left comments on some areas of the code which should hopefully clarify things.

### Checklist


pkg/client/client.go
```diff
@@ -177,9 +177,10 @@ func (a AppUpdateOptions) ToRun() AppRunOptions {
 }
 
 type ImageProgress struct {
-	Total    int64  `json:"total,omitempty"`
-	Complete int64  `json:"complete,omitempty"`
-	Error    string `json:"error,omitempty"`
+	Total       int64  `json:"total,omitempty"`
+	Complete    int64  `json:"complete,omitempty"`
+	Error       string `json:"error,omitempty"`
+	CurrentTask string `json:"currentTask,omitempty"`
```
g-linville, line 183: I added this field so that we can tell the user what is currently happening. This is very useful in particular for `--all-tags`.

pkg/imagesystem/registry.go
```diff
@@ -189,7 +189,7 @@ func isNotInternalRepo(prefix, image string) error {
 	}
 
 	if prefix == "" {
-		return nil
+		prefix = fmt.Sprintf("%s.%s", system.RegistryName, system.ImagesNamespace)
```
g-linville, line 192: I updated this check so that it works with the default internal registry (`registry.acorn-image-system`) if nothing else is specified.

pkg/progressbar/print.go
```diff
@@ -47,10 +48,23 @@ func Print(progress <-chan client.ImageProgress) error {
...
 			}
 
+			if update.CurrentTask != "" && update.CurrentTask != currentTask {
+				if bar != nil {
+					bar.Add(bar.Total - bar.Current)
+					_, _ = bar.Stop()
+					bar = nil
+				}
+				currentTask = update.CurrentTask
+			}
```
g-linville, line 58: Here, we reset the progress bar if the current task has changed, since we need to start a new bar for the new task.

pkg/server/registry/apigroups/acorn/images/copy.go
```diff
@@ -0,0 +1,353 @@
...
+	// metachannel is used to send another channel with updates for each tag to be copied
+	metachannel := make(chan chan ggcrv1.Update)
+	// progress is the channel returned by this function and used to write websocket messages to the client
+	progress := make(chan ImageProgress)
+	wg := sync.WaitGroup{}
+	wg.Add(1)
+
+	// We create an RWMutex for the currentTag variable, since it is used by both goroutines
+	m := sync.RWMutex{}
+	currentTag := ""
```
g-linville, line 261: This, along with the two goroutines that follow, is definitely the grossest part of this code. Would it be worth it to split it out into properly defined functions with parameters, instead of just closures like they are now?

I hate that I had to create a mutex for this, but a shared variable was the only way I could come up with for one goroutine to communicate to the other which tag was currently being copied.

pkg/server/registry/apigroups/acorn/images/copy.go
```diff
@@ -0,0 +1,353 @@
...
+
+	if err := imagesystem.IsNotInternalRepo(ctx, i.client, namespace, args.Dest); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (i *ImageCopy) ImageCopy(ctx context.Context, namespace, sourceImage, destImage string, srcAuth, dstAuth *apiv1.RegistryAuth, force bool) (<-chan ImageProgress, error) {
+	sourceRef, err := i.getImageReference(ctx, sourceImage, namespace)
```
g-linville, line 129: The source is allowed to be a local image, so we use this `getImageReference` function (defined at the bottom of this file), which checks locally and remotely if it wasn't found.

pkg/server/registry/apigroups/acorn/images/copy.go
```diff
@@ -0,0 +1,353 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if sourceRef.Context().RegistryStr() == images.NoDefaultRegistry {
+		return nil, fmt.Errorf("missing registry name (i.e. docker.io, ghcr.io) from source %s", sourceImage)
+	}
+
+	destRef, err := name.ParseReference(destImage, name.WithDefaultRegistry(images.NoDefaultRegistry))
```
g-linville, line 138: The destination is not allowed to be local, so we just parse the reference normally and ensure that there is an explicitly defined registry. Since `acorn push` doesn't implicitly use Docker Hub, I didn't want to implicitly use it here either, in order to be consistent.

pkg/client/image.go
```diff
@@ -63,6 +63,64 @@ func (c *DefaultClient) ImageDetails(ctx context.Context, imageName string, opts
...
+		body.SourceAuth = opts.SourceAuth
+		body.DestAuth = opts.DestAuth
+		body.AllTags = opts.AllTags
+		body.Force = opts.Force
+	}
+
+	url := c.RESTClient.Get().
+		Namespace(c.Namespace).
+		Resource("images").
+		Name(strings.ReplaceAll(fmt.Sprintf("%s-%s", srcImage, dstImage), "/", "+")).
```
g-linville, line 82: The name parameter here is not used server-side at all.

pkg/server/server.go
```diff
@@ -61,7 +61,7 @@ func New(cfg Config) (*server.Server, error) {
 		Version:               cfg.Version,
 		HTTPSListenPort:       7443,
 		LongRunningVerbs:      []string{"watch", "proxy"},
-		LongRunningResources:  []string{"exec", "proxy", "log", "registryport", "port", "push", "pull", "portforward"},
+		LongRunningResources:  []string{"exec", "proxy", "log", "registryport", "port", "push", "pull", "portforward", "copy"},
```
g-linville, line 64: This was needed in order to avoid hitting context deadlines.

thedadams: Great job figuring this out!

pkg/server/registry/apigroups/acorn/images/copy.go
thedadams: I think this is better off as a comment in the code.

pkg/server/registry/apigroups/acorn/images/copy.go
thedadams: Same here.

pkg/server/registry/apigroups/acorn/images/copy.go
```diff
@@ -0,0 +1,353 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if sourceRef.Context().RegistryStr() == images.NoDefaultRegistry {
```
thedadams, line 143: This looks like the same check on line 134. Is this one necessary?

pkg/server/registry/apigroups/acorn/images/copy.go
```diff
@@ -0,0 +1,353 @@
...
+				CurrentTask: fmt.Sprintf("Copying %s to %s", sourceImage, destImage),
+			}
+		}
+	}()
+
+	go func() {
+		defer func() {
+			wg.Wait()
+			close(progress2)
+		}()
```
thedadams, line 196: nit: this (and the `WaitGroup`) are not necessary. You can close the `progress2` channel at the end of the above goroutine.

pkg/server/registry/apigroups/acorn/images/copy.go
```diff
@@ -0,0 +1,353 @@
...
+	return nil
+}
+
+func (i *ImageCopy) ImageCopy(ctx context.Context, namespace, sourceImage, destImage string, srcAuth, dstAuth *apiv1.RegistryAuth, force bool) (<-chan ImageProgress, error) {
+	sourceRef, err := i.getImageReference(ctx, sourceImage, namespace)
+	if err != nil {
+		return nil, err
+	}
+
+	if sourceRef.Context().RegistryStr() == images.NoDefaultRegistry {
```
thedadams, line 134: If `sourceRef` is allowed to be local, should this be OK?

pkg/server/registry/apigroups/acorn/images/copy.go
```diff
@@ -0,0 +1,353 @@
...
+
+	go func() {
+		defer func() {
+			close(metachannel)
+			wg.Wait()
+			close(progress)
+		}()
+
+		for i, imageIndex := range sourceIndexes {
+			currProgress := make(chan ggcrv1.Update)
```
thedadams, line 291: How do you feel about adding a comment here that `remote.WithProgress` will close this channel? Otherwise, it looks like it leaks.

pkg/server/registry/apigroups/acorn/images/copy.go
```diff
@@ -0,0 +1,353 @@
...
+			metachannel <- currProgress
+
+			m.Lock()
+			currentTag = sourceTags[i]
+			m.Unlock()
```
thedadams, line 296: Talked "offline" about passing this with the channel.

pkg/server/registry/apigroups/acorn/images/copy.go
```diff
@@ -0,0 +1,353 @@
...
+	if err != nil {
+		return err
+	}
+
+	sourceDigest, err := sourceIndex.Digest()
+	if err != nil {
+		return err
+	}
+
+	if destDigest.String() != sourceDigest.String() {
```
StrongMonkey, line 328: Should this be equal condition?

g-linville: The goal here is to avoid overwriting an image that already exists. If the digests match, it is okay to overwrite, but if they are different, that means that we will be overwriting an existing tag if we copy. So `!=` is what is intended here.

pkg/client/client.go
tylerslaton: If I do `acorn copy --all-tags <source> <dest>` and then hit cancel out (say through `ctrl`+`c`), does the copy continue happening in the background?

g-linville: Just tested this to see. The copy terminates properly and does not continue in the background

StrongMonkey (approved):
Don't have too much context but PR looks good to me. 

Great job on the work to show progress on copy!

pkg/server/registry/apigroups/acorn/images/copy.go
```diff
@@ -0,0 +1,339 @@
...
+	// metachannel is used to send another channel with updates for each tag to be copied
+	metachannel := make(chan updates)
+	// progress is the channel returned by this function and used to write websocket messages to the client
+	progress := make(chan ImageProgress)
+
+	go func() {
+		defer close(progress)
+		for c := range metachannel {
+			for update := range c.updateChan {
+				var errString = ""
```
thedadams, line 260: nit:
```suggestion
				var errString string
```

thedadams (approved):
Great job with this!

tylerslaton (approved):

PR #1885, opened 2023-07-05
Add info about the shared image registry

g-linville:


docs/docs/60-architecture/02-security-considerations.md
```diff
@@ -76,6 +76,14 @@ acorn:project:view
 acorn:project:view-logs
 ```
 
+### Shared Image Registry
+
+Default installations of Acorn Runtime will deploy an OCI registry into Kubernetes, which will be the default image storage for all projects.
+This registry is not secured with credentials.
+Acorn's API server performs basic checks to try to prevent users from trivially accessing images that belong to other projects, but this is not guaranteed.
+
+You can configure Acorn to use an external OCI registry, but the credentials are shared between all projects, so the security implications around image access are the same.
```
cjellick, line 85:
```suggestion
You can configure Acorn to use an external OCI registry, but the credentials are shared between all projects, so the security implications around image access are the same. Future releases will improve upon this by providing a framework for isolating registries.
```

cjellick (approved):
assuming you accept my edit, lgtm. and you dont have to wait for a second review

thedadams (approved):

PR #1886, opened 2023-07-06
Ensure that delete jobs run when projects are deleted (#1893)

thedadams:
There are a few situations where we need a buffer between a project being deleted and the corresponding namespace being deleted. One such scenario is ensuring that delete jobs run when a project is deleted.

This change adds a project instance, which is this buffer. Using the project instance, we can ensure that all apps are successfully cleaned up before the namespace is deleted. This should also give the user the ability to use the ignore-cleanup functionality when deleting a project.

### Checklist

Requires: https://github.com/acorn-io/baaah/pull/90
Issue: https://github.com/acorn-io/runtime/issues/1893


pkg/controller/defaults/testdata/memory/all-set-overwrite/expected.golden
```diff
@@ -40,16 +40,13 @@ status:
   appStatus: {}
   columns: {}
   conditions:
-    reason: Success
-    status: "True"
-    success: true
+  - error: true
```
thedadams, line 43: I need to fix this.

pkg/controller/appdefinition/testdata/computeclass/all-set/input.yaml
```diff
@@ -10,9 +10,6 @@ spec:
     "": sample-compute-class
 status:
   observedGeneration: 1
-  defaults:
```
thedadams, line 13: This was a duplicate entry.

pkg/controller/defaults/testdata/region/default/expected.golden
```diff
@@ -21,15 +21,13 @@ status:
...
+      *v1.ProjectInstance "app-namespace" not found'
+    reason: Error
+    status: "False"
     type: defaults
-  defaults:
-    memory:
-      "": 0
-      container-name: 0
-    region: local
+  defaults: {}
```
g-linville, line 30: It looks like this and several other tests went from expecting success to expecting failure. Was this intended?

pkg/server/registry/apigroups/acorn/projects/storage.go
```diff
@@ -1,37 +1,76 @@
...
 		WithValidateCreate(validator).
 		WithValidateUpdate(validator).
 		WithTableConverter(tables.ProjectConverter).
 		Build()
 }
+
+type projectCreater struct {
+	creater strategy.Creater
+	client  kclient.Client
+}
```
g-linville, line 41: nit: I know the underlying typo/misspelling comes from mink, but I think this is worth changing anyway

```suggestion
type projectCreator struct {
	creator strategy.Creater
	client  kclient.Client
}
```

(Would need to modify other parts of this file too.)

pkg/uninstall/uninstall.go
```diff
@@ -228,6 +238,28 @@ func Uninstall(ctx context.Context, opts *Options) error {
 		}
 	}
 
+	// Delete projects first. This will allow the acorn-controller to do all the cleanup it needs to do.
+	pterm.Info.Println("Deleting projects")
+	projectList := new(apiv1.ProjectList)
+	projects := make([]kclient.Object, 0, len(projectList.Items))
```
g-linville, line 244: nit: won't `len(projectList.Items)` always be 0 here?

pkg/controller/defaults/testdata/region/default/expected.golden
thedadams: I thought I fixed all of these, my apologies. I will update.

integration/helper/project.go
```diff
@@ -0,0 +1,64 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func TempProject(t *testing.T, client client.WithWatch) *v1.ProjectInstance {
+	t.Helper()
+	return tempCreateHelper(t, client)
+}
+
+func tempCreateHelper(t *testing.T, client client.WithWatch) *v1.ProjectInstance {
```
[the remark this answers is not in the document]
thedadams: There was another function that was using the helper, but I removed it (it wasn't necessary). I will simplify this.

pkg/uninstall/uninstall.go
thedadams: Yes, will update.

g-linville (approved):

pkg/uninstall/uninstall.go
```diff
@@ -80,6 +82,14 @@ func baseResources(ctx context.Context, c kclient.Client) (resources []kclient.O
 		}
 	}
 
+	controllerLease := &coordinationv1.Lease{}
+	if err := c.Get(ctx, kclient.ObjectKey{Namespace: "kube-system", Name: "acorn-controller"}, controllerLease); !apierror.IsNotFound(err) && err != nil {
+		return nil, err
+	}
+	if err == nil {
```
StrongMonkey, line 89: the err here is not the err in the if statement above?

pkg/uninstall/uninstall.go
```diff
@@ -80,6 +82,14 @@ func baseResources(ctx context.Context, c kclient.Client) (resources []kclient.O
 		}
 	}
 
+	controllerLease := &coordinationv1.Lease{}
+	if err := c.Get(ctx, kclient.ObjectKey{Namespace: "kube-system", Name: "acorn-controller"}, controllerLease); !apierror.IsNotFound(err) && err != nil {
```
StrongMonkey, line 86: Does the lease live in acorn-system or kube-system?

pkg/uninstall/uninstall.go
thedadams: You're right. Will fix.

pkg/uninstall/uninstall.go
thedadams: It lives in the `kube-system` namespace.

StrongMonkey (approved):

PR #1889, opened 2023-07-06
Switch to creating a single wildcard record per install

tylerslaton:
Prior to this PR, we would create a record in Acorn DNS for each Ingress that was created. This was heavily used for Apps where their Ingresses would each get a Acorn DNS entry. In doing this, some significant slow down could occur, mainly when publishing builders.

With this PR, we switch this process to creating a single wildcard certificate. This is done by introducing a few things.
1. A `acorn-ingress` Ingress object (and accompanying Service object) that is watched to determine what the wildcard record should point to.
2. Updates to the DNS Daemon to now watch the `acorn-ingress` Ingress and renew it when necessary.
3. Updates to the `ToRecordRequestAndHash` function now create wildcard records depending on the status of the `acorn-ingress` Ingress.
4. Updates to the DNS handler to only process the `acorn-ingress`, instead of every one that was created by Acorn.

Also went ahead and knocked out setting up GoMock for the DNS Client.

### Checklist


pkg/controller/config/dns.go
```diff
@@ -40,17 +42,65 @@ func (h *configHandler) Handle(req router.Request, resp router.Response) error {
...
+		return err
+	}
+
+	return nil
+}
+
+// purgeRecordsIfDisabling checks if we are transitioning AcorDNS from an enabled state to a disabled state and if so calls the
+// acorn DNS service to purge all records for the domain. It is expected that string variable returned by this function
+// will be set as the labels.AcornDNSState annotation on the acorn-dns secret
+func purgeRecordsIfDisabling(req router.Request, domain string, cfg *apiv1.Config, dnsSecret *corev1.Secret,
```
tylerslaton, line 60: Just moved this function up unchanged.

pkg/controller/config/dns.go
```diff
@@ -40,17 +42,65 @@ func (h *configHandler) Handle(req router.Request, resp router.Response) error {
...
+					}
+				}
+				return "", err
+			}
+		}
+	}
+	return state, nil
+}
+
+func (h *configHandler) handleSecret(req router.Request, resp router.Response, cfg *apiv1.Config) (string, error) {
```
tylerslaton, line 92: Broke this out into its own function to reduce the complexity burden on the `Handle` function.

pkg/controller/config/dns.go
```diff
@@ -97,50 +147,85 @@ func (h *configHandler) Handle(req router.Request, resp router.Response) error {
...
+				Port: 80,
+			}},
+			Selector: map[string]string{
+				"app": system.Namespace,
+			},
+		},
+	}
+
+	pt := netv1.PathTypeImplementationSpecific
+	ingress := &netv1.Ingress{
```
tylerslaton, line 192: This is based on the Ingress we create to test out Ingress for clusters during Check.

https://github.com/acorn-io/runtime/blob/a2b288d13600da77ea8a3c4fd9d087bd038f91c2/pkg/install/check.go#L320-L368

pkg/mocks/dns/mock.go
tylerslaton, line 1: This is generated by `gomocks`

pkg/system/constants.go
```diff
@@ -8,6 +8,8 @@ const (
 	LEAccountSecretName  = "acorn-le-account"
 	DefaultUserNamespace = "acorn"
 	DNSSecretName        = "acorn-dns"
+	IngressName          = "acorn-ingress"
```
cjellick, line 11: maybe we can be a bit more explicit about the name of this and its purpose? Thinking `AcornDNSTestIngres`...a bit longer/wordier, but clearer in its intent?

pkg/controller/config/dns.go
cjellick: Purely from an ease-of-review standpoint, i'd personally shy away from just moving a function like this. just throw throws the diff off.

pkg/controller/ingress/dns.go
```diff
@@ -39,6 +39,11 @@ func NewDNSHandler() router.Handler {
 
 // Handle calls the AcornDNS service to create records for ingresses if the acorn-dns feature is enabled
```
cjellick, line 40: can you update the comment to match the new functionality?

pkg/controller/ingress/dns_test.go
```diff
@@ -4,26 +4,28 @@ import (
...
+// field is set (required to create an FQDN).
 //
 // expected.yaml has the same ingress that is in input.yaml, but with the acorn.io/dns-hash annotation set, which indicates
 // the FQDN was successfully created in the DNS service.
 //
 // The client to the DNS service is mocked out.
 func TestBasicDNS(t *testing.T) {
 	h := &handler{
-		dnsClient: &mockClient{},
+		dnsClient: mockDNSClient(t),
```
cjellick, line 28: thanks for doing a proper mock!

pkg/system/constants.go
tylerslaton: Sure! We can do something like that.

pkg/controller/config/dns.go
```diff
@@ -97,50 +147,83 @@ func (h *configHandler) Handle(req router.Request, resp router.Response) error {
...
-		state = *cfg.AcornDNS
-	} else {
-		// The config object includes the acorn domain in the list of ClusterDomains if we are in an enabled state.
-		if slices.Contains(cfg.ClusterDomains, domain) {
-			state = "enabled"
-		} else {
-			state = "disabled"
-		}
+func (h *configHandler) handleIngress(req router.Request, resp router.Response, domain string, acornDNS *string, ingressClassName *string) error {
+	if acornDNS == nil || *acornDNS == "disabled" {
```
cjellick, line 164: This presents a sublte issue...if i run acorn locally, acornDNS will equal `"auto"`, so we will get past this and create the ingress and service.

But in this case, acorn will be configured to use the localhost dns entry - `local.oss-acorn.io`. That will be the domain that is set in config.ClusterDomains.

Later, in ingress/dns.go, the handler will ultimately do the right thing and NOT create the dns entries because the check it does is `if slices.Contains(cfg.ClusterDomains, domain) {` ([see here](https://github.com/acorn-io/runtime/blob/main/pkg/controller/ingress/dns.go#L65)). So, that's good - my main worry was that we were going to create dns entries when the cluster is actually configured to use local. The secondary concern is that in this case we are creating an ingress and service when we don't need to.

So, it would better to decide whether to proceed or not with the same check: `if slices.Contains(cfg.ClusterDomains, domain`, but here's where it gets tricky: the value of cfg.ClusterDomains is populated by logic that is dependent on the acornDNS secret just created - https://github.com/acorn-io/runtime/blob/main/pkg/config/config.go#L150-L176. So, the first time through, when we _actually_ create the dns secret, the config we have may be stale. And even if we got the config here again, getting it uses a cache to retrieve the secret, so it might not be caught up with the secret we just created.

I'm still thinking about how to address this, just wanted to get the problem out there.

pkg/controller/config/dns.go
```diff
@@ -97,50 +147,83 @@ func (h *configHandler) Handle(req router.Request, resp router.Response) error {
...
+			Name: system.IngressName,
+			Labels: map[string]string{
+				labels.AcornManaged: "true",
+			},
+		},
+		Spec: netv1.IngressSpec{
+			IngressClassName: ingressClassName,
+			Rules: []netv1.IngressRule{
+				{
+					Host: strings.TrimPrefix(domain, "."), // Needed since domain starts with a "."
```
cjellick, line 203: this sets the host to something like `x96fr6.oss-acorn.io`. Can we do a sub domain of that? like `dns-test.x96fr6.oss-acorn.io`. That would let us potentially dig that that domain and prove its resolving properly.

pkg/controller/config/dns.go
cjellick: here's an idea: make the handleIngress a handler for the acornDNS secret. So, any time that secret changes, it would trigger the handler. Then, get the config and it should be guaranteed to be "up to date" WRT the acornDNS secret because it otherwise would not have triggered the handler. Then you'd be safe to look at cfg.ClusterDomains

pkg/controller/config/dns.go
```diff
@@ -97,50 +147,83 @@ func (h *configHandler) Handle(req router.Request, resp router.Response) error {
...
+	var err error
+	if ingressClassName == nil {
+		ingressClassName, err = publish.IngressClassNameIfNoDefault(req.Ctx, req.Client)
+		if err != nil {
+			return err
 		}
 	}
-	return state, nil
+
+	svc := &corev1.Service{
```
cjellick, line 176: @thedadams or @njhale - here we create a service and ingress that isnt actually backed by any pods. We just want to get the ingress up so that we can inspect its status.loadbalancer field to figure out what this cluster's external IP or hostname is.

Is there a better way to create an "empty" ingress + service? A little bit beyond my technical knowledge

pkg/controller/config/dns.go
cjellick: Oh I see, you did this because later in ToRecordRequestsAndHash, if the host matches the clusterDomain exactly, you make the record be `*`.

...maybe we should change the impl of ToRecordRequestsAndHash...more on that down there in that function

pkg/dns/helpers.go
```diff
@@ -33,7 +33,9 @@ func ToRecordRequestsAndHash(domain string, ingress *v1.Ingress) ([]RecordReques
 
 	var hosts []string
 	for _, rule := range ingress.Spec.Rules {
-		if strings.HasSuffix(rule.Host, domain) {
+		if rule.Host == strings.TrimPrefix(domain, ".") {
```
cjellick, line 36: What if we simplified this so that the only thing this could create was a wild card record...so basically hosts would only ever be of size 1 and the check would be like
```
f strings.HasSuffix(rule.Host, domain) {
			hosts = append(hosts, "*")
			break
		}
```

cjellick (changes requested):
review finished. looks mostly good. just a few areas to reconsider

pkg/dns/helpers.go
tylerslaton: Makes sense to me.

pkg/controller/config/dns.go
tylerslaton: Ah, yeah I like that better. Good call.

pkg/controller/config/dns.go
tylerslaton: I like it, I'll get that in.

tylerslaton:
@cjellick I believe I've addressed all of your comments. Would you mind giving it another look?

pkg/controller/ingress/dns.go
```diff
@@ -37,8 +37,16 @@ func NewDNSHandler() router.Handler {
 	return s
 }
 
-// Handle calls the AcornDNS service to create records for ingresses if the acorn-dns feature is enabled
+// Handle calls the AcornDNS service to create a wild card record for based on
```
g-linville, line 40: nit:

```suggestion
// Handle calls the AcornDNS service to create a wild card record based on
```

pkg/dns/daemon.go
```diff
@@ -61,71 +62,74 @@ func (d *Daemon) internal(ctx context.Context) (bool, error) {
...
+		logrus.Errorf("Failed to sync ingress: %v", err)
+		return false, nil
+	}
+
+	logrus.Infof("Renewed and synced AcornDNS!")
+
+	return true, nil
+}
+
+func (d *Daemon) getDNSSecret(ctx context.Context) (secret *corev1.Secret, domain string, token string, err error) {
```
g-linville, line 83: Typically, named return values are used for very short functions, and in the function you can just set their values and then `return`. In all the return statements in this function, you are currently listing out all four variables. I would recommend either removing the names from the return values, or just setting the variables to what they need to be and `return`ing.

g-linville (approved):

pkg/system/constants.go
```diff
@@ -8,6 +8,8 @@ const (
 	LEAccountSecretName  = "acorn-le-account"
 	DefaultUserNamespace = "acorn"
 	DNSSecretName        = "acorn-dns"
+	IngressName          = "acorn-dns-test-ingress"
```
cjellick, line 11: was thinking the variable name could be `DNSIngressName` and `DNSServiceName`, just to eliminate ambiguity.

pkg/controller/routes.go
```diff
@@ -111,6 +111,7 @@ func routes(router *router.Router, cfg *rest.Config, registryTransport http.Roun
 	router.Type(&corev1.Pod{}).Selector(managedSelector).HandlerFunc(jobsHandler.SaveJobOutput)
 	router.Type(&netv1.Ingress{}).Selector(managedSelector).Namespace(system.ImagesNamespace).HandlerFunc(gc.GCOrphans)
 	router.Type(&netv1.Ingress{}).Selector(managedSelector).Middleware(ingress.RequireLBs).Handler(ingress.NewDNSHandler())
+	router.Type(&corev1.Secret{}).Selector(managedSelector).HandlerFunc(secrets.HandleDNSSecret)
```
cjellick, line 114: instead of watching all secrets, maybe we should use the Namesapce() route function to limit it.

cjellick: same with the `NewDNSHandler`route. might as well limit that to the the system namesapce too?

pkg/controller/routes.go
```diff
@@ -110,7 +110,8 @@ func routes(router *router.Router, cfg *rest.Config, registryTransport http.Roun
 	router.Type(&corev1.Pod{}).Selector(managedSelector).HandlerFunc(jobs.JobPodOrphanCleanup)
 	router.Type(&corev1.Pod{}).Selector(managedSelector).HandlerFunc(jobsHandler.SaveJobOutput)
 	router.Type(&netv1.Ingress{}).Selector(managedSelector).Namespace(system.ImagesNamespace).HandlerFunc(gc.GCOrphans)
-	router.Type(&netv1.Ingress{}).Selector(managedSelector).Middleware(ingress.RequireLBs).Handler(ingress.NewDNSHandler())
+	router.Type(&netv1.Ingress{}).Selector(managedSelector).Namespace(system.Namespace).Middleware(ingress.RequireLBs).Handler(ingress.NewDNSHandler())
```
thedadams, line 113: If we only care about acting something with a specific name, you can use `.Name("acorn-dns-name")` in addition to the `Namespace` method. This will means you can get rid of the name check in the method.

pkg/controller/secrets/dns.go
```diff
@@ -0,0 +1,105 @@
...
+	if slices.Contains(cfg.ClusterDomains, domain) {
+		state = "enabled"
+	}
+
+	if strings.EqualFold(state, "disabled") {
+		return nil
+	}
+
+	ingressClassName := cfg.IngressClassName
+	if ingressClassName == nil || *ingressClassName == "" {
```
thedadams, line 45: Is this always non-nil when getting the config?

pkg/controller/secrets/dns.go
```diff
@@ -0,0 +1,105 @@
...
+		},
+	}
+
+	pt := netv1.PathTypeImplementationSpecific
+	ingress := &netv1.Ingress{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: system.DNSIngressName,
+			Labels: map[string]string{
+				labels.AcornManaged: "true",
+			},
```
thedadams, line 73: Should we add this label to the `Service` as well?

pkg/dns/daemon.go
```diff
@@ -61,71 +62,73 @@ func (d *Daemon) internal(ctx context.Context) (bool, error) {
...
 		}
+		return nil, fmt.Errorf("problem getting DNS secret %v/%v: %v", system.Namespace, system.DNSSecretName, err)
+	}
+	return dnsSecret, nil
+}
+
+func (d *Daemon) syncIngress(ctx context.Context, domain, token, acornDNSEndpoint string, secret *corev1.Secret) error {
+	var ingress netv1.Ingress
+	err := d.client.Get(ctx, router.Key(system.Namespace, system.DNSIngressName), &ingress)
+	if err != nil {
```
cjellick, line 103: You should probably check for a notFound error and treat it different by not returning an error.

You can validly get a NotFound if you are running locally with cfg.AcornDNS == auto. This should result in no ingress being created.

cjellick (changes requested):
few more thoughts

pkg/controller/secrets/dns.go
tylerslaton: Yes, it is. You are correct and I'll make a change to make the logic less verbose.

pkg/dns/daemon.go
tylerslaton: Gotcha. This has been updated.

pkg/controller/secrets/dns.go
```diff
@@ -0,0 +1,105 @@
...
+		return err
+	}
+
+	// The config object includes the acorn domain in the list of ClusterDomains if we are in an enabled state.
+	state := *cfg.AcornDNS
+	if slices.Contains(cfg.ClusterDomains, domain) {
+		state = "enabled"
+	}
+
+	if strings.EqualFold(state, "disabled") {
```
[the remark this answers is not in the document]
tylerslaton: I believe this logic should now be handled appropriately.

pkg/controller/routes.go
tylerslaton: Updated!

cjellick: oh good. ...I thought there should be a Name() but didnt see an example in here.

pkg/dns/daemon.go
```diff
@@ -61,71 +62,76 @@ func (d *Daemon) internal(ctx context.Context) (bool, error) {
...
-		for _, r := range rrs {
-			ingressMap[FQDNTypePair{FQDN: r.Name, Type: string(r.Type)}] = ingress
+func (d *Daemon) syncIngress(ctx context.Context, domain, token, acornDNSEndpoint string, secret *corev1.Secret) error {
+	var ingress netv1.Ingress
+	err := d.client.Get(ctx, router.Key(system.Namespace, system.DNSIngressName), &ingress)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
 		}
+		return fmt.Errorf("failed to get %v for DNS renewal: %v", system.DNSIngressName, err)
```
cjellick, line 107: nit: should start using `%w` for error wrapping as a best practice

pkg/controller/secrets/dns.go
```diff
@@ -0,0 +1,99 @@
...
+	if domain == "" {
+		return fmt.Errorf("acorn dns secret is misconfigured, missing domain")
+	}
+
+	cfg, err := config.Get(req.Ctx, req.Client)
+	if err != nil {
+		return err
+	}
+
+	if strings.EqualFold(*cfg.AcornDNS, "disabled") || !slices.Contains(cfg.ClusterDomains, domain) {
```
cjellick, line 31: pretty sure the clusterDomains check covers the case where the feature is disabled, but theres no real harm in checking both

tylerslaton: Talked offline and updated.

cjellick (approved):

thedadams (approved):

Issue #1892, opened 2023-07-07
Change Acorn volumes back to using a public name strategy


g-linville:
We used to use a public name strategy for Acorn volumes in the apiserver, but we switched away from it due to bugs surrounding rebinding existing volumes.

After talking through it again, we want to switch back to using public name, with the following constraints:

1. Once a public name is set on a volume, it must never change, even when rebinding.
   - e.g. if I create an app called `app1` that has a volume called `app1.myvol`, and I delete `app1` and rebind `app1.myvol` to a new app called `app2`, the volume's name is still `app1.myvol`; it does not change.
2. When creating a new volume, the controller needs to check to make sure that the public name of the new volume will not match the public name of an existing volume in order to avoid duplicate names.
   - Continuing the previous example, if I were to try to deploy `app1` again, it will try to create another volume called `app1.myvol`, but the controller should block this, since a volume with public name `app1.myvol` already exists.

sangee2004:
Tested with acorn version - `v0.8.0-alpha7-133-ge68fb042+e68fb042`

1. Deploy app - `mytest1` that uses volume using following Acornfile 
```
args: {
fileName: "hello.html"
fileContent: "content"
}
containers: {
    frontend: {
        build: "."
        cmd: ["/root/start.sh", args.fileContent,args.fileName]
        dirs: {
            "/scratch": "volume://scratch-data"
        }
        sidecars: {
            sidecar: {
            ports: publish: "80/http"
            image: "nginx"
            dirs: {
                "/usr/share/nginx/html/test1": "volume://scratch-data"
            }
          }
        }
}
}
volumes:
{
"scratch-data": {}
}
```

2. once app is deployed , volume gets created with public name
```
acorn volumes
NAME                   BOUND-VOLUME   CAPACITY   VOLUME-CLASS   STATUS    ACCESS-MODES   CREATED
mytest1.scratch-data   scratch-data   10G        local-path     bound     RWO            11s ago
```

3. Delete app - `test1` (without deleting volumes) - `acorn rm test1`

4. Deploy another app that binds to existing volume- -`acorn run -n mytestbindvol -v mytest1.scratch-data:my-data` using following Acornfile
```
containers: {
mynginx: {
ports: publish: "80/http"
image: "nginx"
dirs: {
"/usr/share/nginx/html/test1": "volume://my-data"
}
}
}
```

4. Once app gets deployed successfully, the name of the existing volume continues to remain the same as expected:

```
acorn volumes
[...]

PR #1903, opened 2023-07-10
Use a public name strategy for volumes (#1892)

g-linville:
for #1892

This changes the apiserver strategy back to public names for volumes. Previously there was an issue where the public name would change when a volume was bound, which led to syncing issues in Manager. When I made the public name static, then it was possible to have duplicate volume names.

With this PR, the public names do not change when volumes get bound, and the controller performs the proper checks to make sure it is not going to create a volume with a duplicate public name.

### Checklist

### DIFF
--- a/pkg/cli/testdata/all/all_test.txt
+++ b/pkg/cli/testdata/all/all_test.txt
@@ -8,8 +8,8 @@ NAME              APP       IMAGE     STATE     RESTARTCOUNT   CREATED    MESSAG
 found.container                                 0              292y ago   
 
 VOLUMES:
-NAME        APP-NAME   BOUND-VOLUME   CAPACITY   VOLUME-CLASS   STATUS    ACCESS-MODES   CREATED
-found.vol   found      vol            <nil>                                              292y ago
+NAME        BOUND-VOLUME   CAPACITY   VOLUME-CLASS   STATUS    ACCESS-MODES   CREATED
+found.vol   vol            <nil>                                              292y ago
 
 SECRETS:
 NAME           TYPE      KEYS      CREATED
--- a/pkg/controller/appdefinition/testdata/volumes/named-bound/expected.golden
+++ b/pkg/controller/appdefinition/testdata/volumes/named-bound/expected.golden
@@ -101,7 +101,7 @@ metadata:
     acorn.io/app-name: app-name
     acorn.io/app-namespace: app-namespace
     acorn.io/managed: "true"
-    acorn.io/public-name: app-name.foo
+    acorn.io/public-name: other-app.foo
     acorn.io/volume-class: custom-class
     acorn.io/volume-name: foo
   name: foo-bind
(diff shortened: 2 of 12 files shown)


pkg/controller/appdefinition/testdata/volumes/ephemeral-bound/expected.golden
```diff
@@ -99,7 +99,7 @@ metadata:
     acorn.io/app-name: app-name
     acorn.io/app-namespace: app-namespace
     acorn.io/managed: "true"
-    acorn.io/public-name: app-name.foo
+    acorn.io/public-name: other-app.foo
```
g-linville, line 102: This expected value changed because public names are now static and will not change when a new app binds an old volume.

pkg/controller/appdefinition/testdata/volumes/named-bound/expected.golden
```diff
@@ -101,7 +101,7 @@ metadata:
     acorn.io/app-name: app-name
     acorn.io/app-namespace: app-namespace
     acorn.io/managed: "true"
-    acorn.io/public-name: app-name.foo
+    acorn.io/public-name: other-app.foo
```
g-linville, line 104: This expected value changed because public names are now static and will not change when a new app binds an old volume.

pkg/cli/testdata/all/all_test.txt
tylerslaton, line 1: Side thought - we should wire this up to the golden logic.

tylerslaton (approved):

pkg/controller/appdefinition/volume.go
```diff
@@ -201,27 +207,77 @@ func toPVCs(req router.Request, appInstance *v1.AppInstance) (result []kclient.O
 			}
 		}
 
+		// Ensure that no other PersistentVolume exists with the same public name
+		selector := klabels.SelectorFromSet(map[string]string{
+			labels.AcornPublicName: pvc.Labels[labels.AcornPublicName],
+			// Make sure we only list volumes that are part of the same Acorn project
+			labels.AcornAppNamespace: appInstance.Namespace,
+			labels.AcornManaged:      "true",
```
thedadams, line 215: Should this be left out just in case? I think have the namespace makes sense, especially if we use the same selector to list volumes in the API server when GET-ing.

thedadams (approved):

pkg/controller/appdefinition/volume.go
g-linville: I'm a bit confused, are you asking if I should remove the `acorn.io/managed` selector?

thedadams: Yes. My thought is, "What if the user tries to create a volume with the public name and app namespace labels?"

If a user did this, then you wouldn't see that PV here, but it may cause issues if we try to list in other places.

If you think I am being too cautious, then I am fine leaving this as is.

thedadams (approved):

Issue #1897, opened 2023-07-07
Remove `-A` from all commands except `acorn ps`


g-linville:
In this issue we want to accomplish a few things:

1. Remove the `-A` flag from all commands except for `acorn ps`
   - Might want to rename it to something else, like maybe `--all-projects`
1. Switch from `acorn apps` to `acorn ps` as the main command name; leave `acorn apps` as a valid alias
1. Make `-A` (or whatever it is renamed to) only show apps in all projects in the user's "current server" for `acorn ps`
   - For example, if I am logged into `hub.stg.acornlabs.com`, I should only see the projects in that instance of Manager, and not in any others or in my current k8s context
   - There are a couple ways to accomplish this: we can determine which server to use based on the user's default project (`acorn project use`), or we could have some argument to `acorn ps` to specify the server. Try things out and see what works the best.

sangee2004:
Tested with acorn version - `v0.8.0-alpha7-98-gbb130866+bb130866`

1. `-A` option is removed from all commands expect `ps`
2. `acorn help` lists `ps`. Able to use `app` and `ps` to list all apps.
3. `-A` option is supported for `ps`
4. `acorn apps -A` list apps from all projects from the current server. 

```
 acorn project
NAME                                        DEFAULT   REGIONS
acorn                                       *         local
hub.dev.acornlabs.com/sangee2004/mytest1              acorn-us-east-2
hub.stg.acornlabs.com/sangee2004/test1                acorn-us-east-2
hub.stg.acornlabs.com/sangee2004/testnew1             acorn-us-east-2
test1                                                 local
```
1. When in the context of `hub.stg.acornlabs.com/sangee2004/test1` or `hub.stg.acornlabs.com/sangee2004/testnew1` , using `acorn apps -A` lists all apps from both  `hub.stg.acornlabs.com/sangee2004/test1` and `hub.stg.acornlabs.com/sangee2004/testnew1` 
2. When in the context of `hub.dev.acornlabs.com/sangee2004/mytest1` , using `acorn apps -A` lists all apps from only `hub.dev.acornlabs.com/sangee2004/mytest1`
3. When in the context of `acorn` or `test1` , using `acorn apps -A` lists all apps from both `acorn` and `test1`


All the above scenarios work as expected.

sangee2004:
There is an issue with `-j` not working as expected when used with "-A" option.

In this case `-j` option is ignored.

```
acorn projects
NAME                                        DEFAULT   REGIONS
[...]

PR #1907, opened 2023-07-10
Remove -A from almost all commands (#1897)

g-linville:
for #1897

`-A` is now only supported for `acorn ps`. Also, `ps` is the main name for that command now, instead of `apps` (which is still a valid alias).

Additionally, the behavior of `-A` will now only include other projects in the same Acorn "instance" (one installation of Runtime, or one instance of Manager).

### Checklist

### DIFF
--- a/docs/sidebars.js
+++ b/docs/sidebars.js
@@ -99,7 +99,6 @@ const sidebars = {
           "items": [
             "reference/command-line/acorn",
             "reference/command-line/acorn_all",
-            "reference/command-line/acorn_app",
             "reference/command-line/acorn_build",
             "reference/command-line/acorn_check",
             "reference/command-line/acorn_container",
@@ -120,6 +119,7 @@ const sidebars = {
             "reference/command-line/acorn_project_create",
             "reference/command-line/acorn_project_rm",
             "reference/command-line/acorn_project_use",
+            "reference/command-line/acorn_ps",
             "reference/command-line/acorn_pull",
             "reference/command-line/acorn_push",
             "reference/command-line/acorn_offerings",
--- a/pkg/cli/ps.go
+++ b/pkg/cli/ps.go
@@ -3,32 +3,42 @@ package cli
 import (
 	cli "github.com/acorn-io/runtime/pkg/cli/builder"
 	"github.com/acorn-io/runtime/pkg/cli/builder/table"
+	"github.com/acorn-io/runtime/pkg/client"
 	"github.com/acorn-io/runtime/pkg/tables"
 	"github.com/spf13/cobra"
 	"k8s.io/utils/strings/slices"
 )
 
-func NewApp(c CommandContext) *cobra.Command {
-	return cli.Command(&App{client: c.ClientFactory}, cobra.Command{
-		Use:     "app [flags] [APP_NAME...]",
-		Aliases: []string{"apps", "a", "ps"},
+func NewPs(c CommandContext) *cobra.Command {
+	return cli.Command(&Ps{client: c.ClientFactory}, cobra.Command{
+		Use:     "ps [flags] [APP_NAME...]",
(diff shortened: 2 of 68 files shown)


pkg/project/client.go
```diff
@@ -150,8 +150,29 @@ func ParseProject(project string, kubeconfigs map[string]string) (serverOrKubeco
...
+			}
+		} else {
+			for _, p := range projects {
+				if !strings.Contains(p, "/") {
+					result = append(result, p)
+				}
+			}
+		}
+
+		return result, err
```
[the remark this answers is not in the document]
StrongMonkey: yeah, it would nice to embedded that into list function

StrongMonkey (approved):

pkg/project/operations.go
```diff
@@ -165,33 +153,26 @@ func List(ctx context.Context, opts Options) (projects []string, warnings map[st
...
+	var result listResult
 	warnings = map[string]error{}
 
-	listLocalKubeconfig(ctx, &wg, result, opts)
-	if !onlyListLocalKubeconfig {
-		listAcornServer(ctx, &wg, creds, cfg, result)
+	if onlyListLocalKubeconfig || !managerHostExists {
+		result = listLocalKubeconfig(ctx, opts)
+	} else {
+		result = listAcornServer(ctx, managerHost, creds)
```
thedadams, line 169: This `List` function is also used when the user runs `acorn projects`. Should we still return all projects in that case? If so, I don't think this will work in this case.

g-linville: Oh, that is a good point that I did not realize before. Should I just add another parameter to this function which specifies whether to perform this one-server filtering (and restore the old functionality)?

g-linville:
This is annoying, I run `make gen-docs` locally (and outside of JetBrains and VSCode, in a normal terminal) and it shows no changes, but the CI is complaining.

pkg/project/operations.go
thedadams: Yeah, I think that make sense to me.

docs/sidebars.js
```diff
@@ -99,7 +99,6 @@ const sidebars = {
           "items": [
             "reference/command-line/acorn",
             "reference/command-line/acorn_all",
-            "reference/command-line/acorn_app",
```
iwilltry42, line 102: I wonder if we should have redirects for renames :thinking: 
Potentially some blog post is linking to it (though I hope that none links to generated content).

iwilltry42 (approved):
Two questions, but nothing blocking.
LGTM

pkg/cli/ps.go
```diff
@@ -3,32 +3,42 @@ package cli
...
 }
 
-type App struct {
-	All    bool   `usage:"Include stopped apps" short:"a"`
-	Quiet  bool   `usage:"Output only names" short:"q"`
-	Output string `usage:"Output format (json, yaml, {{gotemplate}})" short:"o"`
-	client ClientFactory
+type Ps struct {
+	All         bool   `usage:"Include stopped apps" short:"a"`
+	AllProjects bool   `usage:"Include all projects" short:"A"`
```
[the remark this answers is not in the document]
g-linville: Curious to know how @ibuildthecloud would want this to be phrased. I agree that this help message should be clearer.

docs/sidebars.js
g-linville: Good call, I added a redirect (and tested it to make sure it works). I modified the docs generator a little bit so that it won't clobber the `.mdx` file I created.

thedadams (approved):

pkg/cli/ps.go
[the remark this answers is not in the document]
g-linville: Resolved this today during standup.

Issue #1906, opened 2023-07-10
Always return event details


njhale:
## Description

Requests to list and watch events should never strip event details.

## Background

The details field of a runtime event allows for context specific information to be added. For example, VCS info is included for AppImagePull events.

By default, the runtime API elides the field on list and watch unless otherwise opted-in by the request. The idea was to reduce the overhead when working with lots events containing potentially large `details`.

## Motivation

This was likely a premature optimization since, in practice, the details field has been used to store small payloads, similar in size to their respective objects.

njhale:
Opened #1913 to address something I missed in #1908

sangee2004:
Tested with acorn version - `v0.8.0-alpha7-98-gbb130866+bb130866 `
1. `acorn events -o json` always returns `details` info.
2. There is no support for "--details" flag anymore.

PR #1908, opened 2023-07-10
 Always return event details (#1906)

njhale:
Modify the runtime events API and acorn cli s.t. the details field is always returned intact, and is no longer elided from list|watch responses.

Vanquishes #1906


g-linville (approved):

tylerslaton (approved):

vincent99 (approved):
+1 for vanquishing

PR #1913, opened 2023-07-11
Drop details flag from acorn events subcommand (#1906)

njhale:
Remove the non-functional `--details` flag that was left exposed by #1908 on
the acorn events subcommand.

Completes #1906


tylerslaton (approved):
Approving with the assumption CI becomes green.

g-linville (approved):

PR #1914, opened 2023-07-11
Error early if manager token has already been used

tylerslaton:
This updates the `acorn` binary to fail early when running `acorn login` on an already used token.

### Checklist


pkg/manager/manager.go
```diff
@@ -107,6 +108,8 @@ func Login(ctx context.Context, password, address string) (user string, pass str
 			} else {
 				logrus.Debugf("tokenRequest.Status.Token is empty")
 			}
+		} else if passwordIsSpecified && strings.Contains(err.Error(), "404") {
```
njhale, line 111: nit: we could also return a custom `notFound` error from `httpGet` and use `errors.Is(...)` to be a bit more idiomatic.

pkg/manager/manager.go
```diff
@@ -107,6 +108,8 @@ func Login(ctx context.Context, password, address string) (user string, pass str
...
+			return "", "", fmt.Errorf("specified token does not exist, please create one and try again")
```
njhale, line 112: For cli users, does "create one and try again" really mean "get a new token from the manager UI"? Is there another way to do that right now?

njhale (commented):
Looks good, just one question.

pkg/manager/manager.go
tylerslaton: Yeah, that essentially means go to the UI and get one or run `acorn login <manager env>` (since there is a flow for that)

tylerslaton: What do you think about this?

```
specified token does not exist; please create a token via the web UI or omit the --password flag to request one via your browser
```

iwilltry42 (approved):

pkg/manager/manager.go
njhale: lgtm!

njhale (approved):

PR #1925, opened 2023-07-12
Add AutoUpgrade status field to replace API Server validation

tylerslaton:
As talked about during stand-up towards the end, this updates the AutoUpgrade implication logic to write to the App's status instead of failing on the API Server. This allows for there for be no friction from the API server when `autoupgrade` is not set but is implied.

### Checklist

### DIFF
--- a/pkg/cli/dev.go
+++ b/pkg/cli/dev.go
@@ -3,9 +3,11 @@ package cli
 import (
 	"io"
 
+	"github.com/acorn-io/runtime/pkg/autoupgrade"
 	cli "github.com/acorn-io/runtime/pkg/cli/builder"
 	"github.com/acorn-io/runtime/pkg/dev"
 	"github.com/acorn-io/runtime/pkg/imagesource"
+	"github.com/acorn-io/z"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/controller/routes.go
```diff
@@ -64,6 +64,7 @@ func routes(router *router.Router, cfg *rest.Config, registryTransport http.Roun
 	appRouter.HandlerFunc(appdefinition.PullAppImage(registryTransport, recorder))
 	appRouter.HandlerFunc(images.CreateImages)
 	appRouter.HandlerFunc(appdefinition.ParseAppImage)
+	appRouter.HandlerFunc(appdefinition.SetImpliedAutoUpgrade)
```
g-linville, line 67: I feel like it is a little overkill to have an entire handler for this. I think it should be included in one of the other ones, like `appstatus.PrepareStatus`, or, farther down, `appstatus.SetStatus`. I'm not 100% on this though and am curious to know what others think.

tylerslaton: You bring up a good point. Let me look around for somewhere this may feel more idiomatic.

tylerslaton: Gave it some thought, what if we added an `ObservedAutoUpgrade` field and it gets set in `UpdateObservedFields`?

pkg/cli/dev.go
```diff
@@ -64,6 +66,11 @@ func (s *Dev) Run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	// Ensure opts.AutoUpgrade is set if is implied by opts.AutoUpgradeInterval or imageSource.Image's pattern.
+	if opts.AutoUpgrade == nil || !*opts.AutoUpgrade {
+		opts.AutoUpgrade = z.P(autoupgrade.ImpliedAutoUpgrade(opts.AutoUpgradeInterval, imageSource.Image))
+	}
+
```
thedadams, line 73: If I understand correctly, this is no longer needed because the UI should be looking at the `status` field. Is that right?

If that is correct, can we remove the other place where this was done previously?

tylerslaton: Yeah I was on the fence before you mentioned it about this. After some more though (and the current implementation) I think what you're saying makes sense.

thedadams (approved):

PR #1927, opened 2023-07-13
Add server-side defaults for event fields

njhale:
Default events fields on create and update s.t.: 
- `actor` is to set to the request user (subject)
- `observed` is set to the current time

This will make it easier for clients to record events.


tylerslaton (approved):
I like it, really clean implementation. 👍🏻

iwilltry42 (approved):

g-linville (approved):

PR #1930, opened 2023-07-14
Rename event source to "resource" 

njhale:
Change the "event source" nomenclature to "event resource" in order clarify the reference to on-cluster resources related to an event.


tylerslaton (commented):
Does this also need a PR in manager? Will it break once this goes in?

njhale:
> Does this also need a PR in manager? Will it break once this goes in?

@tylerslaton Yup you're right, it will break event recording and will require a follow up PR in manager. However, the break shouldn't be too impactful since event recording is already best-effort; i.e. recording errors don't block business logic.

tylerslaton (approved):

g-linville (approved):

pkg/server/registry/apigroups/acorn/events/strategy.go
```diff
@@ -187,7 +187,7 @@ type prefix string
 func (p prefix) matches(e apiv1.Event) bool {
 	return p.all() ||
 		strings.HasPrefix(e.Name, string(p)) ||
-		strings.HasPrefix(e.Source.String(), string(p))
+		(e.Resource != nil && strings.HasPrefix(e.Resource.String(), string(p)))
```
thedadams, line 190: Does this work?
```suggestion
		strings.HasPrefix(z.D(e.Resource).String(), string(p))
```

thedadams (approved):

pkg/server/registry/apigroups/acorn/events/strategy.go
njhale: This is actually the exact line that prompted me to add `z.D` (now `z.Dereference`).

Will update this while rebasing.

PR #1931, opened 2023-07-14
Stop waiting for credentials forever

thedadams:
There seems to be a situation where the builder will wait for credentials forever, blocking progress on other builds.

This change does two things:
1. Cancel the receiving of messages when the context is canceled.
2. Timeout if credentials haven't been received in 15 seconds.


g-linville (approved):

StrongMonkey (approved):
Wow good job figuring that out

PR #1932, opened 2023-07-14
Add top-level app field to events

njhale:
### Description

Add an optional app field to events to support referencing associated Acorns.

### DIFF
--- a/pkg/apis/internal.acorn.io/v1/event.go
+++ b/pkg/apis/internal.acorn.io/v1/event.go
@@ -36,6 +36,10 @@ type EventInstance struct {
 	// +optional
 	Actor string `json:"actor,omitempty"`
 
+	// AppName is the name of the app that the event is related to, if any.
+	// +optional
+	AppName string `json:"appName,omitempty"`
+
 	// Source identifies the object the Event is regarding.
 	Source EventSource `json:"source"`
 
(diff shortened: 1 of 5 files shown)


pkg/apis/internal.acorn.io/v1/event.go
```diff
@@ -21,6 +21,15 @@ type EventInstanceList struct {
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
 type EventInstance struct {
+	EventCommon `json:",inline"`
+
+	// App is the name of the app that the event is related to, if any.
+	// +optional
+	App string `json:"app,omitempty"`
```
njhale, line 28: This approach feels very kludgy. Is there an alternative that I missed?

My instinct is that `EventSource` -- to be renamed `EventResource` soon -- can indicate the app the event is related to, if any. If we make it into a list it can track other related objects too; e.g. App and ContainerReplica.

but maybe I'm missing something obvious

ibuildthecloud: The field should actually be called `AppName`.  I think the field is useful, overly generic data structures are generically useless.

njhale: Got it. Thanks for the guidance there.

g-linville (approved):

tylerslaton (approved):

PR #1934, opened 2023-07-14
Retry building up to 5 times if it fails

tylerslaton:
This helps to mask an issue where the first attempt to build would result in a "✗  ERROR:  websocket: bad handshake" message printed. Typically, subsequent attempts would succeed and this just codifies that so a user doesn't have to manually retry.

I am also adding in a VS Code launch configuration for building an Acornfile.

### Checklist

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -2,6 +2,14 @@
     "version": "0.2.0",
     "configurations": [
         {
+            "name": "Build Acornfile",
+            "type": "go",
+            "request": "launch",
+            "mode": "auto",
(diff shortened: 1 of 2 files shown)


StrongMonkey:
My idea is that we should have proper logic [here](https://github.com/tylerslaton/runtime/blob/7c10a05c8de6c42b49d1c3c849fbd1aa613c6183/pkg/client/builder.go#L54) to detect whether builder server is ready and like last fix donnie did to watch app/deployment and check whether deployment is ready

Issue #1472, opened 2023-04-12
Not able to bind secrets (that are not associated with app anymore) to another app.


sangee2004:
Acorn version -  `v0.6.0-143-g94d09a10+94d09a10`

Steps to reproduce the problem:
1. Deploy an app which has secrets.

```
args: password: ""

containers: app: {
  image: "nginx"
  env: {
    PASSWORD: "secret://user-password/password"
  }
  ports: publish: "80/http"
  files: {
      "/usr/share/nginx/html/secret.html": "secret://user-password/password"
  }
}

secrets: "user-password": {
  type: "opaque"
  data: password: "\(args.password)"
}
```

2. Once app is deployed successfully , delete this app without deleting secrets.

3. Deploy another app by binding the secrets (relating to the deleted app) . App deployment is in "pending" state with error - `[secrets: missing: [mytestsec]]` message
```
containers: {
  nginx1: {
    image: "nginx"
    ports: publish: "80/http"
    files: {
      "/usr/share/nginx/html/secret.html": "secret://user-password/password"
    }
}
}
```

```
 % acorn rm mytestsec                                                                  
Removed: mytestsec
% acorn run -n mytestsecbind1 -s mytestsec.user-password:user-password -f AcornfileBind
.....
STATUS: ENDPOINTS[] HEALTHY[0] UPTODATE[0] [secrets: missing: [mytestsec]]
STATUS: ENDPOINTS[http://nginx1-mytestsecbind1-bdbbb2f7.local.on-acorn.io => nginx1:80] HEALTHY[0] UPTODATE[0] [secrets: missing: [mytestsec]]
^C  ✗  ERROR:  terminating watch: context canceled

 % acorn secrets mytestsec.user-password
NAME                      TYPE      KEYS         CREATED
mytestsec.user-password   opaque    [password]   115s ago
sangeethahariharan@Sangeethas-MBP secrets % 
```

g-linville:
Fix just merged to main.

sangee2004:
Tested with `acorn version v0.8.0-alpha7-143-g86de32e8+86de32e8`

1. Deploy an app which has secrets.
2. Once app is deployed successfully , delete this app without deleting secrets.
3. Deploy another app by binding the existing secret (which was created as part of app deployment in step1).
App deployment succeeds and app get access to the existing secret as expected.

PR #1940, opened 2023-07-17
Fix secret binding to support public names (#1472)

g-linville:
for #1472 

Secret bindings were only looking for secrets by exact name and not by public name. This fixes that problem.

### Checklist

### DIFF
--- a/pkg/controller/secrets/testdata/binding/existing.yaml
+++ b/pkg/controller/secrets/testdata/binding/existing.yaml
@@ -1,0 +1,11 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: existing-secret-abcdef
+  namespace: app-namespace
+  labels:
+    acorn.io/public-name: old-app.secret-name
+type: secrets.acorn.io/opaque
+data:
+  # username: myusername
+  username: bXl1c2VybmFtZQ==
--- a/pkg/ref/expr.go
+++ b/pkg/ref/expr.go
@@ -59,6 +59,16 @@ func Lookup(ctx context.Context, req kclient.Client, out kclient.Object, namespa
 		}
 
 		if validSecrets == nil {
+			switch v := out.(type) {
+			case *corev1.Secret:
+				// Support binding existing secrets with "." in the name, i.e. my-old-app.secret-name
+				if err := r.getSecret(v, namespace, fmt.Sprintf("%s.%s", name, parts[i+1]), validSecrets); err == nil {
+					return nil
+				}
+			}
+		}
+
+		if validSecrets == nil {
 			app, err := r.getAcorn(namespace, name)
 			if apierrors.IsNotFound(err) {
 				svc := &v1.ServiceInstance{}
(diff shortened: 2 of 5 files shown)


tylerslaton (approved):

njhale (approved):

pkg/controller/secrets/testdata/binding/existing.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: existing-secret-abcdef
+  namespace: app-namespace
+  labels:
+    acorn.io/public-name: old-app.secret-name
+type: secrets.acorn.io/opaque
+data:
+  # username: myusername
```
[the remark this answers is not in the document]
g-linville: The best part is that Copilot did the base64 for me

g-linville:
Re-requested review because I made a slight change to support secrets with public names that have more than one `.`. This can happen with nested Acorns and services.

pkg/ref/expr.go
```diff
@@ -59,6 +59,14 @@ func Lookup(ctx context.Context, req kclient.Client, out kclient.Object, namespa
 		}
 
 		if validSecrets == nil {
+			switch v := out.(type) {
+			case *corev1.Secret:
+				// Support binding existing secrets with "." in the name, i.e. my-old-app.secret-name
+				if err := r.getSecret(v, namespace, strings.Join(parts, "."), validSecrets); err == nil {
```
StrongMonkey, line 65: Do we want to ignore err here?

g-linville: Yes, we do. This `Lookup` function has multiple different uses. This specific case I am adding is for secret public names only. If we get an error while looking for that, we can just ignore it and move on to the next case.

StrongMonkey: Oh got it. Do we want to just ignore the not found error, but in other case, let's say in case of a network error, just return it? It looks like in other cases we only ignore notFound error.

g-linville: Sure, I think that is a good call. I will go add that

StrongMonkey: Thanks, just doing it safely in case there is other type of error.

StrongMonkey (approved):

pkg/ref/expr.go
```diff
@@ -59,6 +59,13 @@ func Lookup(ctx context.Context, req kclient.Client, out kclient.Object, namespa
 		}
 
 		if validSecrets == nil {
+			if v, ok := out.(*corev1.Secret); ok {
+				// Support binding existing secrets with "." in the name, i.e. my-old-app.secret-name
+				if err := r.getSecret(v, namespace, strings.Join(parts, "."), validSecrets); err == nil || !apierrors.IsNotFound(err) {
```
thedadams, line 64:
```suggestion
				if err := r.getSecret(v, namespace, strings.Join(parts, "."), validSecrets); !apierrors.IsNotFound(err) {
```

pkg/ref/expr.go
```diff
@@ -146,7 +153,28 @@ func (r *resolver) getSecret(secret *corev1.Secret, namespace, name string, vali
...
+
+			if len(secretList.Items) == 1 {
+				secretList.Items[0].DeepCopyInto(secret)
+				return nil
+			}
+		}
+		return err
+	}
+
+	return nil
```
thedadams, line 177: nit:
```suggestion
	if err := r.req.Get(r.ctx, router.Key(namespace, name), secret); !apierrors.IsNotFound(err) {
		return err
	}

	// Try finding by public name
	secretList := &corev1.SecretList{}
	if err := r.req.List(r.ctx, secretList, &kclient.ListOptions{
		LabelSelector: klabels.SelectorFromSet(map[string]string{
			labels.AcornPublicName: name,
		}),
		Namespace: namespace,
	}); err != nil {
		return err
	}

	if len(secretList.Items) == 1 {
		secretList.Items[0].DeepCopyInto(secret)
		return nil
	}

	return nil
```

thedadams (approved):

pkg/ref/expr.go
g-linville: I structured this in an intentional way to return the original `NotFound` error if there are more or less than 1 item in the secretList. I can restructure it similarly to your suggestion here (it would involve preserving the `err` from `r.req.Get` at the beginning and returning it instead of `nil` at the end of the function), but imo the code is a little bit tougher to follow if structured that way, even though it would be less nested.

thedadams: Got it. Leave it as-is. I like it.

thedadams (approved):

PR #1941, opened 2023-07-17
Fix a possible panic from occuring in buildclient.Stream

tylerslaton:
Just got a panic from the check of `response.Body` being nil as the `response` was nil when using the latest main release.

### Checklist

### DIFF
--- a/pkg/buildclient/client.go
+++ b/pkg/buildclient/client.go
@@ -35,11 +35,11 @@ func Stream(ctx context.Context, cwd string, streams *streams.Output, dialer Web
 	conn, response, err := dialer(ctx, wsURL(build.Status.BuildURL), map[string][]string{
 		"X-Acorn-Build-Token": {build.Status.Token},
 	})
-	if response.Body != nil {
+	if response != nil {
 		defer response.Body.Close()
 	}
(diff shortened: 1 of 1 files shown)


thedadams (approved):

pkg/buildclient/client.go
```diff
@@ -35,11 +35,11 @@ func Stream(ctx context.Context, cwd string, streams *streams.Output, dialer Web
 	conn, response, err := dialer(ctx, wsURL(build.Status.BuildURL), map[string][]string{
 		"X-Acorn-Build-Token": {build.Status.Token},
 	})
-	if response.Body != nil {
+	if response != nil {
```
njhale, line 38: Should this be `response != nil && response.Body != nil`?

tylerslaton: I thought about that but it is not needed since `response.Body` is guaranteed to be not nil. Documentation around this [here](https://pkg.go.dev/net/http#Response:~:text=The%20http%20Client%20and%20Transport%20guarantee%20that%20Body%20is%20always%0A%09//%20non%2Dnil).

njhale (approved):

PR #1943, opened 2023-07-17
Add an init function for logserver

g-linville:
This allows consumers of this package to import it and do nothing else. Darren requested this.

### Checklist

### DIFF
--- a/pkg/logserver/logserver.go
+++ b/pkg/logserver/logserver.go
@@ -24,6 +24,10 @@ type Server struct {
 	Debug          bool
 }
 
+func init() {
+	go StartServerWithDefaults()
+}
+
 // StartServerWithDefaults starts the server with default values
 func StartServerWithDefaults() {
 	logrus.SetLevel(logrus.InfoLevel)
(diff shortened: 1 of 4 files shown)


njhale (approved):
ergonomic!

pkg/logserver/logserver.go
```diff
@@ -24,6 +24,10 @@ type Server struct {
 	Debug          bool
 }
 
+func init() {
```
tylerslaton, line 27: Could we break this into its own package?

As this currently stands, this will always call `StartServerWithDefaults` even if the importer didn't directly call it. For me that would be unexpected since there are public functions and types for starting the server. Reading the thread I believe there should be a package in this folder called `init` that a user can import directly to do the init. That or just make everything be private and not directly callable. 

Thoughts?

iwilltry42: +1 for the init package.
That also seems to be what Darren recommended: https://acorn-io.slack.com/archives/C03FU91U4SX/p1689620880004009?thread_ts=1689620609.332749&cid=C03FU91U4SX

g-linville: Fixed

tylerslaton (approved):
Nice!

PR #1950, opened 2023-07-19
Fix RBAC and local image resolution for acorn copy

g-linville:
This fixes two issues related to `acorn copy`:

- RBAC was previously not set for it at all
- The CLI would potentially utilize credentials for Docker Hub when the user was actually referring to a local image, which would lead to errors

### Checklist

### DIFF
--- a/pkg/cli/copy.go
+++ b/pkg/cli/copy.go
@@ -57,6 +57,8 @@ func (a *ImageCopy) Run(cmd *cobra.Command, args []string) error {
 	}
 
 	if !a.AllTags {
+		// Check if the source argument matches the name of a local image, and use it if it does.
+		// If there is an error, ignore it and move on, assuming the argument is a remote image name.
 		if _, err := c.ImageGet(cmd.Context(), args[0]); err == nil {
 			return a.copyLocalToRemote(cmd, c, args, creds)
 		}
--- a/pkg/cli/copy.go
+++ b/pkg/cli/copy.go
@@ -56,17 +56,23 @@ func (a *ImageCopy) Run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	if !a.AllTags {
+		if _, err := c.ImageGet(cmd.Context(), args[0]); err == nil {
+			return a.copyLocalToRemote(cmd, c, args, creds)
+		}
+	}
+
 	source, err := name.ParseReference(args[0])
 	if err != nil {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/cli/copy.go
```diff
@@ -56,17 +56,23 @@ func (a *ImageCopy) Run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	if !a.AllTags {
+		if _, err := c.ImageGet(cmd.Context(), args[0]); err == nil {
+			return a.copyLocalToRemote(cmd, c, args, creds)
+		}
```
thedadams, line 62: So if the `ImageGet` call errors, what does that mean? Can you add a comment here describing this scenario?

pkg/cli/copy.go
```diff
@@ -56,17 +56,23 @@ func (a *ImageCopy) Run(cmd *cobra.Command, args []string) error {
...
+	}
+
 	source, err := name.ParseReference(args[0])
 	if err != nil {
 		return err
 	}
 
-	dest, err := name.ParseReference(args[1])
+	sourceAuth, _, err := creds.Get(cmd.Context(), source.Context().RegistryStr())
```
thedadams, line 70: Is there a reason why these are being switched?

g-linville: Nope, the reordering was unintentional as I moved stuff around. I can switch the order back if you think it makes more sense the other way.

thedadams: No, it's totally fine. I was more or less wondering if this was important and needed a comment, too.

pkg/cli/copy.go
g-linville: I added a comment to the code. If ImageGet returns an error, then either the lookup failed for some reason or the argument just doesn't exist as a local image. Either way we don't really care and can just move on, assuming the user is referring to a remote image.

pkg/cli/copy.go
```diff
@@ -57,6 +57,8 @@ func (a *ImageCopy) Run(cmd *cobra.Command, args []string) error {
 	}
 
 	if !a.AllTags {
+		// Check if the source argument matches the name of a local image, and use it if it does.
```
thedadams, line 60: Awesome! Thank you.

thedadams (approved):

StrongMonkey (approved):

tylerslaton (approved):

PR #1954, opened 2023-07-20
Refactor resources to be a map struct

tylerslaton:
@njhale Brought up a good point that this code didn't originally lend itself well to extension. I'm updating the `Resources` struct to instead be a `map` with defined `enum` types. This will allow external packages to create their own resource concepts easily without having to add anything to the runtime.

### Checklist


pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -7,65 +7,97 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 )
 
+type (
+	Flag            string
+	Count           string
+	PersistentCount string
+	Quantity        string
```
g-linville, line 14: I recommend adding a comment explaining the meaning of each of these.

pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -7,65 +7,97 @@ import (
...
+	// Add the incoming resources to the current ones
+	for flag, value := range incoming.Flags {
+		current.Flags[flag] = value
+	}
+	for count, value := range incoming.Counts {
+		current.Counts[count] += value
+	}
+	for quantity, value := range incoming.Quantities {
+		q := current.Quantities[quantity]
+		q.Add(value)
```
g-linville, line 70: question: does `q` need to be set back on the map again after this?

`current.Quantities[quantity] = q`

pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -7,65 +7,97 @@ import (
...
 }
 
 // Remove will remove the resources of another Resources struct from the current one.
 func (current *Resources) Remove(incoming Resources, all bool) {
-	// Do not allow resources to go below 0
-	nonNegativeSubtract := func(currentVal, incomingVal int) int {
-		difference := currentVal - incomingVal
-		if difference < 0 {
-			difference = 0
+	nonZeroSubtract := func(a, b int) int {
```
g-linville, line 79: I like the improvement to this function but I think the previous name made more sense:

```suggestion
	nonNegativeSubtract := func(a, b int) int {
```

pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -7,65 +7,97 @@ import (
...
-	current.Jobs = nonNegativeSubtract(current.Jobs, incoming.Jobs)
-	current.Volumes = nonNegativeSubtract(current.Volumes, incoming.Volumes)
-	current.Images = nonNegativeSubtract(current.Images, incoming.Images)
-	current.Projects = nonNegativeSubtract(current.Projects, incoming.Projects)
+	for c, value := range incoming.Counts {
+		current.Counts[c] = nonZeroSubtract(current.Counts[c], value)
+	}
+	for quantity, value := range incoming.Quantities {
+		q := current.Quantities[quantity]
+		q.Sub(value)
```
g-linville, line 91: Same question here.

pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -75,31 +107,21 @@ func (current *Resources) Remove(incoming Resources, all bool) {
...
+			exceededResources = append(exceededResources, string(count))
 		}
 	}
-
-	// Define function for checking quantity resources to keep code DRY
-	checkQuantityResource := func(resource string, currentVal, incomingVal resource.Quantity) {
-		if currentVal.Cmp(incomingVal) <= 0 {
-			exceededResources = append(exceededResources, resource)
+	for persistentCount, value := range incoming.PersistentCounts {
+		if current.PersistentCounts[persistentCount] < value {
```
g-linville, line 116: nit: I noticed that the previous code was using `<=` here, should that be the case for the new code as well?

pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -109,52 +131,62 @@ func (current *Resources) Fits(incoming Resources) error {
...
-		current.Jobs == incoming.Jobs &&
-		current.Volumes == incoming.Volumes &&
-		current.Secrets == incoming.Secrets &&
-		current.Images == incoming.Images &&
-		current.Projects == incoming.Projects &&
-		current.VolumeStorage.Cmp(incoming.VolumeStorage) == 0 &&
-		current.Memory.Cmp(incoming.Memory) == 0 &&
-		current.CPU.Cmp(incoming.CPU) == 0
+	for flag, value := range current.Flags {
+		if incoming.Flags[flag] != value {
```
thedadams, line 172: What if there are values in `incoming` that aren't in `current`?

pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -7,65 +7,97 @@ import (
...
-	current.Images += incoming.Images
-	current.Projects += incoming.Projects
-
-	current.VolumeStorage.Add(incoming.VolumeStorage)
-	current.Memory.Add(incoming.Memory)
-	current.CPU.Add(incoming.CPU)
+	// Add the incoming resources to the current ones
+	for flag, value := range incoming.Flags {
+		current.Flags[flag] = value
+	}
```
thedadams, line 64: This makes it possible for you to "add" incoming resources and flip unlimited to `false`. Is that the desired behavior?

pkg/controller/quota/quota.go
```diff
@@ -113,8 +119,10 @@ func addCompute(containers map[string]apiv1.Container, appInstance *apiv1.AppIns
 			requirements = all.Requirements
 		}
 
-		quotaRequest.Spec.Resources.CPU.Add(requirements.Requests["cpu"])
-		quotaRequest.Spec.Resources.Memory.Add(requirements.Requests["memory"])
+		cpu := quotaRequest.Spec.Resources.Quantities[adminv1.CPU]
+		cpu.Add(requirements.Requests["cpu"])
```
thedadams, line 123: I assume you broke this up because the compiler was giving an error when it was combined.

If that is the case, then `cpu` needs to be put back in the map after it is modified.

Same for the others below, and the ones that @g-linville called out.

thedadams (changes requested):

pkg/controller/quota/quota.go
tylerslaton: How would you and @g-linville feel about making the Quantities in the map be pointers? This would make it so that the mutating function called to resources.Quantity types would persist.

g-linville: That sounds good to me.

pkg/apis/internal.admin.acorn.io/v1/resources.go
tylerslaton: Good point. I've overhauled the approach a bit. Please let me know what you think!

pkg/apis/internal.admin.acorn.io/v1/resources.go
tylerslaton: Yeah, I did this intentionally here. My thought was that there is some nice utility when you want to combine a bunch of resources together to be aggregated.

pkg/apis/internal.admin.acorn.io/v1/resources.go
tylerslaton: Yep, good catch. Talked about below but I've gone ahead and updated the Quantities to be pointers. This will eliminate the need for this.

pkg/apis/internal.admin.acorn.io/v1/resources.go
tylerslaton: Done!

pkg/apis/internal.admin.acorn.io/v1/resources.go
tylerslaton: Yeah, just a typo. Good catch.

pkg/apis/internal.admin.acorn.io/v1/resources.go
tylerslaton: Answer above ^ Thanks for bringing this up.

pkg/apis/internal.admin.acorn.io/v1/resources.go
tylerslaton: Yeah so I thought about this when doing this refactor. I think it should actually be `<` because if it is `<=` then this case will fail if the quantity we're comparing to equal. This is not desirable since that is actually a success case. This lead me to add a test suite. Let me know what you think.

pkg/apis/internal.admin.acorn.io/v1/resources.go
thedadams: So if one of a user's quota is unlimited and another one is not, then the aggregated quota could not be unlimited? The previous implementation was that if one of the quotas was unlimited, then the aggregated one would be. You're saying this was an intentional change?

pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -75,86 +128,78 @@ func (current *Resources) Remove(incoming Resources, all bool) {
...
-		current.CPU.Cmp(incoming.CPU) == 0
+	// Convert both structs to JSON and compare the JSON representations.
+	currentJSON, err1 := json.Marshal(current)
+	incomingJSON, err2 := json.Marshal(incoming)
+
+	if err1 != nil || err2 != nil {
+		return false // Unable to marshal, return false.
+	}
+
+	return string(currentJSON) == string(incomingJSON)
```
thedadams, line 204: Should `Resource{}` and `Resource{Flags: Flags{Unlimited: false}}` be considered equal?

They seem semantically equal to me, but I don't think this implementation would consider them to be equal.

pkg/apis/internal.admin.acorn.io/v1/resources.go
tylerslaton: Oh I misread your comment, I'm sorry. I see what you're saying now and will fix it.

tylerslaton: Okay, I went ahead and removed it. I think it makes more sense as it was before than trying to clean-up the toggle-off scenario.

pkg/controller/quota/quota.go
tylerslaton: As I started implementing this doing nil checks everywhere got to be more of a hassle so I just ended up add it back onto the map. like you originally suggested @g-linville.

pkg/controller/quota/quota.go
```diff
@@ -142,7 +153,9 @@ func addStorage(appInstance *apiv1.AppInstance, quotaRequest *adminv1.QuotaReque
 		if err != nil {
 			return err
 		}
-		quotaRequest.Spec.Resources.VolumeStorage.Add(parsedSize)
+		volumeStorage := quotaRequest.Spec.Resources.Quantities[adminv1.VolumeStorage]
+		volumeStorage.Add(parsedSize)
```
g-linville, line 157: This one still needs to get put back into the map.

pkg/apis/internal.admin.acorn.io/v1/resources_test.go
```diff
@@ -0,0 +1,264 @@
...
+				Quantities: Quantities{VolumeStorage: resource.MustParse("1Mi")},
+			},
+			incoming: Resources{
+				Counts:     Counts{Apps: 1},
+				Quantities: Quantities{VolumeStorage: resource.MustParse("1Mi")},
+			},
+			expected: Resources{},
+		},
+		{
+			name: "remove persistent counts with all",
```
g-linville, line 116: Can you add one more test after this one to make sure that PersistentCounts are not affected when `all` is false? Could help us prevent an accidental and nasty bug in the future.

g-linville (approved):

tylerslaton:
Going to be thinking about this a bit more after review. Will open it out of draft when it's ready.

PR #1955, opened 2023-07-20
Introduce a login retry when a user's token is expired

thedadams:
Additionally, a user-agent string is added to requests to manager.

### DIFF
--- a/pkg/manager/client.go
+++ b/pkg/manager/client.go
@@ -6,11 +6,18 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"runtime"
 
+	"github.com/acorn-io/runtime/pkg/version"
 	"github.com/sirupsen/logrus"
 )
 
-var ErrTokenNotFound = fmt.Errorf("token not found")
+var (
+	ErrTokenNotFound = fmt.Errorf("token not found")
+	ErrForbidden     = fmt.Errorf("forbidden")
+
+	userAgent = fmt.Sprintf("acorn/%s (%s; %s)", version.Get().String(), runtime.GOOS, runtime.GOARCH)
+)
 
 type tokenRequest struct {
 	Spec   tokenRequestSpec   `json:"spec,omitempty"`
@@ -46,11 +53,10 @@ type accountStatus struct {
 
 func httpDelete(ctx context.Context, url, token string) {
 	logrus.Debugf("Delete %s", url)
-	req, err := http.NewRequest(http.MethodDelete, url, nil)
+	req, err := newRequest(url, http.MethodDelete, token)
 	if err != nil {
 		return
 	}
-	req.Header.Set("Authorization", "Bearer "+token)
 
 	resp, err := http.DefaultClient.Do(req.WithContext(ctx))
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/project/client.go
```diff
@@ -211,9 +212,20 @@ func getClient(ctx context.Context, cfg *config.CLIConfig, opts Options, project
...
+						if err != nil {
+							return nil, err
+						}
+						loginRetry = false
+						continue
+					} else if err != nil {
+						return nil, err
+					}
+					break
 				}
```
njhale, line 229: @thedadams could you explain the intended behavior here? It looks like the `for` loop is never executed more than once.

pkg/manager/manager.go
```diff
@@ -121,9 +125,31 @@ func Login(ctx context.Context, password, address string) (user string, pass str
...
+		return user, pass, err
+	}
+
+	// reload config, could have changed
+	if newCfg, err := config.ReadCLIConfig(); err != nil {
+		return user, pass, err
+	} else {
+		*cfg = *newCfg
+	}
+	return user, pass, nil
```
njhale, line 149:
```suggestion
	// reload config, could have changed
	newCfg, err := config.ReadCLIConfig()
	if err == nil {
	    *cfg = *newCfg
		return user, pass, err
	}
	
	return user, pass, err
```

pkg/project/client.go
thedadams: The for loop is executed at most twice.

If the ProjectURL call fails with a "forbidden" error on the first try, then we login, set `loginRetry` to false and `continue`. If the ProjectURL call fails for any reason in the second try, then we return the error.

If all errors are nil, then we get to the break.

pkg/manager/client.go
```diff
@@ -6,11 +6,18 @@ import (
...
+	"github.com/acorn-io/runtime/pkg/version"
 	"github.com/sirupsen/logrus"
 )
 
-var ErrTokenNotFound = fmt.Errorf("token not found")
+var (
+	ErrTokenNotFound = fmt.Errorf("token not found")
+	ErrForbidden     = fmt.Errorf("forbidden")
+
+	userAgent = fmt.Sprintf("acorn/%s (%s; %s)", version.Get().String(), runtime.GOOS, runtime.GOARCH)
```
StrongMonkey, line 19: Nit: s/acorn/runtime/g 😄

StrongMonkey (approved):

pkg/manager/client.go
thedadams: The client in this case is the `acorn` CLI. I don't think a user would know anything about `runtime` if they are only using the CLI with manager.

StrongMonkey: Good point

tylerslaton (approved):

pkg/manager/manager.go
```diff
@@ -121,9 +125,31 @@ func Login(ctx context.Context, password, address string) (user string, pass str
...
+	}
+
+	if err = store.Add(ctx, apiv1.Credential{
+		ServerAddress: address,
+		Username:      user,
+		Password:      &pass,
+		LocalStorage:  true,
+	}, true); err != nil {
+		return user, pass, err
+	}
```
g-linville, line 141: What does this new store refer to? Is it just storing the credentials in the local CLI config file?

pkg/project/client.go
njhale: Ah, I see it now. Thanks!

pkg/manager/manager.go
thedadams: Yes, that's correct.

Before this change, the login `Run` function would call `Login` and then store the credential. I moved this is in here so that I could use the `Login` function in the new place and get the new credential stored automatically.

I hope that makes sense.

njhale (approved):

Issue #1956, opened 2023-07-20
"context deadline exceeded" errors relating to events seen repeatedly in api-server logs.


sangee2004:
acorn version - `v0.8.0-alpha7-131-g4bf05b4d+4bf05b4d`

I see a lot of these errors in api-server logs. I am not aware of any specific steps that leads to seeing these errors messages.

```
time="2023-07-20T19:12:32Z" level=warning msg="error forwarding events: [context deadline exceeded]"
time="2023-07-20T19:16:11Z" level=warning msg="error forwarding events: [context deadline exceeded]"
time="2023-07-20T19:19:38Z" level=warning msg="error forwarding events: [context deadline exceeded]"
time="2023-07-20T19:21:36Z" level=warning msg="error forwarding events: [context deadline exceeded]"
time="2023-07-20T19:29:28Z" level=warning msg="error forwarding events: [context deadline exceeded]"
time="2023-07-20T19:53:03Z" level=warning msg="error forwarding events: [context deadline exceeded]"
time="2023-07-20T19:54:08Z" level=warning msg="error forwarding events: [context deadline exceeded]"
time="2023-07-20T19:59:18Z" level=warning msg="error forwarding events: [context deadline exceeded]"
time="2023-07-20T20:01:37Z" level=warning msg="error forwarding events: [context deadline exceeded]"
```

cjellick:
@njhale is going to revisit eventually

njhale:
see https://github.com/acorn-io/runtime/pull/1957#issuecomment-1697787854

PR #1957, opened 2023-07-20
Decrease log noise for event watch

njhale:
Omit error logs for event watch requests that exceed their context
deadline. This reduces log noise when clients don't close watch requests properly.

Addresses #1956

### DIFF
--- a/pkg/server/registry/apigroups/acorn/events/strategy.go
+++ b/pkg/server/registry/apigroups/acorn/events/strategy.go
@@ -51,8 +51,8 @@ func (s *eventStrategy) Watch(ctx context.Context, namespace string, opts storag
 	go func() {
 		defer close(result)
 
-		if err := q.filterChannel(ctx, events, result); !channels.NilOrCanceled(err) {
-			logrus.Warnf("error forwarding events: [%v]", err)
+		if err := q.filterChannel(ctx, events, result); !channels.NilOrCanceled(err) && !errors.Is(err, context.DeadlineExceeded) {
+			logrus.WithError(err).Warn("error forwarding events")
 		}
 	}()
 


pkg/server/registry/apigroups/acorn/events/strategy.go
```diff
@@ -51,8 +51,8 @@ func (s *eventStrategy) Watch(ctx context.Context, namespace string, opts storag
 	go func() {
 		defer close(result)
 
-		if err := q.filterChannel(ctx, events, result); !channels.NilOrCanceled(err) {
-			logrus.Warnf("error forwarding events: [%v]", err)
+		if err := q.filterChannel(ctx, events, result); !channels.NilOrCanceled(err) && !errors.Is(err, context.DeadlineExceeded) {
```
njhale, line 54: nit: makes me want `z.ErrorIs(err, context.Canceled, context.DeadlineExceeded)`

thedadams:
The only reason I am hesitant about this change is that we had some watch context deadline exceeded errors previously that were symptoms of an actual issue. Are we confident that these deadline exceeded errors are not a problem?

g-linville:
Ditto what Donnie said. I think we might be missing another LongRunningResource.

njhale:
> The only reason I am hesitant about this change is that we had some watch context deadline exceeded errors previously that were symptoms of an actual issue. Are we confident that these deadline exceeded errors are not a problem?

Fair enough. I'll print the request headers to see if I can track down the source of the request and go from there.

g-linville:
> @g-linville IIUC long running resources are specifically for custom sub-resources. I don't know of an instance where we set it for a "normal" resource like events.

I think that is what we typically use it for but I think it might be worth trying for events also, if mink will accept it for non-subresources. If the events take so long to stream that we hit the context deadline, then that seems like the most straightforward way to fix it, unless there is also a way to somehow make the event streaming faster.

njhale:
I'm going to close this PR out since it's gone stale; @ibuildthecloud  has mentioned that he doesn't like the [channel abstractions](https://github.com/acorn-io/runtime/blob/main/pkg/channels/channels.go) I'm using anyway, so I was already planning on refactoring this at some point -- maybe I'll spot something when I do.

Issue #1672, opened 2023-05-30
Failed to update managedFields for `Events`


njhale:
## Description

The `acorn-api` fails to update the managedFields for `Event` objects, resulting in the following error message at runtime:

```sh
$ kubectl -n acorn-system logs deployments/acorn-api | grep 'managedFields'
E0526 15:37:57.167900       1 fieldmanager.go:152] "[SHOULD NOT HAPPEN] failed to update managedFields" err="failed to convert new object (acorn/d0a61e151e749976f8f5f80b9d33589b; api.acorn.io/v1, Kind=Event) to smd typed: .details.resourceVersion: expected map, got &{<map[string]interface {} Value> resourceVersion 1144 6}" versionKind="api.acorn.io/v1, Kind=Event" namespace="acorn" name="d0a61e151e749976f8f5f80b9d33589b"
```

## Version Info

runtime (acorn-io/acorn) commit: `ea361703448b82cf1b612ebc9d22b8c03b89b8f0`

## Repro. Steps

Run an app that results in an image being pulled; e.g. `acorn run ghcr.io/acorn-io/library/hello-world`

## Thoughts/Notes

It looks like field manager has trouble converting `GenericMap` into `AppImagePullEventDetails`; the `resourceVersion` field in particular, which is [set here][1].

[1]: https://github.com/acorn-io/acorn/blob/b0bec71ef59e9eaf272f8b758f2c542b331271bb/pkg/controller/appdefinition/pullappimage.go#L147

sangee2004:
Tested with acorn version `v0.8.0-alpha7-151-gc1b7f440+c1b7f440`

Deploy few apps from images in external registry.
Deploy few apps from Acornfile.

No error messages relating to update manageFileds seen in api logs . Checked using` kubectl -n acorn-system logs deployments/acorn-api | grep 'managedFields'`

PR #1961, opened 2023-07-21
Fix openapi schema generation for event fields

njhale:
Set formatting and extensions for the observed and details fields
respectively. This prevents field manager from complaining when events
are created.

Addresses #1672

### DIFF
--- a/pkg/openapi/map.go
+++ b/pkg/openapi/map.go
@@ -10,7 +10,7 @@ func GetOpenAPIDefinitions(ref common.ReferenceCallback) map[string]common.OpenA
 	result := generated.GetOpenAPIDefinitions(ref)
 	for _, v := range result {
 		for name := range v.Schema.SchemaProps.Properties {
-			if name == "deployArgs" || name == "buildArgs" || name == "params" {
+			if name == "deployArgs" || name == "buildArgs" || name == "params" || name == "details" {
 				v.Schema.SchemaProps.Properties[name] = spec.Schema{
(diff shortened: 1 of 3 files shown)


pkg/openapi/map.go
```diff
@@ -10,7 +10,7 @@ func GetOpenAPIDefinitions(ref common.ReferenceCallback) map[string]common.OpenA
 	result := generated.GetOpenAPIDefinitions(ref)
 	for _, v := range result {
 		for name := range v.Schema.SchemaProps.Properties {
-			if name == "deployArgs" || name == "buildArgs" || name == "params" {
+			if name == "deployArgs" || name == "buildArgs" || name == "params" || name == "details" {
```
ibuildthecloud, line 13: Oh, I forgot about this hack.  This explains a lot of errors I see because this list is not comprehensive for all the places we use GenericMap.  There's got to be a better more proper approach for this.

njhale: I burnt a lot of time trying to sniff out a better way to do this. I eventually stumbled onto https://github.com/kubernetes/kube-openapi/tree/master/pkg/generators#custom-openapi-type-definitions 

... unfortunately, `func (_ GenericMap) OpenAPIDefinition() openapi.OpenAPIDefinition {//...` wasn't doing the trick (curiously enough, implementing the other two methods for `MicroTime` did though).

g-linville (approved):

tylerslaton (approved):

iwilltry42 (approved):

PR #1979, opened 2023-07-25
Wait for all parent apps to be deleted before deleting children

thedadams:
When a project is deleted, we clean up all apps. The problem with this current implementation is that all apps will be deleted, regardless of whether it was a child app. This leads to child apps have issues deleting.

This change will first ensure all "parent acorns" are deleted first, giving the child acorns time to delete. After all such acorns are deleted, then the remaining child acorns are also deleted.

### Checklist

### DIFF
--- a/pkg/project/handlers.go
+++ b/pkg/project/handlers.go
@@ -10,6 +10,8 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	klabels "k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
 	kclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/project/handlers.go
```diff
@@ -46,21 +48,35 @@ func CreateNamespace(req router.Request, resp router.Response) error {
...
+	if err != nil {
 		return err
 	}
 
-	for _, app := range apps.Items {
-		if app.DeletionTimestamp.IsZero() {
-			if err := req.Client.Delete(req.Ctx, &app); err != nil && !apierrors.IsNotFound(err) {
-				return err
+	// First delete all "parent" apps. If no "parent" apps exist, then ensure that all apps are deleted.
+	for _, ls := range []klabels.Selector{klabels.NewSelector().Add(*parentAppRequirement), klabels.Everything()} {
```
njhale, line 58: `klabels.Everything()` will definitely work, but is there a specific reason this was chosen over something like `childAppRequirement`?

njhale (approved):
Nice!

pkg/project/handlers.go
thedadams: Two reasons I did it this way:
1. It's obvious that we want to clean up everything; and
2. One fewer error to handle.

tylerslaton (approved):

PR #1990, opened 2023-07-28
Don't require an API connection for render on a file

ibuildthecloud:

### DIFF
--- a/pkg/cli/render.go
+++ b/pkg/cli/render.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/acorn-io/aml/pkg/cue"
 	cli "github.com/acorn-io/runtime/pkg/cli/builder"
+	client2 "github.com/acorn-io/runtime/pkg/client"
 	"github.com/acorn-io/runtime/pkg/imagesource"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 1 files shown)


pkg/cli/render.go
```diff
@@ -5,6 +5,7 @@ import (
 
 	"github.com/acorn-io/aml/pkg/cue"
 	cli "github.com/acorn-io/runtime/pkg/cli/builder"
+	client2 "github.com/acorn-io/runtime/pkg/client"
```
njhale, line 8: nit: maybe `rclient`? also, it doesn't look like `client` is being shadowed by a variable (even if it's not a great name)

pkg/cli/render.go
```diff
@@ -27,12 +28,21 @@ type Render struct {
...
+
+	_, file, err := imageAndArgs.ResolveImageAndFile()
 	if err != nil {
 		return err
 	}
-
-	imageAndArgs := imagesource.NewImageSource(s.File, args, s.Profile, nil, false)
+	if file == "" {
+		// Lazily create client so that local file render doesn't require an API connection
+		c, err = s.client.CreateDefault()
```
njhale, line 41: I may be mistaken, but isn't this what the [DeferredClient](https://github.com/acorn-io/runtime/blob/5e6684245d93e7e2b256698da3fbacd90d3ed0bc/pkg/client/deferred.go#L13) is for?

njhale (approved):

Issue #1887, opened 2023-07-06
`acorn logs` don't work while things are deleting


cjellick:
`acorn logs` don't work while things are deleting.  That makes it hard to troubleshoot stuck deleting stuff.

> I'm pretty sure this is because I look at !DeletionTimestemp.IsZero() as when to decide to terminate

[Slack Message](https://acorn-io.slack.com/archives/C03FU91U4SX/p1688623023927599?thread_ts=1688623023.927599&cid=C03FU91U4SX)

keyallis:
I've added an integration test to cover this functionality, that test pulls in a new acornfile that will hang when a deletion is triggered and then checks to make sure it gets a new log message following that event

PR #1998, opened 2023-07-28
`acorn log` works through deletion period (#1887)

keyallis:
#1887

### Checklist

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -61,6 +61,10 @@ func (w *watching) shouldWatch(kind, namespace, name string) bool {
 	return true
 }
 
+func objectDeleted(obj metav1.Object) bool {
+	return !obj.GetDeletionTimestamp().IsZero() && len(obj.GetFinalizers()) == 0
+}
+
 type Message struct {
 	Line          string
 	Pod           *corev1.Pod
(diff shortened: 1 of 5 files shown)


tylerslaton (commented):
Implementation looks good to me. How would you feel about adding an integration test for this [here](https://github.com/acorn-io/runtime/blob/c3753292e77f39069d8a6fd665970560666de346/integration/log/log_test.go)?

pkg/log/log.go
```diff
@@ -61,6 +61,10 @@ func (w *watching) shouldWatch(kind, namespace, name string) bool {
 	return true
 }
 
+func objectDeleted(obj metav1.Object) bool {
```
tylerslaton, line 64: This cans/should be a follow-up (if we do want to do it), but I think this would be a good `z` candidate. What do you think?

tylerslaton: Ah and here's a link to `z` if you haven't seen it yet. 

https://github.com/acorn-io/z

keyallis: IMO this feels a bit more specialized than what it appears `z` is intended to capture. If in the future this ends up being a more useful utility we could put it in there, but for the time being I think it can stay in this file as an unexported func

pkg/log/log.go
```diff
@@ -224,7 +228,7 @@ func Container(ctx context.Context, pod *corev1.Pod, name string, output chan<-
 			}
 		} else if err == nil {
 			pod, err := options.PodClient.Pods(pod.Namespace).Get(ctx, pod.Name, metav1.GetOptions{})
-			if err == nil && (!pod.DeletionTimestamp.IsZero() || pod.Status.Phase == corev1.PodSucceeded || pod.Status.Phase == corev1.PodFailed) {
+			if err == nil && (objectDeleted(pod) || pod.Status.Phase == corev1.PodSucceeded || pod.Status.Phase == corev1.PodFailed) {
```
thedadams, line 231: I'm not sure if it is possible to get an object that has been deleted and has no finalizers, but it's probably fine here?

thedadams (approved):

g-linville (approved):

StrongMonkey (approved):

Issue #2000, opened 2023-07-31
Improve auto-upgrade enablement UX


tylerslaton:
There are some current UX gaps when it comes to setting auto-upgrade. We should improve them such that each of the below items are addressed:

* Keep the structure API as-is
* the server side/backend/daemon should only look at isPattern() or autoUpgrade == true. those are the only scenarios that puts an app into auto-upgrade mode. Implications of this:
  * If an app just has notifyUpgrade set, it is NOT autoUpgrade
  * If an app just has interval set, it is NOT autoUpgrade
  * If an app just has a pattern for its tag, it IS autoUpgrade
  * If app just as autoUpgrade set, it IS autoUpgrade
* CLI should compensate to turn on autoUpgrade if related fields are set. it should set autoUpgrade to true if:
  * the interval is set
  * if notifyUpgrade is set
  * the image is a pattern (maybe this one is not necessary and it could actually possibly complicate things)
* we'll have to consider update scenarios that turn off auto-upgrade. Seems like one cannot currently unset interval from the cli. if thats true, we should fix that. This is tracked in https://github.com/acorn-io/runtime/issues/2008

tylerslaton:
Related https://github.com/acorn-io/manager/issues/801

tylerslaton:
> we'll have to consider update scenarios that turn off auto-upgrade. Seems like one cannot currently unset interval from the cli. if thats true, we should fix that.
Everything except this is adjusted now. Going to be looking at that and the other linked issue.

tylerslaton:
Also realted https://github.com/acorn-io/runtime/issues/1996

sangee2004:
Tested with acorn version v0.8.0-alpha7-210-g2e8339a2+2e8339a2

Following tests are done from CLI:
1. Deploy apps with images in autoupgrade format (no --auto-upgrade  or --interval set or --notify-upgrade)
2. Deploy apps with only  `--notify-upgrade` set
3. Deploy apps with only `--interval` set
4. Deploy apps with only `--auto-upgrade` set

Autoupgrade gets triggered as expected in all the above cases.

sangee2004:
Following tests were done for nested Acorns:

Deploy nested apps using the following Acornfile:

```
acorns: {
    "my-acorn1": {
        image: "docker.io/sangeetha/myfirstacorn:v#.#.#"
    }
   "my-acorn2": {
       image: "docker.io/sangeetha/myfirstacorn:latest"
        notifyUpgrade: true
   }
[...]

PR #2002, opened 2023-07-31
Update auto-upgrade logic to have less assumptions

tylerslaton:
for #2000 

The original auto-upgrade assumed that setting an app to auto-upgrade could be done by:
1. Setting AutoUpgrade to true
2. Having a tag is an auto-upgrade pattern
3. Setting an interval
4. Setting NotifyUpgrade to true

Now the only two ways to set an app to auto-upgrade are (from an API level):
1. Setting AutoUpgrade to true
2. Having a tag is an auto-upgrade pattern

To patch up the UX such that it is the same as before from the CLI's perspective there is new code that sets `AutoUpgrade` to true if it is implied. This effectually mantains the same UX as before with the backend behaving in a more explicit manner.

### Checklist

### DIFF
--- a/pkg/autoupgrade/daemon.go
+++ b/pkg/autoupgrade/daemon.go
@@ -319,10 +319,12 @@ func AutoUpgradePattern(image string) (string, bool) {
 	return tag, strings.ContainsAny(tag, "#*")
 }
 
-// ImpliedAutoUpgrade returns boolean indicating if auto-upgrade is implied either
-// by a pattern in the app.Spec.image or app.Spec.AutoUpgradeInterval being specified.
-func ImpliedAutoUpgrade(interval, image string) bool {
-	if _, isPattern := AutoUpgradePattern(image); isPattern || interval != "" {
+// ImpliedAutoUpgrade returns boolean indicating if auto-upgrade is implied either:
+// 1. A pattern specified in the image
+// 2. A non-empty interval
+// 3. app.Spec.NotifyUpgrade set to true
+func ImpliedAutoUpgrade(image, interval string, notify bool) bool {
+	if _, isPattern := AutoUpgradePattern(image); isPattern || interval != "" || notify {
 		return true
 	}
 	return false
@@ -330,7 +332,7 @@ func ImpliedAutoUpgrade(interval, image string) bool {
 
 func Mode(appSpec v1.AppInstanceSpec) (string, bool) {
 	_, isPat := AutoUpgradePattern(appSpec.Image)
-	on := appSpec.GetAutoUpgrade() || appSpec.GetNotifyUpgrade() || isPat
+	on := appSpec.GetAutoUpgrade() || isPat
 
 	if !on {
 		return "", false
--- a/pkg/autoupgrade/daemon.go
+++ b/pkg/autoupgrade/daemon.go
@@ -319,11 +319,11 @@ func AutoUpgradePattern(image string) (string, bool) {
 	return tag, strings.ContainsAny(tag, "#*")
 }
 
-// ImpliedAutoUpgrade returns boolean indicating if auto-upgrade is implied either:
+// Implied returns a boolean indicating if auto-upgrade is implied by either:
 // 1. A pattern specified in the image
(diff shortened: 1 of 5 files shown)


pkg/controller/appdefinition/generation.go
```diff
@@ -16,7 +17,11 @@ func UpdateObservedFields(req router.Request, resp router.Response) error {
 
 func impliedAutoUpgrade(appSpec v1.AppInstanceSpec) bool {
```
tylerslaton, line 18: This is most likely no longer needed. Holding off on deleting for the moment until I know for sure that we don't want something keeping track of this at a Status level.

tylerslaton: Cancel that, I think it is actually still necessary since the situation where a AutoUpgrade pattern is set still counts for AutoUpgrade being enabled even if `app.Spec.AutoUpgrade` is not. 

Additionally, we use this status field in the table printer for Apps to show if AutoUpgrade is enabled.

pkg/autoupgrade/daemon_test.go
```diff
@@ -471,6 +474,10 @@ func TestDaemonSync(t *testing.T) {
 			app.Spec.AutoUpgrade = ptrTrue
 			app.Spec.Image = "acorn/acorn-enabled:latest"
 		case "notify-app":
+			app.Spec.AutoUpgrade = ptrTrue
+			app.Spec.NotifyUpgrade = ptrTrue
+			app.Spec.Image = "acorn/acorn-notify:latest"
+		case "notify-app-alone":
```
g-linville, line 480: I think the test cases in the table test below this switch need to be updated to add this new `"notify-app-alone"` case.

tylerslaton: So all of these test cases are actually testing all of the apps defined on line 466. As such, I explicitly did not add a case to test the `notify-app-alone` app since it should not appear in any of the results

pkg/autoupgrade/daemon.go
```diff
@@ -319,18 +319,20 @@ func AutoUpgradePattern(image string) (string, bool) {
...
+	if _, isPattern := AutoUpgradePattern(image); isPattern || interval != "" || notify {
 		return true
 	}
 	return false
 }
 
 func Mode(appSpec v1.AppInstanceSpec) (string, bool) {
 	_, isPat := AutoUpgradePattern(appSpec.Image)
-	on := appSpec.GetAutoUpgrade() || appSpec.GetNotifyUpgrade() || isPat
+	on := appSpec.GetAutoUpgrade() || isPat
```
thedadams, line 335: Do we care about existing apps? If a user has an existing app that has only "notify upgrade" enabled, it seems that their apps will stop getting upgraded after this change. Is that correct? Is that what we want?

tylerslaton: That's a good question. In my mind this is intended since this is no longer behavior we want to support. However, I think it's worth raising to the team to be sure so I'll go ahead and ask.

pkg/controller/appdefinition/generation.go
```diff
@@ -10,14 +10,11 @@ func UpdateObservedFields(req router.Request, resp router.Response) error {
 	app := req.Object.(*v1.AppInstance)
 	app.Status.ObservedImageDigest = app.Status.AppImage.Digest
 	app.Status.ObservedGeneration = app.Generation
-	app.Status.ObservedAutoUpgrade = impliedAutoUpgrade(app.Spec)
+	app.Status.ObservedAutoUpgrade = autoUpgradeEnabled(app.Spec)
```
thedadams, line 13: Is this field still needed?

thedadams (approved):

pkg/controller/appdefinition/generation.go
tylerslaton: Yes, I believe so. Namely for the case that a pattern is used but autoupgrade is not set. More context [here](https://github.com/acorn-io/runtime/pull/2002#discussion_r1281080837).

PR #2003, opened 2023-08-01
Redact sensitive info before logging build messages

njhale:
This prevents credential leakage when using higher log levels during builds.

Addresses https://github.com/acorn-io/manager/issues/828

### DIFF
--- a/pkg/buildclient/messages.go
+++ b/pkg/buildclient/messages.go
@@ -199,7 +199,7 @@ func (m *WebsocketMessages) run(ctx context.Context) error {
 		if err := m.conn.ReadJSON(msg); err != nil {
 			return err
 		}
-		logrus.Tracef("Read build message %s", msg)
+		logrus.Tracef("Read build message %s", redact(msg))
 		if m.handler != nil {
 			if err := m.handler(msg); err != nil {
 				return err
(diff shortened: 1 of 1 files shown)


pkg/buildclient/messages.go
```diff
@@ -229,8 +230,22 @@ func (m *WebsocketMessages) Recv() (<-chan *Message, func()) {
 }
 
 func (m *WebsocketMessages) Send(msg *Message) error {
-	logrus.Tracef("Send build message %s", msg)
+	logrus.Tracef("Send build message %s", redact(msg))
```
njhale, line 233: This could probably be achieved through a custom logrus formatter as well, but I think this is concise and works well enough.

tylerslaton: Yeah, no need to right now unless you can think of other places that need redactions.

pkg/buildclient/messages.go
```diff
@@ -229,8 +230,22 @@ func (m *WebsocketMessages) Recv() (<-chan *Message, func()) {
...
 	m.lock.Lock()
 	defer m.lock.Unlock()
 	return m.conn.WriteJSON(msg)
 }
+
+// redact removes sensitive information from a Message.
+// Use this before logging a message.
+func redact(msg *Message) *Message {
+	redacted := z.Dereference(msg)
```
ibuildthecloud, line 242: This is really a gratuitous use of z.Deference, don't use this in place of a `if msg != nil`.  Code tricks obscure the intention and decrease readibility. z.Dereference is mostly acceptable when dealing it native types (*int, *string, *bool) because they are very unnatural to interact with, but a struct pointer shouldn't fall into that bucket.

njhale: That's reasonable, I'll amend this to

```go
var redacted Message
if msg != nil {
    redacted = *msg
}
```

and update the godocs for `z.Dereference` to prevent future misuse

njhale: Done.

https://github.com/acorn-io/z/pull/6

g-linville (approved):

PR #2006, opened 2023-08-02
Propagate ports from main container service to additional services for the same container

g-linville:
This one is a bit tough to explain. Let's start with this Acornfile:

```cue
containers: "my-container": {
  image: "nginx:latest"
  ports: "80/http"
}

services: "nginx": {
  ports: "8080:80/http"
  container: "my-container"
}
```

Prior to this PR, the controller would render two Kubernetes Services for this. The first is  `my-container`, accessible on port 80 and sending traffic to port 80 on the `my-container` container. The second is `nginx`, accessible on port 8080 and sending traffic to port 80 on the `my-container` container. Fairly straightforward and works great.

...that is, until Istio comes into the picture. Istio [does not like](https://istio.io/latest/docs/reference/config/analysis/ist0137) when multiple services have the same target port on the same pod, but with different public-facing ports. When we were running like this, we had weird traffic routing issues where Envoy would forward traffic to a stale pod IP that no longer belonged to any pods.

The fix? When we create additional `services` for a specific container (i.e., the `container` field is specified, as it is in our example), propagate all of the HTTP ports from the "main" service for the container (the one whose name is the same as the name of the container). Istio will still complain when you run `istioctl analyze -A`, but the routing issues disappear and things work again.

Also in this PR, we block these additional `services` from being allowed to specify any non-HTTP ports.

I also added a new unit test to cover this case.

### Checklist

### DIFF
--- a/pkg/ports/ports.go
+++ b/pkg/ports/ports.go
@@ -1,12 +1,12 @@
 package ports
 
 import (
+	"sort"
 	"strconv"
 	"strings"
 
 	"github.com/acorn-io/baaah/pkg/typed"
 	v1 "github.com/acorn-io/runtime/pkg/apis/internal.acorn.io/v1"
-	"golang.org/x/exp/slices"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
@@ -85,8 +85,8 @@ func RemoveNonHTTPPorts(ports []corev1.ServicePort) []corev1.ServicePort {
 }
 
 func SortPorts(ports []corev1.ServicePort) []corev1.ServicePort {
-	slices.SortFunc[corev1.ServicePort](ports, func(i, j corev1.ServicePort) bool {
-		return i.Port < j.Port
+	sort.Slice(ports, func(i, j int) bool {
+		return ports[i].Port < ports[j].Port
 	})
 	return ports
 }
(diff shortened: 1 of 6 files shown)


pkg/services/services.go
```diff
@@ -41,7 +41,29 @@ func toContainerLabelsService(service *v1.ServiceInstance) (result []kclient.Obj
...
+			if port.Protocol != v1.ProtocolHTTP {
+				return nil, fmt.Errorf("container service %s has non-HTTP port %d\nservices defined for existing containers must contain only HTTP ports", service.Name, port.Port)
+			}
+		}
+
+		// Get the main ServiceInstance for this container.
+		mainService := &v1.ServiceInstance{}
+		if err = c.Get(ctx, kclient.ObjectKey{Name: service.Spec.Container, Namespace: service.Namespace}, mainService); err != nil {
+			return
+		}
```
thedadams, line 60: Are we guaranteed that such a service exists? Should we handle `IsNotFound` errors here?

g-linville: I believe it is guaranteed, yes. This still feels like an awkward assumption to make though, so I will add in some logic to ignore the `IsNotFound` error in case it comes up.

njhale (approved):

pkg/ports/ports.go
```diff
@@ -82,3 +83,10 @@ func RemoveNonHTTPPorts(ports []corev1.ServicePort) []corev1.ServicePort {
 	}
 	return result
 }
+
+func SortPorts(ports []corev1.ServicePort) []corev1.ServicePort {
+	slices.SortFunc[corev1.ServicePort](ports, func(i, j corev1.ServicePort) bool {
+		return i.Port < j.Port
+	})
```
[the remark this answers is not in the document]
g-linville: This is probably the function I was looking for. I knew something like this was out there but found this `slices.SortFunc` first.

tylerslaton (approved):

Issue #831, opened 2022-11-01
job restart


cjellick:

sangee2004:
Tested with acorn version -  `v0.8.0-rc4-29-g6a019fe7+6a019fe7`

Able to list and get jobs using `acorn job` command for jobs with or without a schedule.

Able to validate the jobs in following states - `pending`, `running` , `failing` and `completed`.

There is an issue with tracking state of the job when restart is attempted which is tracked in https://github.com/acorn-io/runtime/issues/2092

PR #2022, opened 2023-08-04
Add jobs command and jobs restart subcommand

tylerslaton:
for #831 

This PR adds the ability to list jobs and restart them.

### List all jobs in the project
```
$ acorn jobs
NAME               STATUS          LAST RUN       NEXT RUN       CREATED
purple-field.foo   completed       1m ago         33s from now   13m ago
purple-field.bar   completed       13m ago        N/A            13m ago
purple-field.baz   running         13m ago        N/A            13m ago
purple-field.bat   failed          1m ago         1m from now    13m ago
```

### Restart a job 
```
$ acorn jobs rs purple-field.bar
purple-field.bar

$ acorn jobs
NAME               STATUS          LAST RUN       NEXT RUN       CREATED
purple-field.bar   running         1s ago         N/A            1s ago
purple-field.foo   completed       1m ago         33s from now   13m ago
purple-field.baz   running         13m ago        N/A            13m ago
purple-field.bat   failed          1m ago         1m from now    13m ago
```

### Checklist


pkg/server/registry/apigroups/acorn/jobs/translator.go
```diff
@@ -0,0 +1,134 @@
...
+	"k8s.io/apiserver/pkg/storage"
+	kclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type Translator struct {
+	client kclient.Client
+}
+
+func (t *Translator) FromPublicName(ctx context.Context, namespace, name string) (string, string, error) {
+	parts := strings.Split(name, ".")
```
ibuildthecloud, line 26: There's a helper method to split: https://github.com/acorn-io/runtime/blob/85cfe8c01b57609ae5213adbed04c9e2611e4fb8/pkg/publicname/publicname.go#L23

It return appName, resourceName.  If the string has no `.` in it it will return appName as "".

ibuildthecloud, line 26: There's a helper method for this https://github.com/acorn-io/runtime/blob/85cfe8c01b57609ae5213adbed04c9e2611e4fb8/pkg/publicname/publicname.go#L23

It splits into parentName, childName.  If it can't split (no `.`) it will return the second argument as `""`

tylerslaton: Oh sweet, that's nice.

pkg/cli/testdata/MockClient.go
```diff
@@ -505,6 +511,67 @@ func (m *MockClient) ContainerReplicaPortForward(ctx context.Context, name strin
...
+			Status:     apiv1.JobStatus{},
+		}, nil
+
+	case "found.job":
+		return &apiv1.Job{
+			TypeMeta:   metav1.TypeMeta{},
+			ObjectMeta: metav1.ObjectMeta{Name: "found.job"},
+			Spec:       apiv1.JobSpec{AppName: "found"},
+			Status:     apiv1.JobStatus{},
+		}, nil
```
g-linville, line 557: nit:

```suggestion
	case "found", "found.job":
		return &apiv1.Job{
			TypeMeta:   metav1.TypeMeta{},
			ObjectMeta: metav1.ObjectMeta{Name: "found.job"},
			Spec:       apiv1.JobSpec{AppName: "found"},
			Status:     apiv1.JobStatus{},
		}, nil
```

pkg/client/jobs.go
```diff
@@ -0,0 +1,77 @@
...
+	}, job)
+}
+
+func (c *DefaultClient) JobList(ctx context.Context, opts *JobListOptions) ([]apiv1.Job, error) {
+	result := &apiv1.JobList{}
+
+	selector, err := k8slabels.Parse(fmt.Sprintf("%s=%s", labels.AcornManaged, "true"))
+	if err != nil {
+		return nil, err
+	}
```
g-linville, line 30: Is this label selector necessary? We are querying for an Acorn API object, so I don't think any of them should even exist without this label on them.

pkg/server/registry/apigroups/acorn/jobs/translator.go
```diff
@@ -0,0 +1,130 @@
...
+	appName, jobName := publicname.Split(name)
+
+	app := &apiv1.App{}
+	err := t.client.Get(ctx, router.Key(namespace, appName), app)
+	if err != nil {
+		return namespace, name, err
+	}
+
+	namespace = app.Status.Namespace
+	return namespace, strings.Split(jobName, ":")[0], nil
```
g-linville, line 36: I'm a little confused...how can the jobName have a `:` in it?

pkg/server/registry/apigroups/acorn/jobs/translator.go
```diff
@@ -0,0 +1,130 @@
...
+		Columns: apiv1.JobColumns{
+			App:   result.Spec.AppName,
+			State: determineState(job.Status),
+		},
+	}
+
+	return result
+}
+
+// DetermineState returns the state of the job based on the
```
g-linville, line 117: nit:

```suggestion
// determineState returns the state of the job based on the
```

pkg/server/registry/apigroups/acorn/jobs/translator.go
```diff
@@ -0,0 +1,130 @@
...
+	)
+
+	result := &apiv1.Job{
+		ObjectMeta: job.ObjectMeta,
+	}
+
+	result.Name = name
+	result.Namespace = namespace
+	result.OwnerReferences = nil
+	result.UID = job.UID
```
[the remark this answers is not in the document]
tylerslaton: Yep! You're right, thanks. I'll update this.

pkg/client/jobs.go
tylerslaton: This is some leftover code, thanks for catching it. I had a previous implementation where the client just got the underlying `batchv1.Job` object. However, that no logic longer exists and I will update this code accordingly.

pkg/client/jobs.go
```diff
@@ -0,0 +1,77 @@
...
+	return result.Items, nil
+}
+
+func (c *DefaultClient) JobDelete(ctx context.Context, name string) (*apiv1.Job, error) {
+	job, err := c.JobGet(ctx, name)
+	if apierrors.IsNotFound(err) {
+		return nil, nil
+	} else if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
tylerslaton: Right now we have to get the job so that we can return it at the end. When we do get that job and it doesn't exist it would be some extra work to then call a delete on it. Any thoughts on that?

tylerslaton:
Note to reviewers - I've updated this PR to now have a custom strategy. This is because `Job` encompasses both `batchv1.Job` and `batchv1.CronJob` resources under-the-hood (if a `schedule` is set). I could've added some fancy logic into the Translator to process this but creating a custom strategy felt like a slightly cleaner way of doing this.

Open question to those reviewing as well - programmatically, what makes sense to do when `acorn jobs rs <job with schedule>` is called? I can delete the underlying `CronJob`, but it will still wait until its `Schedule` is processed. The current thought is to delete any `batchv1.Job` objects associated with the CronJob and create a new `batchv1.Job` resource owned by it but that feels somewhat like a controller type operation.

pkg/server/registry/apigroups/acorn/jobs/translator.go
tylerslaton: This line was a causality of the switch to a custom strategy. However, I believe this is only `container` specific and not relevant to `jobs`. As a result this would've likely been removed anyway.

pkg/server/registry/apigroups/acorn/jobs/strategy.go
```diff
@@ -0,0 +1,257 @@
...
+			Field: fields.SelectorFromSet(fields.Set{"metadata.name": name}),
+		},
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	acornJobs := list.(*apiv1.JobList)
+	for _, acornJob := range acornJobs.Items {
+		if acornJob.Name == name {
```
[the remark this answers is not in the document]
tylerslaton: So this was part of a larger problem that I fixed. The List function was not properly responding to options sent to it. That is fixed, I updated the Field selector to be a label selector (due to the necessary translation of appName and jobName) and I've add the logic mentioned to check if there is only 1 and return it if there is.

g-linville (approved):
Lgtm. Imo the only sub-optimal thing about the UX here is that you have to wait the full schedule time for the CronJob to run again after doing `acorn job rs <cronjob name>`. But I don't think it is worth the complication of creating a new job to run it immediately.

pkg/server/registry/apigroups/acorn/jobs/restart.go
```diff
@@ -0,0 +1,102 @@
...
+		},
+		Spec: spec,
+	}
+	if err := s.client.Create(ctx, newJob); err != nil {
+		return err
+	}
+
+	// Update the cronjob status to reflect the new job being created for it.
+	s.client.Get(ctx, kclient.ObjectKey{Name: cron.Name, Namespace: cron.Namespace}, cron)
+	cron.Status.LastScheduleTime = &metav1.Time{Time: newJob.CreationTimestamp.Time}
```
tylerslaton, line 100: There is a gap where the CronJob is not really aware that the Job we create above is owned by it. As a result things like `Active` are not populate or (if we fill it) are not cleaned-up as Jobs go away. I will be thinking through this but would love any advice.

tylerslaton:
This PR has been further updated to now include three main changes such as the:
1. Use of `jobs/restart` as a sub-resource. This is necessary to properly restart `Job`s in Manager as well as Runtime.
2. Ability to run `acorn jobs rs` on a `CronJob` and have it create a `Job` using its template
3. Use of `AppStatus` in `Job` gets and lists. This is also necessary for Manager use.

pkg/server/registry/apigroups/acorn/jobs/restart.go
```diff
@@ -0,0 +1,105 @@
...
+		job := &batchv1.Job{}
+		err := s.client.Get(ctx, kclient.ObjectKey{Name: jobRef.Name, Namespace: jobRef.Namespace}, job)
+		if err != nil {
+			return err
+		}
+
+		if err := s.client.Delete(ctx, job); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
tylerslaton: I'm opinion to changes here. I added this as it felt unexpected that a Job created by the CronJob would remain running after a trigger `acorn restart`. This makes it pretty easy to have multiple overlapping instances of a Job running for a CronJob. Is that desired behavior?

pkg/server/registry/apigroups/acorn/jobs/restart.go
```diff
@@ -0,0 +1,105 @@
...
+	return &apiv1.JobRestart{}
+}
+
+func (s *restartStrategy) Create(ctx context.Context, obj types.Object) (types.Object, error) {
+	ri, _ := request.RequestInfoFrom(ctx)
+
+	if ri.Name == "" || ri.Namespace == "" {
+		return obj, nil
+	}
+	key := kclient.ObjectKey{Name: ri.Name, Namespace: ri.Namespace}
```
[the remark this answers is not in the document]
tylerslaton: I've updated this logic recently in an attempt to satisfy this logic. Please let me know what you think!

pkg/client/jobs.go
```diff
@@ -0,0 +1,63 @@
...
+
+	return result.Items, nil
+}
+
+func (c *DefaultClient) JobRestart(ctx context.Context, name string) error {
+	// Get the job so we can get the namespace that the batchv1 resource lives in.
+	job, err := c.JobGet(ctx, name)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
tylerslaton: You are correct. This logic has been updated to reflect this.

pkg/cli/jobs_restart.go
```diff
@@ -0,0 +1,42 @@
...
+
+func NewJobRestart(c CommandContext) *cobra.Command {
+	cd := &JobRestart{client: c.ClientFactory}
+	cmd := cli.Command(cd, cobra.Command{
+		Use: "restart [JOB_NAME...]",
+		Example: `
+acorn job restart app-name.job-name`,
+		SilenceUsage:      true,
+		Short:             "Restart a job",
+		Aliases:           []string{"rm", "kill", "rs"},
```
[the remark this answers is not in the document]
tylerslaton: I mainly added these to fit in with the the `acorn container rm/kill` command but as this has evolved I no-longer think that is relevant. I've gone ahead and removed them.

pkg/client/jobs.go
```diff
@@ -0,0 +1,55 @@
...
+		return result.Items[i].CreationTimestamp.After(result.Items[j].CreationTimestamp.Time)
+	})
+
+	return result.Items, nil
+}
+
+func (c *DefaultClient) JobRestart(ctx context.Context, name string) error {
+	rs := &apiv1.JobRestart{}
+
+	// Since "." is not allowed in k8s job name, we replace it with "_" in the Name
```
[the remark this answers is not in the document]
tylerslaton: Fixed!

pkg/controller/appstatus/jobs.go
```diff
@@ -180,3 +211,36 @@ func (a *appStatusRenderer) isJobReady(jobName string) (ready bool, err error) {
...
+// 	}
+// 	if status.ErrorCount > 0 {
+// 		return fmt.Sprintf("Running (Failed %v time(s))", status.ErrorCount)
+// 	}
+// 	if status.RunningCount > 0 {
+// 		return "Running"
+// 	}
+// 	return "Pending"
+// }
+
```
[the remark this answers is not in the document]
tylerslaton: Yep, fixed.

pkg/server/registry/apigroups/acorn/jobs/restart.go
```diff
@@ -0,0 +1,121 @@
...
+			GenerateName: cron.Name + "-",
+			Namespace:    cron.Namespace,
+			Labels:       template.Labels,
+			Annotations:  template.Annotations,
+			OwnerReferences: []metav1.OwnerReference{{
+				APIVersion: batchv1.SchemeGroupVersion.String(),
+				Kind:       "CronJob",
+				Name:       cron.Name,
+				UID:        cron.UID,
+			}},
```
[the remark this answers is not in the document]
tylerslaton: I went ahead and used it anyway to be more inline with what kubectl does.

pkg/server/registry/apigroups/acorn/jobs/restart.go
```diff
@@ -0,0 +1,118 @@
...
+
+func (s *restartStrategy) New() types.Object {
+	return &apiv1.JobRestart{}
+}
+
+func (s *restartStrategy) Create(ctx context.Context, obj types.Object) (types.Object, error) {
+	ri, _ := request.RequestInfoFrom(ctx)
+
+	if ri.Namespace == "" {
+		return obj, nil
```
thedadams, line 38: Should this return a `namespace is required` error? Additionally, should we error if a name is not provided?

pkg/server/registry/apigroups/acorn/jobs/restart.go
```diff
@@ -0,0 +1,118 @@
...
+
+	key := kclient.ObjectKey{Namespace: job.Status.JobNamespace, Name: job.Status.JobName}
+
+	cronToRestart := &batchv1.CronJob{}
+	if err = s.client.Get(ctx, key, cronToRestart); err == nil {
+		return obj, s.restartCronJob(ctx, cronToRestart)
+	}
+
+	if !apierrors.IsNotFound(err) {
+		return nil, err
```
thedadams, line 57: Do you want to return the `obj` if `err == nil`? Or just always?

pkg/client/jobs.go
```diff
@@ -0,0 +1,55 @@
...
+
+	return result.Items, nil
+}
+
+func (c *DefaultClient) JobRestart(ctx context.Context, name string) error {
+	rs := &apiv1.JobRestart{}
+
+	// Since "." is not allowed in k8s job name, we replace it with "-" in the Name
+	// section of the request and send the public name in the labels.
+	rs.Labels = map[string]string{labels.AcornPublicName: name}
```
thedadams, line 48: nit, so there is no ambiguity between the object and the request:
```suggestion
	rs := &apiv1.JobRestart{
		ObjectMeta: metav1.ObjectMeta{
			Name: name,
			Namespace: namespace,
			Labels: map[string]string{labels.AcornPublicName: name}
	}

	// Since "." is not allowed in k8s job name, we replace it with "-" in the Name
	// section of the request and send the public name in the labels.
```

thedadams: Second comment: if the `name` doesn't have to be changed to remove the `.`s, do you need this label?

pkg/server/registry/apigroups/acorn/jobs/restart.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	if err := s.client.Create(ctx, newJob); err != nil {
+		return err
+	}
+
+	// Update the cronjob status to reflect the new job being created for it.
+	err := s.client.Get(ctx, kclient.ObjectKey{Name: cron.Name, Namespace: cron.Namespace}, cron)
+	if err != nil {
+		return err
+	}
```
thedadams, line 115: You already have the the CronJob via the variable `cron`. Do you need to get it here again?

pkg/client/jobs.go
tylerslaton: This label is no longer needed since I can remove it using the `validator.NoValidation` logic you outlined above.

pkg/server/registry/apigroups/acorn/jobs/restart.go
tylerslaton: You are right, we should return the object always.

pkg/server/registry/apigroups/acorn/jobs/restart.go
tylerslaton: I was mainly afraid of conflict errors here as I'm editing the Jobs that the CronJob creates above. However, testing this has dissuaded my fears as I do not see any conflicts.

pkg/server/registry/apigroups/acorn/jobs/restart.go
```diff
@@ -0,0 +1,118 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apiserver/pkg/endpoints/request"
+	"k8s.io/apiserver/pkg/registry/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func NewRestart(c client.WithWatch) rest.Storage {
+	return stores.NewBuilder(c.Scheme(), &apiv1.JobRestart{}).
+		WithCreate(&restartStrategy{client: c}).
```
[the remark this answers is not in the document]
tylerslaton: I went with the `validator.NoValidation` approach here as that felt the most natural. Thank you!

integration/client/jobs/jobs_test.go
```diff
@@ -0,0 +1,165 @@
...
+
+	jobs, err := c.JobList(ctx, nil)
+	if err != nil {
+		t.Error(err)
+	}
+
+	assert.Len(t, jobs, 2)
+	if len(jobs) != 2 {
+		t.Error("Expected only 2 jobs")
+	}
```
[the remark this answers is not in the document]
tylerslaton: Fixed.

iwilltry42 (approved):

pkg/server/registry/apigroups/acorn/jobs/restart.go
thedadams: Additionally, I think we would want conflicts to be returned. For example, what if the conflict means that another run of the job kicked off on its own?

pkg/server/registry/apigroups/acorn/jobs/restart.go
```diff
@@ -0,0 +1,113 @@
...
+
+func (s *restartStrategy) Create(ctx context.Context, obj types.Object) (types.Object, error) {
+	ri, _ := request.RequestInfoFrom(ctx)
+
+	if ri.Namespace == "" || ri.Name == "" {
+		return obj, nil
+	}
+
+	// Find the job to restart by looking it up with the public name. The only reason we don't set the actual name
+	// to be the public name is because the public name includes a "." which is not allowed in k8s names.
```
thedadams, line 43: nit: outdated

thedadams (approved):

pkg/cli/jobs.go
```diff
@@ -0,0 +1,88 @@
...
+		jobs, err := c.JobList(cmd.Context(), nil)
+		if err != nil {
+			return err
+		}
+		for _, job := range jobs {
+			if slices.Contains(args, job.Name) {
+				out.Write(&job)
+			}
+		}
+	}
```
njhale, line 74: The behavior here seems a bit inconsistent between different arg cardinatilities:
A) Given one name, it can be an app or a job
B) Given > 1 name, they can only be jobs

I'd expect the behavior of A to apply to each argument when multiple are given.

Related: You can get `NotFound` errors with a single argument, but not with multiple.

pkg/server/registry/apigroups/acorn/jobs/restart.go
tylerslaton: Fixed!

integration/client/client.go
```diff
@@ -20,6 +20,19 @@ func NewImageWithSidecar(t *testing.T, namespace string) string {
...
+func NewImageWithJobs(t *testing.T, namespace string) string {
+	t.Helper()
+
+	c := helper.BuilderClient(t, namespace)
+	image, err := c.AcornImageBuild(helper.GetCTX(t), "../testdata/job/Acornfile", &client.AcornImageBuildOptions{
+		Cwd: "../testdata/job",
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
```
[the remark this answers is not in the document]
tylerslaton: Updated all the code relevant to the changes here. We'll have to do a larger refactor in a separate PR.

pkg/cli/jobs.go
tylerslaton: Alright, I pretty dramatically overhauled this logic to be simpler and more comprehensive. If we like these changes I'll also make a follow-up in the container replicas equivalent code, since what I'm changing was originally pulled from there.

pkg/server/registry/apigroups/acorn/jobs/restart.go
```diff
@@ -0,0 +1,112 @@
...
+	// it here we can ensure that the job is restarted immediately.
+	template := cron.Spec.JobTemplate
+	spec := template.Spec
+	spec.TTLSecondsAfterFinished = new(int32) // Want to delete the job immediately after it finishes.
+	newJob := &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: cron.Name + "-",
+			Namespace:    cron.Namespace,
+			Labels:       template.Labels,
+			Annotations:  template.Annotations,
```
g-linville, line 99: Does this match the behavior of the k8s controller for CronJobs? Does it just copy labels and annotations from the CronJob to the Job it creates for it? Idk whether it does or not but I think we should make sure we are consistent with that behavior.

g-linville (approved):

pkg/server/registry/apigroups/acorn/jobs/restart.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func (s *restartStrategy) restartCronJob(ctx context.Context, cron *batchv1.CronJob) error {
+	// Find all active jobs and delete them to make way for new ones.
+	for _, jobRef := range cron.Status.Active {
+		job := &batchv1.Job{}
+		err := s.client.Get(ctx, kclient.ObjectKey{Name: jobRef.Name, Namespace: jobRef.Namespace}, job)
+		if err != nil {
+			return err
+		}
```
thedadams, line 78: Getting the job here is no necessary.

tylerslaton: You're right, updated this now.

pkg/server/registry/apigroups/acorn/jobs/restart.go
```diff
@@ -0,0 +1,112 @@
...
+
+		// Delete the Job and set the propagation policy to foreground so that the dependent resources (Pods)
+		// created by the Job are also deleted.
+		opts := &client.DeleteOptions{PropagationPolicy: z.Pointer(metav1.DeletePropagationForeground)}
+		if err := s.client.Delete(ctx, job, opts); err != nil {
```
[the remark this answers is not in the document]
tylerslaton: Agreed and added.

pkg/server/registry/apigroups/acorn/jobs/restart.go
```diff
@@ -0,0 +1,108 @@
...
+	// Find all active jobs and delete them to make way for new ones.
+	for _, jobRef := range cron.Status.Active {
+		job := &batchv1.Job{ObjectMeta: metav1.ObjectMeta{Name: jobRef.Name, Namespace: jobRef.Namespace}}
+
+		// Delete the Job and set the propagation policy to foreground so that the dependent resources (Pods)
+		// created by the Job are also deleted.
+		opts := &client.DeleteOptions{PropagationPolicy: z.Pointer(metav1.DeletePropagationForeground)}
+		if err := s.client.Delete(ctx, job, opts); !apierrors.IsNotFound(err) {
+			return err
+		}
```
thedadams, line 81:
```suggestion
		if err := s.client.Delete(ctx, job, opts); err !=nil && !apierrors.IsNotFound(err) {
			return err
		}
```

thedadams (approved):

Issue #2026, opened 2023-08-07
`401` error when getting imageDetails in Manager


iwilltry42:
```
  ✗  ERROR:  error getting signature for image foo/bar:baz: HEAD https://foobar.dkr.ecr.us-east-2.amazonaws.com/v2/acorn/foo/bar/manifests/sha256-ad36c36a6ca075eda5a8758e5d6ec85724d4e739aba6c9eb7c72263fb4fd7db6.sig: unexpected status code 401 Unauthorized (HEAD responses have no body, use GET for details)
  ```
  
  Slack thread: https://acorn-io.slack.com/archives/C03FU91U4SX/p1691177374735949

**Comment by sangee2004** [comment]
Tested on `beta`

Able to get image details successfully  for both local and remote images that are signed.

`acorn image details <image-name>` fetches the signature details in `signatureDigest`  field.


Local image:

```
 acorn image sign  mytestnew --key ~/santest3.key -a qatag=ok
  •  Signing Image mytestnew (digest: index.docker.io/library/mytestnew@sha256:7a567c1f23887d340d0a30737c66d01d735e494116abbe0a8c19eb3ea5ec07be) using key /Users/sangeethahariharan/santest3.key
? Enter password for private key /Users/sangeethahariharan/santest3.key: ********
  ✔  Done: Pushed signature sha256:14037cebdd2dc1762d1d98d8db8a6ab73e2118c58ac437b8b7196dd8adcf6490
sangeethahariharan@Sangeethas-MacBook-Pro example1 % acorn images
REPOSITORY   TAG       IMAGE-ID
mytestnew    latest    7a567c1f2388
sangeethahariharan@Sangeethas-MacBook-Pro example1 % acorn image details mytestnew                               
---
items:
- appImage:
    acornfile: "args: {\n\tnewApp:  true\n\toldtext: \"old\"\n\tnewtext: \"v17\"\n}\ncontainers:
      {\n\tmyweb: {\n\timage: \"nginx\"\n\tports: publish: \"80/http\"\n\tfiles: {\n\t\t\"/usr/share/nginx/html/index.html\":
      std.ifelse(args.newApp, args.newtext, args.oldtext)\n\t\t\"/usr/share/nginx/html/test.html\":
      \ \"test1\"\n\t\t\"/usr/share/nginx/html/test2.html\": \"test2\"\n\t}\n}\n}\n"
[...]

PR #2027, opened 2023-08-07
change: properly use local auth when signing/verifying images and getting signatures in imagedetails

iwilltry42:
Ref #2026 (I couldn't really reproduce this issue, but the code was problematic anyway)

### Checklist

### DIFF
--- a/pkg/imagedetails/imagedetails.go
+++ b/pkg/imagedetails/imagedetails.go
@@ -82,6 +82,12 @@ func GetImageDetails(ctx context.Context, c kclient.Client, namespace, imageName
 	if err != nil {
 		return nil, err
 	}
+
+	opts, err = images.GetAuthenticationRemoteOptions(ctx, c, namespace, opts...)
+	if err != nil {
+		return nil, err
+	}
+
 	_, sigHash, err := acornsign.FindSignature(imgRef.Context().Digest(appImageWithData.AppImage.Digest), opts...)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


pkg/imagedetails/imagedetails.go
```diff
@@ -82,6 +82,12 @@ func GetImageDetails(ctx context.Context, c kclient.Client, namespace, imageName
 	if err != nil {
 		return nil, err
 	}
+
+	opts, err = images.GetAuthenticationRemoteOptions(ctx, c, namespace, opts...)
+	if err != nil {
+		return nil, err
+	}
```
g-linville, line 89: I don't think this piece is necessary. Without this PR, I am already able to run `acorn image details <private image>` and it uses my credentials properly to fetch the information.

iwilltry42: That's exactly the part I'm unsure about, since that's the point where I think https://github.com/acorn-io/runtime/issues/2026 could come from.
Since in the Manager setup, there won't be local auth for ECR, so without this, we'd have no auth at all to use here.
The other function up there (https://github.com/acorn-io/runtime/pull/2027/files/668bfe95105a7c374ea4049c7deb6c4436a337db#diff-c0f7bfbedf36fddcbdbede8bb9f17a86a8e8bd885e6f4fb2886ccd4c2b190a6aR76) does this within the function body and apparently didn't yield the error from the issue.
I couldn't reproduce the `401` myself, so this is more of a guess and would at least make the remote.Options consistent between `images.PullAppImageWithDataFiles` and `acornsign.FindSignature` :thinking:

g-linville: Okay, that works for me then. Hopefully this will fix the 401.

g-linville (approved):

njhale (approved):

PR #2028, opened 2023-08-07
Record lifecycle events for managed containers

njhale:
Record created, restarted, and deleted acorn events for containers
managed as part of an Acorn.

Addresses https://github.com/acorn-io/manager/issues/701


njhale:
We've decided not to implement this as a controller.

PR #2034, opened 2023-08-08
Add Cilium proxy visibility annotations to pods

g-linville:
These are needed in order to get HTTP metrics from Cilium.

### Checklist

### DIFF
--- a/pkg/controller/appdefinition/deploy.go
+++ b/pkg/controller/appdefinition/deploy.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"net/url"
 	"path"
 	"strconv"
@@ -486,6 +487,7 @@ func podAnnotations(appInstance *v1.AppInstance, container v1.Container) map[str
 		labels.AcornContainerSpec: containerAnnotation(container),
 	}
 	addPrometheusAnnotations(annotations, container)
+	addCiliumAnnotations(annotations, container)
 
 	images := map[string]string{}
 	addImageAnnotations(images, appInstance, container)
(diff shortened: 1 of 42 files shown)


pkg/controller/appdefinition/deploy.go
```diff
@@ -522,6 +524,18 @@ func addPrometheusAnnotations(annotations map[string]string, container v1.Contai
...
 
+func addCiliumAnnotations(annotations map[string]string, container v1.Container) {
+	var visibilityConfigs []string
+	for _, port := range container.Ports {
+		if port.Protocol == v1.ProtocolHTTP {
+			visibilityConfigs = append(visibilityConfigs, fmt.Sprintf("<Ingress/%d/TCP/HTTP>", port.TargetPort))
+		}
+	}
+	if len(visibilityConfigs) > 0 {
+		annotations[labels.CiliumProxyVisibility] = strings.Join(visibilityConfigs, ",")
```
thedadams, line 535: What are the chance that this exceeds 253 characters?

thedadams (approved):

iwilltry42 (approved):

pkg/labels/labels.go
```diff
@@ -71,6 +71,8 @@ const (
 	PrometheusScrape = "prometheus.io/scrape"
 	PrometheusPath   = "prometheus.io/path"
 	PrometheusPort   = "prometheus.io/port"
+
+	CiliumProxyVisibility = "policy.cilium.io/proxy-visibility"
```
tylerslaton, line 75: Think its worth using this package?

https://github.com/cilium/cilium/blob/696167cb625e3b61b42412a37f7b02adcce2ff3f/pkg/annotation/k8s.go#L107

tylerslaton (approved):

pkg/controller/appdefinition/deploy.go
iwilltry42: @thedadams Isn't 253 the character limit of the annotation prefix only?
I'm not aware on a limit for the annotation value :thinking: 
I just know that the total size of all keys and values must not exceed  256 KB.

g-linville: I just tested this with like 15 ports defined on one container (each is five digits long) and the annotation is created properly with no issues.

pkg/labels/labels.go
g-linville: I just tried it out but there are multiple mutual dependencies between us and Cilium, so adding this updated some of those and I'm a little afraid of causing a problem and would just prefer to have this in our code.

PR #2037, opened 2023-08-09
Ensure proper deletion of child apps that have child apps

thedadams:
There was a change to handle deleting apps in the proper order when projects are deleted. That change didn't take into account the fact that child apps could have child apps. This change addresses this specifically.

### DIFF
--- a/pkg/project/handlers.go
+++ b/pkg/project/handlers.go
@@ -10,8 +10,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	klabels "k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/utils/strings/slices"
 	kclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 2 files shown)


ibuildthecloud (approved):

pkg/project/handlers.go
```diff
@@ -57,34 +55,30 @@ func CreateNamespace(req router.Request, resp router.Response) error {
...
-			if app.DeletionTimestamp.IsZero() {
-				if err = req.Client.Delete(req.Ctx, &app); err != nil && !apierrors.IsNotFound(err) {
-					return err
-				}
+	// Note: using index here to avoid the loop variable issue.
+	for i := range apps.Items {
+		// If the app's parent is gone, then ensure this app is deleted.
+		if _, ok := existingApps[apps.Items[i].Labels[labels.AcornParentAcornName]]; !ok && apps.Items[i].DeletionTimestamp.IsZero() {
+			if err := req.Client.Delete(req.Ctx, &apps.Items[i]); err != nil && apierrors.IsNotFound(err) {
+				return err
```
njhale, line 75: I think it may be better to aggregate the errors and return after attempting to delete all the derelict apps.

pkg/project/handlers.go
```diff
@@ -57,34 +55,30 @@ func CreateNamespace(req router.Request, resp router.Response) error {
...
-		for _, app := range apps.Items {
-			if app.DeletionTimestamp.IsZero() {
-				if err = req.Client.Delete(req.Ctx, &app); err != nil && !apierrors.IsNotFound(err) {
-					return err
-				}
+	// Note: using index here to avoid the loop variable issue.
+	for i := range apps.Items {
+		// If the app's parent is gone, then ensure this app is deleted.
+		if _, ok := existingApps[apps.Items[i].Labels[labels.AcornParentAcornName]]; !ok && apps.Items[i].DeletionTimestamp.IsZero() {
+			if err := req.Client.Delete(req.Ctx, &apps.Items[i]); err != nil && apierrors.IsNotFound(err) {
```
njhale, line 74: do we want to also force remove other finalizers? if the parent was already deleted, doesn't that mean cleanup is already in a bad state?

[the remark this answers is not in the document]
thedadams: Thanks! Fixed.

njhale (approved):

pkg/project/handlers.go
thedadams: Not necessarily. If you delete an app that has a database, then the database will get leftover and may need to be cleaned up properly.

thedadams: I should also add that a user can use `--ignore-cleanup` on an app if the project is being deleted.

njhale: ah, totally, thanks

PR #2080, opened 2023-08-16
Refactor GenericMap to prevent managedField issues

njhale:
This change lets us generate the correct OpenAPI schema for `GenericMap` without the need to [mutate the schema for fields, by name, at runtime](https://github.com/acorn-io/runtime/blob/c24d9c1fa2771950f11561aa62554475cc6fcfaa/pkg/openapi/map.go#L13). It will stop the "managedFields" message from being printed for preexisting usages of `GenericMap` and prevent it for all new usages going forward.


njhale:
Looks like I'm still having some issues here. Will update.

njhale:
OK -- this is all ready to go

pkg/apis/internal.acorn.io/v1/appspec.go
```diff
@@ -749,7 +749,7 @@ type Secret struct {
 	Name        string            `json:"name,omitempty"`
 	Description string            `json:"description,omitempty"`
 	Type        string            `json:"type,omitempty"`
-	Params      GenericMap        `json:"params,omitempty"`
+	Params      *GenericMap       `json:"params,omitempty"`
```
ibuildthecloud, line 752: Since this is a pointer everywhere I feel like the chance of panic is extremely. Like I'm reading all the changes and trying to see if you first checked for nil.  What if you add a GetData() to `GenericMap` that will first check if the receiver is nil and then return nil.  So then existing code that did `foo.DeployArgs` and just be changed to `foo.DeployArgs.GetData()` and still be safe.

njhale: I like this. Added.

pkg/apis/internal.acorn.io/v1/map.go
```diff
@@ -3,25 +3,98 @@ package v1
...
+
+// Merge merges the given map into this map, returning a new map, leaving the original unchanged.
+func (g *GenericMap) Merge(from *GenericMap) *GenericMap {
+	if from == nil || from.Data == nil {
+		return g
+	}
+
+	if g == nil {
+		g = &GenericMap{}
+	}
```
g-linville, line 76: The godoc comment for this function says that `g` will remain unchanged, but this technically is a change to `g`. Not sure whether that is actually a concern or not, but I wanted to point it out just in case.

g-linville (approved):

pkg/apis/internal.acorn.io/v1/map.go
njhale: Since `g` is a copy of the original pointer, and is scoped only to this function, this doesn't change the value of g at the call site; see https://go.dev/play/p/ADokiqhZV5N

pkg/apis/internal.acorn.io/v1/map.go
```diff
@@ -3,25 +3,98 @@ package v1
...
+	if g == nil {
+		return nil
+	}
+
+	return g.Data
+}
+
+// Merge merges the given map into this map, returning a new map, leaving the original unchanged.
+func (g *GenericMap) Merge(from *GenericMap) *GenericMap {
+	if from == nil || from.Data == nil {
```
thedadams, line 70:
```suggestion
	if from.GetData() == nil {
```

pkg/apis/internal.acorn.io/v1/map.go
```diff
@@ -3,25 +3,98 @@ package v1
...
+		return g
+	}
+
+	if g == nil {
+		g = &GenericMap{}
+	}
```
thedadams, line 76: nit: not necessary, `typed.Concat` will do the right thing.

pkg/apis/internal.acorn.io/v1/map.go
```diff
@@ -3,25 +3,98 @@ package v1
...
+
+	return &GenericMap{
+		Data: typed.Concat(g.Data, from.Data),
+	}
+}
 
-func (in GenericMap) MarshalJSON() ([]byte, error) {
-	return json.Marshal((map[string]interface{})(in))
+// MarshalJSON may get called on pointers or values, so implement MarshalJSON on value.
+// http://stackoverflow.com/questions/21390979/custom-marshaljson-never-gets-called-in-go
```
thedadams, line 84: nit: It is enough to leave your comment. The stack overflow link doesn't provide enough value here.

njhale:
The integration tests are failing in a way that makes it hard to tell what the exact problem is. `TestServiceInfo` fails on line `30` with:

```
 #5 DONE 0.0s
time="2023-09-04T03:16:12Z" level=error msg="apiserver received an error that is not an metav1.Status: &errors.errorString{s:\"failed to find image information for nested acorn [nested]\"}: failed to find image information for nested acorn [nested]\n" error="<nil>"
    services_test.go:30: App.api.acorn.io "divine-moon" is invalid: spec.image: Invalid value: "2eb7bb425cf1072e11a12b3a822d29ae019f38994a15e8be54b2b8bea66eb394": failed to find image information for nested acorn [nested]
```

when attempting to run an Acorn built earlier in the same test

I know that it must be happening client-side since I'm executing the test in my branch against a local instance running `runtime:main`. Unless I'm missing something obvious, I suspect there's some issue with marshaling and/or marshaling GenericMap that silently breaks the build|run. I haven't managed to pin down exactly _where_ this is happening yet though.

integration/services/services_test.go
```diff
@@ -25,6 +28,10 @@ func TestServiceInfo(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	imageJSON := string(z.MustBe(json.MarshalIndent(image, "", "    ")))
+
+	fmt.Println("imageJSON:", imageJSON)
```
njhale, line 33: I added this to help debug an integration test failure. Will remove this before merging.

pkg/apis/internal.acorn.io/v1/map_test.go
```diff
@@ -0,0 +1,33 @@
...
+	require.NoError(t, (&gm2).UnmarshalJSON(marshaled))
+	require.Equal(t, gm.GetData(), gm2.GetData())
+
+	var gm3 GenericMap
+	marshaled, err = json.Marshal((*gm).GetData())
+	require.NoError(t, err)
+
+	require.NoError(t, (&gm3).UnmarshalJSON(marshaled))
+	require.Equal(t, gm.GetData(), gm3.GetData())
+}
```
njhale, line 33: going to clean this up a bit before merging

njhale:
This is all ready for review. Thanks for the patience folks.

thedadams (approved):

pkg/appdefinition/lookup.go
```diff
@@ -8,6 +8,18 @@ import (
...
 
+var semantic = equality.Semantic.Copy()
+
+func init() {
+	// Add custom equality functions
+	z.Must(semantic.AddFunc(func(a, b v1.AcornBuild) bool {
+		return a.OriginalImage == b.OriginalImage &&
+			a.Context == b.Context &&
+			a.Acornfile == b.Acornfile &&
+			equality.Semantic.DeepEqual(a.BuildArgs.GetData(), b.BuildArgs.GetData())
```
njhale, line 19: This was added to satisfy a preexisting requirement in the code that depends on equality between zero value of `v1.GenericMap` and `nil`; when we switched from a named map type to a nil, the default equality provided by `equality.Semantic` were no longer sufficient.

pkg/server/registry/apigroups/acorn/apps/events.go
```diff
@@ -80,7 +80,7 @@ func (s *eventRecordingStrategy) Create(ctx context.Context, obj types.Object) (
 		},
 		Type:        AppCreateEventType,
 		Severity:    apiv1.EventSeverityInfo,
-		Details:     details,
+		Details:     &details,
```
tylerslaton, line 83: Curious, what's the necessity here? Did we change how Details works?

tylerslaton (commented):
Have you tried integrating this into manager and seeing what happens?

pkg/server/registry/apigroups/acorn/apps/events.go
njhale: `Mapify` now returns a struct and all usages of `GenericMap` on the API types are now pointers.

njhale:
> Have you tried integrating this into manager and seeing what happens?

I've smoke-tested the backwards and forwards compatibility in runtime, so this should work just fine. A PR for manager using this branch is on its way.

thedadams (approved):

tylerslaton (approved):

PR #2086, opened 2023-08-17
Add a global acorn config option

njhale:
Add a global `--config-file` option to the acorn cli that allows users to
set the acorn config file to use during command execution.

e.g. Set the current project to `foo` in `./special-config.yaml`

```sh
$ acorn --config-file special-config.yaml project use foo
```

If both `--config-file` and `ACORN_CONFIG_FILE` are provided, `--config-file`
takes precedence.

### DIFF
--- a/pkg/cli/acorn.go
+++ b/pkg/cli/acorn.go
@@ -76,7 +76,7 @@ func New() *cobra.Command {
 	)
 	// This will produce an error if the project flag doesn't exist or a completion function has already
 	// been registered for this flag. Not returning the error since neither of these is likely occur.
-	if err := root.RegisterFlagCompletionFunc("project", newCompletion(cmdContext.ClientFactory, projectsCompletion).complete); err != nil {
+	if err := root.RegisterFlagCompletionFunc("project", newCompletion(cmdContext.ClientFactory, projectsCompletion(cmdContext.ClientFactory)).complete); err != nil {
 		root.Printf("Error registering completion function for -j flag: %v\n", err)
 	}
 	root.InitDefaultHelpCmd()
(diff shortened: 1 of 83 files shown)


pkg/imagesource/helper.go
```diff
@@ -25,9 +25,13 @@ type ImageSource struct {
 	// NoDefaultRegistry - if true, indicates that no container registry should be assumed for the Image.
 	// This is used if the ImageSource is for an app with auto-upgrade enabled.
 	NoDefaultRegistry bool
+
+	// acornConfig is the path to the acorn config file.
+	acornConfig string
 }
 
-func NewImageSource(file string, args, profiles, platforms []string, noDefaultReg bool) (result ImageSource) {
+func NewImageSource(acornConfig string, file string, args, profiles, platforms []string, noDefaultReg bool) (result ImageSource) {
```
njhale, line 33: something about using "" as an argument to signal default behavior doesn't sit well with me...

pkg/cli/auth.go
```diff
@@ -25,7 +25,7 @@ func getAuthForImage(ctx context.Context, clientFactory ClientFactory, image str
 		return nil, nil
 	}
 
-	creds, err := imagesource.GetCreds(c)
+	creds, err := imagesource.GetCreds(clientFactory.Options().AcornConfig, c)
```
njhale, line 28: `clientFactory.Options().AcornConfig` is pretty verbose, and we invoke it often -- should I just add a convenience method to `ClientFactory`? e.g. `clientFactory.AcornConfig()`

pkg/cli/acorn.go
```diff
@@ -84,10 +84,11 @@ func New() *cobra.Command {
 }
 
 type Acorn struct {
-	Kubeconfig string `usage:"Explicitly use kubeconfig file, overriding the default context" env:"ACORN_KUBECONFIG"`
-	Project    string `usage:"Project to work in" short:"j" env:"ACORN_PROJECT"`
-	Debug      bool   `usage:"Enable debug logging" env:"ACORN_DEBUG"`
-	DebugLevel int    `usage:"Debug log level (valid 0-9) (default 7)" env:"ACORN_DEBUG_LEVEL"`
+	AcornConfig string `usage:"Path of the acorn config file to use" name:"config" env:"ACORN_CONFIG_FILE"`
```
njhale, line 87: Is AcornConfig a reasonable name? maybeCLIConfig works better with our existing naming scheme?

g-linville:
I'm a little confused as to why we want this. I can't come up with a reason why I would want to specify a different acorn config file.

njhale:
@g-linville 

> I'm a little confused as to why we want this. I can't come up with a reason why I would want to specify a different acorn config file.

It's useful when you're working in separate contexts concurrently -- e.g. more than one acorn, where each gets pushed to repos in the same registry that require different authz -- and you don't want to change the shared config.

We also _already_ consume an `ACORN_CONFIG_FILE` environment variable to this end, so adding this option provides parity (like we have with `ACORN_PROJECT` and `-j`).

thedadams (approved):

Issue #2081, opened 2023-08-16
Builders should act like they are panicking when shutting down


thedadams:
The builders use `logrus.Fatal` every time they shutdown. This produces a false-positive for issues.

It would be better if a stack-trace wasn't produced on a graceful shutdown to avoid this confusion.

thedadams:
The easiest way I found to test this is the following:
1. Run `runtime` locally with `acorn install`.
2. Run a build so that the `buildkitd` deployment is created.
3. In two separate terminals watch the logs of both containers in the pod: `kubectl -n acorn-image-system logs buildkitd-7cd9fd44c6-wwgxc -f` and `kubectl -n acorn-image-system logs buildkitd-7cd9fd44c6-wwgxc -c service -f`
4. In another terminal, restart the deployment: `kubectl -n acorn-image-system rollout restart deployment buildkitd`

Once the fix is in, there should be no stack traces printed when the containers exit.

sangee2004:
Able to reproduce the issue on seeing stack traces printed in the logs when testing with `acorn version v0.8.0-rc4-27-g20d5e79e+20d5e79e` by following the steps mentioned in the issues.


Tested with `acorn version v0.8.0-rc4-29-g6a019fe7+6a019fe7`
When following the same steps mentioned in the issue , there is no stack traces printed in the logs.

Following are the logs that were printed when attempting `kubectl -n acorn-image-system rollout restart deployment buildkitd`
 
`kubectl -n acorn-image-system logs buildkitd-686c95c556-p27f6 -f `
```
time="2023-08-17T17:50:34Z" level=info msg="stopping server"
buildkitd: context canceled
rpc error: code = NotFound desc = an error occurred when try to find container "402a5b067b4b4bbce4f76e0aeb7103136b9a1e7170955d8a80d0d547f39a618a": not found%                 
```

kubectl -n acorn-image-system logs buildkitd-686c95c556-p27f6 -c service -f 
```
time="2023-08-17T17:50:34Z" level=fatal msg="accept tcp [::]:5000: use of closed network connection"
rpc error: code = NotFound desc = an error occurred when try to find container "4da741bf1818c18ca92d779583dd39b76f07bd12c5ec4f11de85f6766702a5a7": not found%                 
```

PR #2087, opened 2023-08-17
Stop printing stack traces when buildkitd pod is deleted (#2081)

thedadams:
There are two fixes here related to the deletion of the buildkitd pod:
1. A SIGQUIT was being sent to the pod, which was causing both containers to print their stack traces. This is fixed in the Dockerfile
2. The service container was not exiting gracefully because it wasn't paying attention to the context. Now it will watch the context and exit when it closes.

### Checklist

Issue: https://github.com/acorn-io/runtime/issues/2081

### DIFF
--- a/pkg/cli/buildserver.go
+++ b/pkg/cli/buildserver.go
@@ -3,6 +3,7 @@ package cli
 import (
 	"fmt"
 	"net/http"
+	"os"
 
 	"github.com/acorn-io/runtime/pkg/buildserver"
 	cli "github.com/acorn-io/runtime/pkg/cli/builder"
@@ -33,7 +34,7 @@ type BuildServer struct {
 	ForwardService string `usage:"Forwarding Address" env:"ACORN_BUILD_SERVER_FORWARD_SERVICE"`
 }
 
-func (s *BuildServer) Run(cmd *cobra.Command, args []string) error {
+func (s *BuildServer) Run(cmd *cobra.Command, _ []string) error {
 	c, err := k8sclient.Default()
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


tylerslaton (approved):

cjellick (approved):

pkg/cli/buildserver.go
```diff
@@ -47,15 +48,27 @@ func (s *BuildServer) Run(cmd *cobra.Command, args []string) error {
...
+	go func() {
+		// Nothing here is using the context. Therefore, when the context is canceled, we should just exit.
+		<-cmd.Context().Done()
+		if p != nil {
+			if err = p.Close(); err != nil {
+				logrus.Errorf("Error closing proxy: %v", err)
+			}
+		}
+		os.Exit(0)
+	}()
```
njhale, line 70: I don't think we need need to use a pointer for `p` since both closures capture it in scope:

```go
	var p tcpproxy.Proxy
	if s.ForwardService != "" {
		p.AddRoute(fmt.Sprintf(":%d", s.ForwardPort), tcpproxy.To(s.ForwardService))
		go func() {
			logrus.Infof("Forwarding :%d to %s", s.ForwardPort, s.ForwardService)
			logrus.Fatal(p.Run())
		}()
	}

	go func() {
		// Nothing here is using the context. Therefore, when the context is canceled, we should just exit.
		<-cmd.Context().Done()
	    if err = p.Close(); err != nil {
			logrus.Errorf("Error closing proxy: %v", err)
		}

		os.Exit(0)
	}()
```

njhale (approved):
Left a suggestion, but IMO this is g2g as is

PR #2095, opened 2023-08-18
Add Cilium proxy visibility annotations

g-linville:
This adds back the Cilium proxy-visibility annotations. This time it should work properly.

Note that I had to adjust NetworkPolicies to get this to work. For HTTP probes (readiness, helath, etc.), I have to open up that port to the world in order for the probe to pass, otherwise Cilium will block it. We can improve this in the future by using CiliumNetworkPolicies, which are more fine-grained and will enable us to specify that only traffic coming from the host should be allowed to the probe port, rather than from anywhere. Until then, this is a downgrade of our NetworkPolicies. If that is a blocker for this PR, I can get to work on the CNPs before we merge this.

### Checklist

### DIFF
--- a/pkg/controller/appdefinition/deploy.go
+++ b/pkg/controller/appdefinition/deploy.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/acorn-io/baaah/pkg/router"
 	"github.com/acorn-io/baaah/pkg/typed"
+	apiv1 "github.com/acorn-io/runtime/pkg/apis/api.acorn.io/v1"
 	v1 "github.com/acorn-io/runtime/pkg/apis/internal.acorn.io/v1"
 	"github.com/acorn-io/runtime/pkg/appdefinition"
 	"github.com/acorn-io/runtime/pkg/condition"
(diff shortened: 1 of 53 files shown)


pkg/controller/appdefinition/deploy.go
```diff
@@ -553,6 +554,18 @@ func addPrometheusAnnotations(annotations map[string]string, container v1.Contai
...
 
+func addCiliumAnnotations(annotations map[string]string, container v1.Container) {
+	var visibilityConfigs []string
+	for _, port := range container.Ports {
+		if port.Protocol == v1.ProtocolHTTP {
+			visibilityConfigs = append(visibilityConfigs, fmt.Sprintf("<Ingress/%d/TCP/HTTP>", port.TargetPort))
+		}
+	}
+	if len(visibilityConfigs) > 0 {
+		annotations[labels.CiliumProxyVisibility] = strings.Join(visibilityConfigs, ",")
```
StrongMonkey, line 565: What is this annotion doing?

thedadams (approved):

tylerslaton (approved):

pkg/controller/appdefinition/deploy.go
g-linville: These annotations tell Cilium to pay attention to this traffic and route it through the layer 7 proxy in order to generate metrics for it. It doesn't do this by default.

g-linville:
Re-requesting review because I added the `--use-cilium-layer-7` parameter.

pkg/controller/appdefinition/deploy.go
```diff
@@ -553,6 +555,20 @@ func addPrometheusAnnotations(annotations map[string]string, container v1.Contai
 	}
 }
 
+func addCiliumAnnotations(annotations map[string]string, container v1.Container, cfg *apiv1.Config) {
+	if cfg.UseCiliumLayer7 != nil && *cfg.UseCiliumLayer7 {
```
thedadams, line 559: Most of our config stuff is setup so that you don't have to check nil-ness. Is that the case here?

Regardless, can you use `z.Dereference`?

tylerslaton (approved):
Looks like you just need to generate the test output to get the tests to pass.

thedadams (approved):

PR #2105, opened 2023-08-21
Depot integration

ibuildthecloud:
This is hack to leverage depot.dev for building. This will change in
the future and this is only a quick experiment.


cjellick:
Steps to use this:
1. Get a depot.dev account and proejct
2. Configure acorn to use an external registry by setting the install flag `--internal-registry-prefix` to your external registry of choice
3. Create a secret in your acorn proejct: 1acorn -j dev secret create depot-builder-key --data projectId=${PROJECT_ID},token=${TOKEN}`

PR #2109, opened 2023-08-21
Add the ability to specify unlimited for individual resources

tylerslaton:
This adds the ability for `Unlimited` to be more granularly set per resource.

### Checklist

### DIFF
--- a/pkg/apis/internal.admin.acorn.io/v1/resources.go
+++ b/pkg/apis/internal.admin.acorn.io/v1/resources.go
@@ -20,8 +20,6 @@ func UnlimitedQuantity() resource.Quantity {
 // external controllers to programmatically set the resources easier. Calls to
 // its functions are mutating.
 type Resources struct {
-	Unlimited bool `json:"unlimited,omitempty"`
-
 	Apps       int `json:"apps,omitempty"`
 	Containers int `json:"containers,omitempty"`
 	Jobs       int `json:"jobs,omitempty"`
@@ -127,10 +125,6 @@ func (current *Resources) Remove(incoming Resources, all bool) {
 // an aggregated error will be returned with all exceeded resources.
 // If the current resources defines unlimited, then it will always fit.
 func (current *Resources) Fits(incoming Resources) error {
-	if current.Unlimited {
-		return nil
-	}
-
 	exceededResources := []string{}
 
 	// Define function for checking int resources to keep code DRY
@@ -204,8 +198,7 @@ func (current *Resources) NonEmptyString() string {
 // Equals will check if the current Resources struct is equal to another. This is useful
 // to avoid needing to do a deep equal on the entire struct.
 func (current *Resources) Equals(incoming Resources) bool {
-	return current.Unlimited == incoming.Unlimited &&
-		current.Apps == incoming.Apps &&
+	return current.Apps == incoming.Apps &&
 		current.Containers == incoming.Containers &&
 		current.Jobs == incoming.Jobs &&
 		current.Volumes == incoming.Volumes &&
--- a/pkg/apis/internal.admin.acorn.io/v1/resources.go
+++ b/pkg/apis/internal.admin.acorn.io/v1/resources.go
@@ -9,6 +9,13 @@ import (
 
 var ErrExceededResources = fmt.Errorf("quota would be exceeded for resources")
 
+const Unlimited = -1
+
+// NewUnlimitedQuantity creates a Quantity with an Unlimited value
+func UnlimitedQuantity() resource.Quantity {
+	return *resource.NewQuantity(Unlimited, resource.DecimalSI)
+}
+
 // Resources is a struct separate from the QuotaRequestInstanceSpec to allow for
 // external controllers to programmatically set the resources easier. Calls to
 // its functions are mutating.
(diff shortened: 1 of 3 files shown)


pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -84,14 +91,14 @@ func (current *Resources) Fits(incoming Resources) error {
...
-		if currentVal < incomingVal {
+		if currentVal != Unlimited && currentVal < incomingVal {
 			exceededResources = append(exceededResources, resource)
 		}
 	}
 
 	// Define function for checking quantity resources to keep code DRY
 	checkQuantityResource := func(resource string, currentVal, incomingVal resource.Quantity) {
-		if currentVal.Cmp(incomingVal) < 0 {
+		if !currentVal.Equal(UnlimitedQuantity()) && currentVal.Cmp(incomingVal) < 0 {
```
thedadams, line 101: Personally, I would like to find a way to get rid of `UnlimiteQuantity`. Can `CmpInt64(Unlimited)` be used here instead?

tylerslaton: Sure, I can do that.

Do you want to get rid of the `UnlimitedQuantity` function entirely? I mainly made it to make setting unlimited quantities more straight forward. Without it there would be a lot of `resource.NewQuantity(Unlimited,  resource.DecimalSI)` set. Is that something we'd want as it is more explicit?

thedadams: No, I think it's fine then. Thanks for the explanation.

thedadams (approved):

keyallis (approved):

tylerslaton:
@keyallis @thedadams There was an edge case the original code you reviewed did not account for. I put it into a separate commit with some unit tests. Please let me know what you think.

pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -30,44 +37,78 @@ type Resources struct {
...
 func (current *Resources) Remove(incoming Resources, all bool) {
-	// Do not allow resources to go below 0
-	nonNegativeSubtract := func(currentVal, incomingVal int) int {
-		difference := currentVal - incomingVal
+	sub := func(c, i int) int {
+		if c == Unlimited && i == Unlimited {
+			return 0
+		} else if c == Unlimited || i == Unlimited {
+			return c
+		}
```
keyallis, line 75: I think your current code has a bug if `c` is some value and `i` is `Unlimited`.

Cases:
1. `c unlimited i unlimited -> 0`
1. `c unlimited i value -> unlimited`
1. `c value i unlimited -> 0`
1. `c value i value -> fallthrough case`

Here's the logic I came up with:
```
if i == Unlimited {
   return 0
} else if c == Unlimited {
   return c
}
```

pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -30,44 +37,78 @@ type Resources struct {
...
-	current.Jobs = nonNegativeSubtract(current.Jobs, incoming.Jobs)
-	current.Volumes = nonNegativeSubtract(current.Volumes, incoming.Volumes)
-	current.Images = nonNegativeSubtract(current.Images, incoming.Images)
-	current.Projects = nonNegativeSubtract(current.Projects, incoming.Projects)
+	subQuantity := func(c, i resource.Quantity) resource.Quantity {
+		if c.Equal(UnlimitedQuantity()) && i.Equal(UnlimitedQuantity()) {
+			return *resource.NewQuantity(0, c.Format)
+		} else if c.Equal(UnlimitedQuantity()) || i.Equal(UnlimitedQuantity()) {
+			return c
+		}
```
keyallis, line 89: Same comment as above

pkg/apis/internal.admin.acorn.io/v1/resources_test.go
```diff
@@ -83,44 +129,94 @@ func TestRemove(t *testing.T) {
...
+			},
+			incoming: Resources{
+				Apps:   Unlimited,
+				Memory: UnlimitedQuantity(),
+			},
+			expected: Resources{
+				Apps:   1,
+				Memory: resource.MustParse("1Mi"),
+			},
+		},
```
keyallis, line 206: This is the test case where we would want to determine if the expected should be zeroed out or just left as current

keyallis (approved):

pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -9,6 +9,13 @@ import (
 
 var ErrExceededResources = fmt.Errorf("quota would be exceeded for resources")
 
+const Unlimited = -1
```
njhale, line 12: Does this need to be exported? Is it used elsewhere? I'm assuming the answer is "yes", but figured I'd ask anyway.

njhale (approved):
One question, but not blocking.

LGTM

pkg/apis/internal.admin.acorn.io/v1/resources.go
tylerslaton: Yep, it will be used elsewhere, namely in Manager.

pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -30,44 +37,75 @@ type Resources struct {
...
-// Remove will remove the resources of another Resources struct from the current one.
+// Remove will remove the resources of another Resources struct from the current one. Calling remove
+// will be a no-op for any resource values that are set to unlimited.
 func (current *Resources) Remove(incoming Resources, all bool) {
-	// Do not allow resources to go below 0
-	nonNegativeSubtract := func(currentVal, incomingVal int) int {
-		difference := currentVal - incomingVal
+	sub := func(c, i int) int {
+		if c == Unlimited || i == Unlimited {
+			return c
```
thedadams, line 73: Should this be
```suggestion
			return Unlimited
```

pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -30,44 +37,75 @@ type Resources struct {
...
 
-	current.Apps = nonNegativeSubtract(current.Apps, incoming.Apps)
-	current.Containers = nonNegativeSubtract(current.Containers, incoming.Containers)
-	current.Jobs = nonNegativeSubtract(current.Jobs, incoming.Jobs)
-	current.Volumes = nonNegativeSubtract(current.Volumes, incoming.Volumes)
-	current.Images = nonNegativeSubtract(current.Images, incoming.Images)
-	current.Projects = nonNegativeSubtract(current.Projects, incoming.Projects)
+	subQuantity := func(c, i resource.Quantity) resource.Quantity {
+		if c.Equal(UnlimitedQuantity()) || i.Equal(UnlimitedQuantity()) {
+			return c
```
thedadams, line 85: Should this be
```suggestion
			return UnlimitedQuantity()
```

pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -30,44 +37,75 @@ type Resources struct {
...
+	add := func(c, i int) int {
+		if c == Unlimited || i == Unlimited {
+			return Unlimited
+		}
+		return c + i
+	}
+
+	addQuantity := func(c, i resource.Quantity) resource.Quantity {
+		if c.Equal(UnlimitedQuantity()) || i.Equal(UnlimitedQuantity()) {
+			return UnlimitedQuantity()
```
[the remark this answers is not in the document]
tylerslaton: You're right, updating this code now.

pkg/apis/internal.admin.acorn.io/v1/resources.go
thedadams: @tylerslaton Just wanted to double-check whether this is correct. Same for the below.

tylerslaton: Talked about this offline and left a comment to reflect why this code exists and what it is doing.

pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -30,44 +35,88 @@ type Resources struct {
...
-	nonNegativeSubtract := func(currentVal, incomingVal int) int {
-		difference := currentVal - incomingVal
+	sub := func(c, i int) int {
+		// We don't expect this situation to happen. This is because there should not be a situation
+		// where we are removing from or with unlimited resources. However if it does, we want to
+		// be careful and handle it. With that in mind the logic here is as follows:
+		//
+		// 1. If the current value is unlimited, then removing a non-unlimited value should not change
+		//    the current value.
+		// 2. If the current value is not unlimited, then removing a non-unlimited value should not
```
thedadams, line 77: nit:
```suggestion
		// 2. If the current value is not unlimited, then removing an unlimited value should not
```

thedadams (approved):

PR #2114, opened 2023-08-23
Add install opt to set annotations on acorn-api pods

njhale:
Enable custom annotations to be set on the acorn-api deployment by adding an `--api-server-pod-annotations` option to the `acorn install` subcommand

e.g.

```sh
$ acorn install --api-server-pod-annotations foo=bar
```

### DIFF
--- a/docs/docs/100-reference/01-command-line/acorn_install.md
+++ b/docs/docs/100-reference/01-command-line/acorn_install.md
@@ -27,6 +27,7 @@ acorn install
       --allow-user-metadata-namespace strings           Allow these namespaces to propagate labels and annotations to dependent objects, no effect if --ignore-user-labels-and-annotations not true
       --api-server-cpu string                           The CPU to allocate to the runtime-api-server in the format of <req>:<limit> (example 200m:1000m)
       --api-server-memory string                        The memory to allocate to the runtime-api-server in the format of <req>:<limit> (example 256Mi:1Gi)
+      --api-server-pod-annotations stringArray          annotations to apply to acorn-api pods
       --api-server-replicas int                         acorn-api deployment replica count
       --auto-upgrade-interval string                    For apps configured with automatic upgrades enabled, the interval at which to check for new versions. Upgrade intervals configured at the application level cannot be smaller than this. (default '5m' - 5 minutes)
       --aws-identity-provider-arn string                ARN of cluster's OpenID Connect provider registered in AWS
--- a/pkg/cli/install.go
+++ b/pkg/cli/install.go
@@ -35,6 +35,7 @@ type Install struct {
 	Output string `usage:"Output manifests instead of applying them (json, yaml)" short:"o"`
 
 	APIServerReplicas                  *int     `usage:"acorn-api deployment replica count" name:"api-server-replicas"`
+	APIServerPodAnnotations            []string `usage:"annotations to apply to acorn-api pods" name:"api-server-pod-annotations" split:"false"`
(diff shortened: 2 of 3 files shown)


pkg/cli/install.go
```diff
@@ -107,3 +110,15 @@ func (i *Install) Run(cmd *cobra.Command, args []string) error {
 
 	return install.Install(cmd.Context(), image, opts)
 }
+
+func parseAnnotations(annotations []string) (map[string]string, error) {
+	result := map[string]string{}
```
iwilltry42, line 115: Nit:
```suggestion
	result := make(map[string]string, len(annotations))
```

docs/docs/100-reference/01-command-line/acorn_install.md
```diff
@@ -27,6 +27,7 @@ acorn install
       --allow-user-metadata-namespace strings           Allow these namespaces to propagate labels and annotations to dependent objects, no effect if --ignore-user-labels-and-annotations not true
       --api-server-cpu string                           The CPU to allocate to the runtime-api-server in the format of <req>:<limit> (example 200m:1000m)
       --api-server-memory string                        The memory to allocate to the runtime-api-server in the format of <req>:<limit> (example 256Mi:1Gi)
+      --api-server-pod-annotations stringArray          annotations to apply to acorn-api pods
```
iwilltry42, line 30: Is there any chance that we'll need this for the controller pod as well soon?

pkg/cli/install.go
```diff
@@ -83,22 +84,24 @@ func (i *Install) Run(cmd *cobra.Command, args []string) error {
...
 	}
 
-	annotations := map[string]string{}
-	for _, anno := range i.ControllerServiceAccountAnnotation {
-		k, v, ok := strings.Cut(anno, "=")
-		if !ok {
-			return fmt.Errorf("--controller-service-account-annotation must be in key=value format got [%s]", anno)
-		}
-		annotations[k] = v
+	controllerAnnotations, err := parseAnnotations(i.ControllerServiceAccountAnnotation)
```
iwilltry42, line 87: Maybe `controllerSAAnnotations` for clarity?

pkg/install/install.go
```diff
@@ -597,6 +603,29 @@ func Deployments(runtimeImage string, apiServerReplicas, controllerReplicas int,
...
 
+func replacePodAnnotations(annotations map[string]string, objs []kclient.Object) ([]kclient.Object, error) {
+	val := map[string]any{}
+	for k, v := range annotations {
+		val[k] = v
+	}
+
+	if len(val) == 0 {
+		return objs, nil
+	}
```
iwilltry42, line 614:
```suggestion
	if len(annotations) == 0 {
		return objs, nil
	}
	
	val := make(map[string]any, len(annotations))
	for k, v := range annotations {
		val[k] = v
	}
```

iwilltry42 (approved):
Some comments, but nothing blocking :+1:

tylerslaton (approved):

pkg/cli/install.go
njhale: done

docs/docs/100-reference/01-command-line/acorn_install.md
njhale: Maybe -- I think we should add it when we need it though.

pkg/cli/install.go
njhale: done

g-linville (approved):
+1 to all of Thorsten's comments, approved

Issue #2119, opened 2023-08-24
Improve error message to indicate that local credentials needed for `acorn image copy` 


sangee2004:
acorn version  - v0.8.0-rc4-58-gbe49d0bd+be49d0bd

**Steps to reproduce the problem:**
1. Attempt `acorn image copy` between 2 remote registries 

2. Have login credentials created for `docker.io` and` ghcr.io` using `acorn login`
3. Copy image between remote repositories using `acorn image cp  docker.io/sangeetha/myfirstserviceagain:v0.0.1   ghcr.io/sangee2004/myfirstservice:v0.0.1`
This fails with following error message:
```
  ✗  ERROR:  GET https://ghcr.io/token?scope=repository%3Asangee2004%2Fmyfirstserviceagain%3Apull&service=ghcr.io: DENIED: requested access to the resource is denied
```
**Expected Behavior:**
`acorn image copy` succeeds only when we have local credentials made available using `acorn login -l` command.

Error message presented to the user should provide this information.

g-linville:
I improved the help and error messages for `acorn copy`.

sangee2004:
acorn version - `v0.8.0-rc5+15fc5478`

Following error message presented when `acorn image copy` is attempted when no local creds are present:
```
Registry authorization failed. Ensure that you have the correct permissions to push to this registry. Run 'acorn login -l <registry>' if you have not logged in yet.
```


Following text seen in `acorn image copy --help`
````

  This command copies Acorn images between remote image registries.
  It does not interact with images stored in the Acorn internal registry, or with the Acorn API in any way.
  To set up credentials for a registry, use 'acorn login -l <registry>'. It only works with locally stored credentials.

```

PR #2120, opened 2023-08-24
Improve help and error messages for `acorn copy` (#2119)

g-linville:
for #2119

This improves the error messages for `acorn copy` when authn/authz fails. It also returns an error when the user tries to use `-a` with the `acorn copy <image> <new tag>` syntax, since that is undefined behavior.

I also made it clear in the help message that the user needs to log in to registries with `acorn login -l` since this command does not interact with the Acorn API.

I also improved the tag validation for `acorn copy <image> <new tag>` syntax.

### Checklist

### DIFF
--- a/pkg/cli/copy.go
+++ b/pkg/cli/copy.go
@@ -276,6 +276,12 @@ func (a *ImageCopy) copyTag(source name.Reference, newTag string, sourceOpts []r
 
 	dest := source.Context().Tag(newTag)
 
+	// Parse it again to make sure that the tag provided by the user is valid
+	_, err = name.ParseReference(dest.String())
+	if err != nil {
+		return err
+	}
+
 	if !a.Force {
 		if err := errIfImageExistsAndIsDifferent(sourceIndex, dest, sourceOpts); err != nil {
 			return err
--- a/pkg/cli/copy.go
+++ b/pkg/cli/copy.go
@@ -25,7 +25,8 @@ func NewImageCopy(c CommandContext) *cobra.Command {
 		Use: `copy [flags] SOURCE DESTINATION
 
   This command copies Acorn images between remote image registries.
-  It does not interact with images stored in the Acorn internal registry, or with the Acorn API in any way.`,
+  It does not interact with images stored in the Acorn internal registry, or with the Acorn API in any way.
+  To set up credentials for a registry, use 'acorn login -l <registry>'. It only works with locally stored credentials.`,
 		Aliases:           []string{"cp"},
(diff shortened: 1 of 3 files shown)


pkg/cli/copy.go
```diff
@@ -248,13 +264,24 @@ func (a *ImageCopy) copyRepo(args []string, sourceOpts, destOpts []remote.Option
...
 	sourceIndex, err := remote.Index(source, sourceOpts...)
 	if err != nil {
 		return err
 	}
 
 	dest := source.Context().Tag(newTag)
 
+	// Parse it again to make sure that the tag provided by the user is valid
+	_, err = name.ParseReference(dest.String())
+	if err != nil {
```
njhale, line 281: nit: you can condense this to a single line

```suggestion
	if _, err = name.ParseReference(dest.String()); err != nil {
```

pkg/cli/copy.go
```diff
@@ -48,7 +50,21 @@ type ImageCopy struct {
...
+func (a *ImageCopy) Run(cmd *cobra.Command, args []string) (err error) {
+	// Print a helpful error message for the user if they end up getting an authentication error
+	defer func() {
+		if err != nil {
+			var terr *transport.Error
+			if ok := errors.As(err, &terr); ok {
+				if terr.StatusCode == http.StatusForbidden {
+					logrus.Warnf("Registry authentication failed. Try running 'acorn login -l <registry>'")
+				} else if terr.StatusCode == http.StatusUnauthorized {
+					logrus.Warnf("Registry authorization failed. Ensure that you have the correct permissions to push to this registry. Run 'acorn login -l <registry>' if you have not logged in yet.")
```
njhale, line 62: Two questions about using `logrus.Warnf` here:
- Does it print to stderr or stdout?
- Does it print out extra logrus stuff like timestamp?

I mainly ask because, for user facing error messages, I think we want to to stderr and we probably don't want timestamps.

njhale (commented):
Looks good. Left two questions/comments for you @g-linville

pkg/cli/copy.go
g-linville: Good call. I don't know if it was printing to stderr or stdout, and you're right that the log-like formatting is probably not what we want to show to the user.

I replaced both of these `logrus.Warnf` calls with `fmt.Fprintln`, which allows me to plainly write to stderr.

tylerslaton (approved):

keyallis (approved):

njhale (approved):

Issue #2130, opened 2023-08-28
Not able to list projects created in 0.7.1 after upgrade to 0.8.0-rc5.


sangee2004:
**Version** -  `0.7.1` upgraded to `0.8.0-rc5`

**Steps to reproduce the problem:**
1. Install acorn version `v0.7.1+df5bf694`
Create a project - (`hello1` in my case.
I built and pushed an image from this project.

2. . Install acorn version  `v0.8.0-rc5+15fc5478`

Not able to list or use project `hello1`
```
 acorn project           
NAME      DEFAULT   REGIONS
acorn     *         local
sangeethahariharan@Sangeethas-MacBook-Pro example1 % acorn project use hello1
  ✗  ERROR:  failed to find project hello1, use "acorn projects" to list valid project names: projects.api.acorn.io "hello1" not found
```

```
 kubectl describe ns hello1 
Name:         hello1
Labels:       acorn.io/project=true
              kubernetes.io/metadata.name=hello1
Annotations:  acorn.io/calculated-project-default-region: local
              acorn.io/project-default-region: 
              acorn.io/project-supported-regions: 
Status:       Active

No resource quota.

No LimitRange resource.
sangeethahariharan@Sangeethas-MacBook-Pro example1 % kubectl describe ns acorn  
Name:         acorn
Labels:       acorn.io/managed=true
              acorn.io/project=true
              apply.acorn.io/hash=5b51b6b4a16a4c00326788cbc9078b97da1676fd
              kubernetes.io/metadata.name=acorn
Annotations:  apply.acorn.io/applied:
                H4sIAAAAAAAA/3xRQQ7TMBD8CtpzUpImcRpL3LggDiCEuCAOa3vbmjrryHYKqMrfkVMqSqHcvDvj0czsBUZKaDAhyAsgs0+YrOe4jtPkfmxQ+8Ab61/6b0yhPJxPIMFyosDofqPnun...
              apply.acorn.io/owner-gvk: internal.acorn.io/v1, Kind=ProjectInstance
              apply.acorn.io/owner-name: acorn
              apply.acorn.io/owner-namespace: 
              apply.acorn.io/owner-sub-context: acorn-controller
Status:       Active

No resource quota.

No LimitRange resource.
```

Note - Able to list and work with `acorn` project.

sangee2004:
Tested with acorn version - `v0.8.0-rc5-8-g4fb6453a`

1. Install acorn version`v0.7.1+df5bf694`
[...]

PR #2132, opened 2023-08-28
Migrate legacy project namespaces (#2130)

thedadams:
If a project namespace existing prior to upgrading to the latest version, then the namespace needs to be migrated to a new project instance. This change does this by using a proper label selector to find namespaces that need to be migrated.

### Checklist

Issue: https://github.com/acorn-io/runtime/issues/2130

### DIFF
--- a/pkg/server/registry/apigroups/acorn/projects/storage.go
+++ b/pkg/server/registry/apigroups/acorn/projects/storage.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/acorn-io/baaah/pkg/router"
 	"github.com/acorn-io/mink/pkg/stores"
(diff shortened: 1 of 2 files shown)


pkg/server/registry/apigroups/acorn/projects/storage.go
```diff
@@ -77,3 +84,40 @@ func (pr *projectCreater) Create(ctx context.Context, object types.Object) (type
...
+	// Once a project namespace is managed by the project instance, then this label will be set on it.
+	// Therefore, using this label selector ensures the migration only happens once for a namespace.
+	notManaged, err := klabels.NewRequirement(labels.AcornManaged, selection.DoesNotExist, nil)
+	if err != nil {
+		return err
+	}
+
+	nsSelector := klabels.SelectorFromSet(map[string]string{
+		labels.AcornProject: "true",
+	}).Add(*notManaged)
```
njhale, line 98: nit: can these be built once and memoized? or are they not threadsafe?

pkg/server/registry/apigroups/acorn/projects/storage.go
```diff
@@ -77,3 +84,40 @@ func (pr *projectCreater) Create(ctx context.Context, object types.Object) (type
...
+			},
+			Spec: v1.ProjectInstanceSpec{
+				DefaultRegion: ns.Annotations[labels.AcornProjectDefaultRegion],
+			},
+		}
+
+		if supportedRegions := ns.Annotations[labels.AcornProjectSupportedRegions]; supportedRegions != "" {
+			project.Spec.SupportedRegions = strings.Split(supportedRegions, ",")
+		}
+		if err = c.Create(ctx, project); err != nil && !apierrors.IsAlreadyExists(err) {
```
njhale, line 117: If the project store isn't up and running until after `migrateLegacyNamespaces` returns, how can this client talk to the Project API? Does it hit the "old" API of the pod upgraded from? If that's the case, it doesn't seem like this would work as a net-new deployment.

pkg/server/registry/apigroups/acorn/projects/storage.go
thedadams: They are built at startup and then not needed again.

pkg/server/registry/apigroups/acorn/projects/storage.go
thedadams: This create call is hitting using a `ProjectInstance`, which is stored in Kubernetes.

thedadams: Confirmed I am able to run `acorn install` with these changes on a fresh Kubernetes cluster.

pkg/server/registry/apigroups/acorn/projects/storage.go
njhale: wow, yup, carry on.

cjellick (approved):

njhale (approved):

PR #2138, opened 2023-08-30
Add --quiet to install

cjellick:
This will suppress all installation output except for errors
encountered. The motivation is to use this flag when manager installs
acorn because these interactive terminal messages aren't useful in that
scenario.

### Checklist

### DIFF
--- a/pkg/cli/install.go
+++ b/pkg/cli/install.go
@@ -29,10 +29,10 @@ acorn install`,
 }
 
 type Install struct {
-	SkipChecks bool `usage:"Bypass installation checks"`
-
-	Image  string `usage:"Override the default image used for the deployment"`
-	Output string `usage:"Output manifests instead of applying them (json, yaml)" short:"o"`
+	SkipChecks bool   `usage:"Bypass installation checks"`
+	Quiet      bool   `usage:"Only output errors encountered during installation"`
+	Image      string `usage:"Override the default image used for the deployment"`
+	Output     string `usage:"Output manifests instead of applying them (json, yaml)" short:"o"`
 
 	APIServerReplicas                  *int     `usage:"acorn-api deployment replica count" name:"api-server-replicas"`
 	APIServerPodAnnotations            []string `usage:"annotations to apply to acorn-api pods" name:"api-server-pod-annotations" split:"false"`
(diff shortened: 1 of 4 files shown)


pkg/cli/install.go
```diff
@@ -29,10 +29,10 @@ acorn install`,
 }
 
 type Install struct {
-	SkipChecks bool `usage:"Bypass installation checks"`
-
-	Image  string `usage:"Override the default image used for the deployment"`
-	Output string `usage:"Output manifests instead of applying them (json, yaml)" short:"o"`
+	SkipChecks bool   `usage:"Bypass installation checks"`
+	Quiet      bool   `usage:"Only output errors encountered during installation"`
```
njhale, line 33: nit: is there a reason not to add `-q` as a shorthand?

```suggestion
	Quiet      bool   `usage:"Only output errors encountered during installation" short:"q"`
```

pkg/term/quiet.go
```diff
@@ -0,0 +1,42 @@
...
+	// No-op, we're being quiet
+}
+
+func (s *quietSpinner) SuccessWithWarning(format string, v ...any) {
+	// No-op, we're being quiet
+}
+
+func (s *quietSpinner) Success() {
+	// No-op, we're being quiet
+}
```
njhale, line 42: nit:

```suggestion
func (*quietSpinner) Infof(string, ...any) {
	// No-op, we're being quiet
}

func (*quietSpinner) SuccessWithWarning(string, ...any) {
	// No-op, we're being quiet
}

func (*quietSpinner) Success() {
	// No-op, we're being quiet
}
```

njhale (approved):
left some suggestions, but I'm cool with this going in as is too.

pkg/cli/install.go
cjellick: After you called this out, I looked at the rest of the install flags and most don't have shorthands - only three do: help, output, and `-m, --workload-memory-default`.

`help` and `output` make sense because they're kind of meta, but given that the rest don't have shorthands, I don't want to introduce it for this. I'd rather consider doing it holistically in the future. `-m` kind of feels like the odd man out at this point

pkg/term/quiet.go
cjellick: took this suggestion, but locally and then force pushed

PR #2141, opened 2023-08-31
Quiet more install output

cjellick:

### Checklist

### DIFF
--- a/pkg/install/install.go
+++ b/pkg/install/install.go
@@ -162,7 +162,7 @@ func Install(ctx context.Context, image string, opts *Options) error {
 			}
 			opts.Config.LetsEncryptEmail = result
 		}
-		pterm.Info.Println("You've enabled automatic TLS certificate provisioning with Let's Encrypt. This can take a few minutes to configure.")
+		info(opts, "You've enabled automatic TLS certificate provisioning with Let's Encrypt. This can take a few minutes to configure.")
 	}
 
 	// Validate E-Mail address provided for Let's Encrypt registration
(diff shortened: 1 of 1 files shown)


pkg/install/install.go
```diff
@@ -304,6 +304,12 @@ func Install(ctx context.Context, image string, opts *Options) error {
 	return nil
 }
 
+func info(opts *Options, a ...any) {
+	if !opts.Quiet {
+		pterm.Info.Println(a...)
+	}
+}
```
njhale, line 311: If we're not opposed to some global state, we could avoid passing `opts` around everywhere by refactoring this to be more like a default logger

```go
type quietPrinter struct {
    quiet bool
}

func (q quietPrinter) Info(a ...any) {
    if !q.quiet {
		pterm.Info.Println(a...)
	}
}

// At package scope
var qterm quietPrinter

// ...
func (o *Options) complete() *Options {
// ...
    qterm.Quiet = o.Quiet // Not thread-safe, but you easily could; e.g. qterm.SetQuiet(o.Quiet) might encapsulate locking
// ...
}

//
// Anywhere after complete() is called
qterm.Info("workload-memory-default is being set to workload-memory-maximum. If this is not intended please specify workload-memory-default to non-zero value")
```

You could also imagine this being a bit prettier refactored to look more like `pterm` itself (i.e. make `qterm` a package that decorates `pterm` more precisely)

njhale (approved):
Left a suggestion, but it's probably overkill for this

lgtm otherwise

g-linville (approved):

PR #2143, opened 2023-09-05
Split resources into BaseResources and QuotaRequestResources

tylerslaton:
This is being done for extensibility. The BaseResources will be used to specify resources that should always be tracked in any scope. QuotaRequestResources embeds the BaseResources struct and adds on the ability to track secrets. This is necessitated by the fact that secrets are not trackable at all scopes.

### Checklist


pkg/controller/quota/quota.go
```diff
@@ -79,11 +79,13 @@ func EnsureQuotaRequest(req router.Request, resp router.Response) error {
...
-				Jobs:       len(app.Jobs),
-				Volumes:    len(app.Volumes),
-				Images:     len(app.Images),
+			Resources: adminv1.QuotaRequestResources{
+				BaseResources: adminv1.BaseResources{
+					Containers: len(app.Containers),
+					Jobs:       len(app.Jobs),
+					Volumes:    len(app.Volumes),
+					Images:     len(app.Images),
+				},
```
tylerslaton, line 88: This is an example of how downstream dependencies will embed and use BaseResources.

pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -77,135 +44,52 @@ func (current *Resources) Remove(incoming Resources, all bool) {
...
-	checkResource("Secrets", current.Secrets)
-	checkResource("Images", current.Images)
-	checkResource("Projects", current.Projects)
-
-	checkQuantityResource("VolumeStorage", current.VolumeStorage)
-	checkQuantityResource("Memory", current.Memory)
-	checkQuantityResource("Cpu", current.CPU)
-
-	return strings.Join(resources, ", ")
+	return ""
```
thedadams, line 87: It's strange to me that something called `NonEmptyString` can return an empty string.

tylerslaton: I agree, this is not a great name for it. It came from the function that consumes it but I don't think that's the right fit. I'll change this.

pkg/apis/internal.admin.acorn.io/v1/baseresources_test.go
```diff
@@ -103,122 +103,115 @@ func TestAdd(t *testing.T) {
...
 				VolumeStorage: resource.MustParse("1Mi"),
 			},
-			incoming: Resources{
-				Secrets:       1,
+			incoming: BaseResources{
 				VolumeStorage: resource.MustParse("1Mi"),
 			},
-			expected: Resources{
-				Secrets:       1,
+			expected: BaseResources{
```
g-linville, line 170: This test is no longer useful/needed since secrets are no longer part of it.

pkg/apis/internal.admin.acorn.io/v1/baseresources_test.go
```diff
@@ -103,122 +103,115 @@ func TestAdd(t *testing.T) {
...
 				VolumeStorage: resource.MustParse("1Mi"),
 			},
-			incoming: Resources{
-				Secrets:       1,
+			incoming: BaseResources{
 				VolumeStorage: resource.MustParse("1Mi"),
 			},
 			all:      true,
-			expected: Resources{},
+			expected: BaseResources{},
```
g-linville, line 160: This test is no longer useful/needed now that secrets are not part of it.

pkg/apis/internal.admin.acorn.io/v1/resources.go
tylerslaton: Alright, updated these to be `ResourceToString` and `QuantityResourceToString` with a header comment. Also tweaked their logic a bit.

pkg/apis/internal.admin.acorn.io/v1/quotarequests.go
```diff
@@ -70,3 +73,50 @@ type QuotaRequestInstanceList struct {
...
+	current.BaseResources.Remove(incoming.BaseResources, all)
+}
+
+// Fits will check if a group of resources will be able to contain
+// another group of resources. If the resources are not able to fit,
+// an aggregated error will be returned with all exceeded resources.
+// If the current resources defines unlimited, then it will always fit.
+func (current *QuotaRequestResources) Fits(incoming QuotaRequestResources) error {
+	exceededResources := []string{}
+	Fits(exceededResources, "Secrets", current.Secrets, incoming.Secrets)
```
g-linville, line 102: I noticed in `BaseResources`'s implementation of `Fits`, `exceededResources` was being set to the return value of `Fits` for each resource, like this:

```suggestion
	exceededResources := []string{}
	exceededResources = Fits(exceededResources, "Secrets", current.Secrets, incoming.Secrets)
```

Does that need to happen here as well?

pkg/apis/internal.admin.acorn.io/v1/quotarequests.go
```diff
@@ -70,3 +73,50 @@ type QuotaRequestInstanceList struct {
...
+	// Build an aggregated error message for the exceeded resources
+	if len(exceededResources) > 0 {
+		return fmt.Errorf("%w: %s", ErrExceededResources, strings.Join(exceededResources, ", "))
+	}
+	return current.BaseResources.Fits(incoming.BaseResources)
```
g-linville, line 107: nit: should we wait until checking the return value of `BaseResources.Fits`, and then aggregate the two errors together, if that is possible?

pkg/apis/internal.admin.acorn.io/v1/quotarequests_test.go
```diff
@@ -0,0 +1,492 @@
...
+	}
+	// Run the test cases
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			tc.current.Remove(tc.incoming, tc.all)
+			assert.True(t, tc.current.Equals(tc.expected))
+		})
+	}
+}
+func TestEquals(t *testing.T) {
```
g-linville, line 308: suggestion: add two more test cases: one where the only thing unequal between `current` and `incoming` is `Secrets`, and one where the only thing unequal between `current` and `incoming` is something inside of `BaseResources`.

This would help make sure that both the logic that checks the secrets is correct, as well as the logic that calls the equal function on `BaseResources`.

pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -77,135 +44,56 @@ func (current *Resources) Remove(incoming Resources, all bool) {
...
+}
 
-	// Define function for checking quantity resources to keep code DRY
-	checkQuantityResource := func(resource string, currentVal resource.Quantity) {
-		if !currentVal.IsZero() {
-			resources = append(resources, resource)
-		}
+// ResourceToString will return a string representation of the resource and value
+// if its the value is greater than 0.
+func ResourceToString(resource string, value int) []string {
```
g-linville, line 85: Why does this return `[]string` instead of `string`?

pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -2,12 +2,14 @@ package v1
...
 	"fmt"
-	"strings"
 
 	"k8s.io/apimachinery/pkg/api/resource"
 )
 
-var ErrExceededResources = fmt.Errorf("quota would be exceeded for resources")
+var (
+	ErrExceededResources       = fmt.Errorf("quota would be exceeded for resources")
+	comparableUnlimitedQuntity = UnlimitedQuantity()
```
g-linville, line 11: nit:

```suggestion
	comparableUnlimitedQuantity = UnlimitedQuantity()
```

pkg/apis/internal.admin.acorn.io/v1/quotarequests.go
tylerslaton: Good eye, yes.

pkg/apis/internal.admin.acorn.io/v1/quotarequests.go
tylerslaton: Yep, that's possible. I'll add that.

pkg/apis/internal.admin.acorn.io/v1/resources.go
tylerslaton: I implemented it this way for a reason that no longer makes sense to me. I'll simplify this down and it'll use string instead.

pkg/apis/internal.admin.acorn.io/v1/quotarequests_test.go
tylerslaton: Good call. I added some similar cases to fits for similar reasons.

pkg/apis/internal.admin.acorn.io/v1/quotarequests.go
```diff
@@ -98,20 +99,29 @@ func (current *QuotaRequestResources) Remove(incoming QuotaRequestResources, all
...
 // If the current resources defines unlimited, then it will always fit.
 func (current *QuotaRequestResources) Fits(incoming QuotaRequestResources) error {
-	exceededResources := []string{}
-	Fits(exceededResources, "Secrets", current.Secrets, incoming.Secrets)
+	exceededResources := Fits([]string{}, "Secrets", current.Secrets, incoming.Secrets)
+
 	// Build an aggregated error message for the exceeded resources
+	var err error
 	if len(exceededResources) > 0 {
 		return fmt.Errorf("%w: %s", ErrExceededResources, strings.Join(exceededResources, ", "))
```
g-linville, line 107:
```suggestion
		err = fmt.Errorf("%w: %s", ErrExceededResources, strings.Join(exceededResources, ", "))
```

pkg/apis/internal.admin.acorn.io/v1/quotarequests_test.go
```diff
@@ -337,6 +337,41 @@ func TestEquals(t *testing.T) {
...
+		},
+		{
+			name: "unequal base resources only",
+			current: QuotaRequestResources{
+				BaseResources: BaseResources{
+					Apps:          1,
+					Containers:    1,
+					VolumeStorage: resource.MustParse("1Mi"),
+				},
+				Secrets: 1,
```
g-linville, line 365: remove in order for this test to do what we want it to

```suggestion
```

tylerslaton: You're right that something needs to change here, but I actually want to add the `Secrets: 1` to the incoming.

pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -77,135 +45,64 @@ func (current *Resources) Remove(incoming Resources, all bool) {
...
 
-	// Define function for checking quantity resources to keep code DRY
-	checkQuantityResource := func(resource string, currentVal resource.Quantity) {
-		if !currentVal.IsZero() {
-			resources = append(resources, resource)
+	for resource, quantity := range quantityResources {
+		if quantity.CmpInt64(0) > 0 {
+			resourceStrings = append(resourceStrings, fmt.Sprintf("%s: %s", resource, quantity.String()))
+		}
+		if quantity.Equal(comparableUnlimitedQuantity) {
```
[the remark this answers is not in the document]
tylerslaton: I made this an `else if` which I believe is what you meant?

pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -77,135 +45,64 @@ func (current *Resources) Remove(incoming Resources, all bool) {
...
-		return fmt.Errorf("%w: %s", ErrExceededResources, strings.Join(exceededResources, ", "))
+func Fits(toAppend []string, resource string, current, incoming int) []string {
+	if current != Unlimited && current < incoming {
+		return append(toAppend, resource)
 	}
+	return toAppend
+}
 
-	return nil
+func FitsQuantity(toAppend []string, resource string, current, incoming resource.Quantity) []string {
```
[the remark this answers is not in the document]
tylerslaton: Updated this to be a bit more intuitive.

pkg/apis/internal.admin.acorn.io/v1/baseresources_test.go
tylerslaton: Talked offline, but `VolumeStorage` is a persistent quantity, not a count. The same logic should still be applied so I think all that needs changing here is the name of the test.

pkg/apis/internal.admin.acorn.io/v1/resources.go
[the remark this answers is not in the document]
thedadams: Yes, thank you.

pkg/apis/internal.admin.acorn.io/v1/baseresources_test.go
```diff
@@ -103,122 +103,115 @@ func TestAdd(t *testing.T) {
...
 		},
 		{
-			name: "does not remove persistent counts without all",
-			current: Resources{
-				Secrets:       1,
+			name: "does not remove persistent resources counts without all",
```
g-linville, line 163: nit:

```suggestion
			name: "does not remove persistent resources without all",
```

pkg/apis/internal.admin.acorn.io/v1/quotarequests_test.go
```diff
@@ -0,0 +1,598 @@
...
+					VolumeStorage: resource.MustParse("2Mi"),
+				},
+				Secrets: 2,
+			},
+			expected: QuotaRequestResources{
+				BaseResources: BaseResources{},
+			},
+		},
+		{
+			name: "remove resources counts with all",
```
g-linville, line 191: nit:

```suggestion
			name: "remove persistent resources with all",
```

pkg/apis/internal.admin.acorn.io/v1/quotarequests_test.go
```diff
@@ -0,0 +1,598 @@
...
+				},
+				Secrets: 1,
+			},
+			all: true,
+			expected: QuotaRequestResources{
+				BaseResources: BaseResources{},
+			},
+		},
+		{
+			name: "does not remove resources counts without all",
```
g-linville, line 210: nit:

```suggestion
			name: "does not remove persistent resources without all",
```

g-linville (approved):

pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -77,135 +45,63 @@ func (current *Resources) Remove(incoming Resources, all bool) {
...
-func (current *Resources) NonEmptyString() string {
-	var resources []string
+// ResourceToString will return a string representation of the resource and value
+// if its value is greater than 0.
+func ResourcesToString(resources map[string]int, quantityResources map[string]resource.Quantity) string {
+	var resourceStrings []string
 
-	// Define function for checking int resources to keep code DRY
-	checkResource := func(resource string, value int) {
+	for resource, value := range resources {
```
thedadams, line 90: Similar comment to above: this will lead to inconsistent strings. Not sure if this will affect anything, but wanted to call it out just in case.

pkg/apis/internal.admin.acorn.io/v1/resources.go
```diff
@@ -77,135 +45,63 @@ func (current *Resources) Remove(incoming Resources, all bool) {
...
-	// Define function for checking quantity resources to keep code DRY
-	checkQuantityResource := func(resource string, currentVal resource.Quantity) {
-		if !currentVal.IsZero() {
-			resources = append(resources, resource)
+	for resource, quantity := range quantityResources {
+		if quantity.CmpInt64(0) > 0 {
+			resourceStrings = append(resourceStrings, fmt.Sprintf("%s: %s", resource, quantity.String()))
+		} else if quantity.Equal(comparableUnlimitedQuantity) {
+			resourceStrings = append(resourceStrings, fmt.Sprintf("%s: unlimited", resource))
 		}
```
njhale, line 103: tangent: idk why, but I grok switch statements _so_ much more easily than multi-branch if-else statements e.g.

```go
switch {
case quantity.CmpInt64(0) > 0:
  // ...
case quantity.Equal(comparableUnlimitedQuantity):
  // ...
}
```

pkg/apis/internal.admin.acorn.io/v1/quotarequests.go
```diff
@@ -70,3 +73,58 @@ type QuotaRequestInstanceList struct {
...
+	if all {
+		current.Secrets = Sub(current.Secrets, incoming.Secrets)
+	}
+	current.BaseResources.Remove(incoming.BaseResources, all)
+}
+
+// Fits will check if a group QuotaRequestResources will be able to contain
+// another group of QuotaRequestResources. If the QuotaRequestResources are not able to fit,
+// an aggregated error will be returned with all exceeded QuotaRequestResources.
+// If the current QuotaRequestResources defines unlimited, then it will always fit.
```
njhale, line 101: 💯 very nice go doc comments

pkg/apis/internal.admin.acorn.io/v1/baseresources.go
```diff
@@ -0,0 +1,121 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+// BaseResources defines resources that should be tracked at any scoped. The two main exclusions
+// currently are Secrets and Projects as they have situations they should be not be tracked.
+
```
njhale, line 12:
```suggestion
```

njhale (approved):

pkg/apis/internal.admin.acorn.io/v1/resources.go
thedadams: @tylerslaton Curious what your thoughts are here.

tylerslaton: Ah, let me think on this. My bad, thought I covered this one already.

tylerslaton: Alright I think I came up with a good solution here. Now I'm using the `typed.Sorted` function here. This felt right as it keeps the simplicity of the function while maintaining determinism. Thoughts?

pkg/apis/internal.admin.acorn.io/v1/baseresources.go
```diff
@@ -0,0 +1,121 @@
...
+func (current *BaseResources) Fits(incoming BaseResources) error {
+	exceededResources := []string{}
+
+	for resource, values := range map[string][]int{
+		"Apps":       {current.Apps, incoming.Apps},
+		"Containers": {current.Containers, incoming.Containers},
+		"Jobs":       {current.Jobs, incoming.Jobs},
+		"Volumes":    {current.Volumes, incoming.Volumes},
+		"Images":     {current.Images, incoming.Images},
+	} {
```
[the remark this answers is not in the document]
tylerslaton: I believe that this has been addressed based on the implementation you suggested. Thanks Nick! For the future there is also a `typed.Sorted` function that could be used in situations like this.

pkg/apis/internal.admin.acorn.io/v1/resources.go
thedadams: That's fine. My overall opinion would be that we use the same approach in all of these cases (you uses an anonymous struct in the other case), but I won't hold up approval on that.

thedadams (approved):

njhale (approved):

tylerslaton:
For reference, I'm holding off merging this until some the downstream consumer of this change is ready to merge as well. At that point I will merge both together 👍🏻

Issue #2115, opened 2023-08-23
fix: `image verify` should not print key on failure


iwilltry42:
As reported by @g-linville via Slack

This happened when referencing a private key file during verification.
First of all, given a path to a file, we shouldn't log the contents.
Second, we should reject private keys to avoid sending them to the server.

Solutions:
a) extract the public key from the private key and use that one
b) just reject as soon as we notice it's a private key

sangee2004:
Tested with acorn version - `v0.8.0-13-gab9787b8+ab9787b8`

Following error message is presented to user when `acorn image verify` is done using a privatekey.
```
  ✗  ERROR:  failed to load public key from invalid multi-line key reference
```

Issue #2063, opened 2023-08-14
Not able to sign images using openssh keys.


sangee2004:
**Acorn version** - `v0.8.0-rc4+b6a95e8a`

**Steps to reproduce the problem:**

**Scenario1:**
1. Generate keys with `ssh-keygen -t rsa`
Try to sign an image with any of the above keys

`acorn image sign sangeetha/myfirstacorn:v0.0.4 --key <private key>`

```
acorn image sign sangeetha/myfirstacorn:v0.0.4 --key sangeegh1
  •  Signing Image sangeetha/myfirstacorn:v0.0.4 (digest: index.docker.io/sangeetha/myfirstacorn@sha256:a6e655dab6d39796d297a1b8445c549594ae19a5874e1edb287622490df41833) using key sangeegh1
? Enter password for private key sangeegh1: *********
  ✗  ERROR:  unsupported private key
```

2. Generate keys with `ssh-keygen -t rsa -m pem`
```
acorn image sign sangeetha/myfirstacorn:v0.0.4 --key sangeepem
  •  Signing Image sangeetha/myfirstacorn:v0.0.4 (digest: index.docker.io/sangeetha/myfirstacorn@sha256:a6e655dab6d39796d297a1b8445c549594ae19a5874e1edb287622490df41833) using key sangeepem
? Enter password for private key sangeepem: *********
  ✗  ERROR:  error parsing rsa private key: asn1: structure error: tags don't match (16 vs {class:0 tag:11 length:79 isCompound:true}) {optional:false explicit:false application:false private:false defaultValue:<nil> tag:<nil> stringType:0 timeType:0 set:false omitEmpty:false} pkcs1PrivateKey @2
```

3. Generate keys with `ssh-keygen -t ed25519`
```
 acorn image sign sangeetha/myfirstacorn:v0.0.4 --key sangeeed
  •  Signing Image sangeetha/myfirstacorn:v0.0.4 (digest: index.docker.io/sangeetha/myfirstacorn@sha256:a6e655dab6d39796d297a1b8445c549594ae19a5874e1edb287622490df41833) using key sangeeed
? Enter password for private key sangeeed: ********
  ✗  ERROR:  unsupported private key
```

Note - I am able to sign images successfully using keys generated by cosign . But I am not allowed to store these keys in github.

sangee2004:
Tested with acorn version - `acorn version v0.8.0-rc4-58-gbe49d0bd+be49d0bd`

[...]

PR #2145, opened 2023-09-06
Fix: Importing Private Keys (SSH Keys) for signing and deny them for verification  (#2063 + #2115)

iwilltry42:
Using SSH Keys for `acorn image sign` just failed everytime with `✗  ERROR:  decrypt: encrypted: unexpected kdf parameters` this is fixed with this PR.
Thispart  is mostly an update of the cosign/sigstore/crypto dependencies, which forced me to make some adjustments in the code.

Ref #2063

Also, we allowed private keys for verification, which lead to printing them in logs on client and potentially server-side.
This is now prevented as well. Users will see an error in the CLI when trying to use a private key for `acorn image verify`.
If we want, we could allow them, catch the issue early on, import the private key and extract the public key... a bit of extra code I'm not sure is required.

Ref #2115 

### Checklist


g-linville (approved):
Very nice

thedadams (approved):

Issue #2149, opened 2023-09-07
Feature: `acorn.io/signed-ame` annotation on every signature


iwilltry42:
... to be used in every signature verification

This is to make sure on verification time that this exact name (reference) was used to sign the image.

E.g. 
1. sign the image as `foo/bar:v1`
2. verify the image as `foo/bar:v1` -> Success
3. tag the image as `some/other:tag
4. verify the image as `some/other:tag` -> Failure
5. verify the image as `some/other:tag` with `--no-verify-name` -> Success

@sangee2004 with this, signatures created with older versions will fail verification unless you use `--no-verify-name`.

sangee2004:
Tested with acorn version - `v0.8.0-21-g6bb72c0c+6bb72c0c`
1. Build image
```
%acorn build -t mytest:v1 .
```
2. Sign image
```
 % acorn image sign mytest:v1  --key ~/sangeeed
  •  Signing Image mytest:v1 (digest: index.docker.io/library/mytest@sha256:bc2802c1b5cfc7704c69018dd84491f0ea5aee6131445f50dcdfa180085c57a7) using key /Users/sangeethahariharan/sangeeed
? Enter password for private key /Users/sangeethahariharan/sangeeed: ********
  ✔  Created signature sha256:f34af5b83aef85be5b790ccbae1f1caf8934ff3b89f02c78df278d62f9e17462
```
3. Verify image
```
% acorn image verify mytest:v1  --key ~/sangeeed.pub
  •  Verifying Image mytest:v1 (digest: index.docker.io/library/mytest@sha256:bc2802c1b5cfc7704c69018dd84491f0ea5aee6131445f50dcdfa180085c57a7) using key /Users/sangeethahariharan/sangeeed.pub
  ✔  Signature verified
```
4. Tag this image as a different image tag
```
 % acorn tag mytest:v1  mytestnew:v1      
```
5. Verify new image tag. Verification fails as expected.
```           
 % acorn image verify mytestnew:v1  --key ~/sangeeed.pub
  •  Verifying Image mytestnew:v1 (digest: index.docker.io/library/mytestnew@sha256:bc2802c1b5cfc7704c69018dd84491f0ea5aee6131445f50dcdfa180085c57a7) using key /Users/sangeethahariharan/sangeeed.pub
[...]

PR #2150, opened 2023-09-07
add: default acorn.io/signed-name annotation for signatures (opt-out for verification) (#2149)

iwilltry42:
This is a security requirement for IRBs in Manager and for upcoming IRAs in Runtime.
During signing this is transparent for the user (we just add that annotation - it may be overridden by the user, making verification fail where the check is enforced).
During verification it can be turned off by the `--no-verify-name` flag (open to better names) and we will enforce it where we require it.

This required some changes in the way we verify the signature annotations, since the upstream packages we use for this are a little too strict (they're for Kubernetes label validation).

New flags: `--no-verify-name` on `acorn image verify` and `acorn pull`.

Ref #2149

### Checklist

### DIFF
--- a/pkg/cosign/types.go
+++ b/pkg/cosign/types.go
@@ -1,0 +1,11 @@
+package cosign
+
+const (
+	SignatureAnnotationSignedRef = "acorn.io/signedRef" // If an image was signed by `acorn image sign foo/bar:v1`, this annotation should be set to `foo/bar:v1` (the payload usually only includes the image digest)
+)
+
+func GetDefaultSignatureAnnotations(imageRef string) map[string]interface{} {
+	return map[string]interface{}{
+		SignatureAnnotationSignedRef: imageRef,
+	}
+}
--- a/pkg/cosign/types.go
+++ b/pkg/cosign/types.go
@@ -1,11 +1,11 @@
 package cosign
 
 const (
-	SignatureAnnotationSignedRef = "acorn.io/signedRef" // If an image was signed by `acorn image sign foo/bar:v1`, this annotation should be set to `foo/bar:v1` (the payload usually only includes the image digest)
+	SignatureAnnotationSignedName = "acorn.io/signedName" // If an image was signed by `acorn image sign foo/bar:v1`, this annotation should be set to `foo/bar:v1` (the payload usually only includes the image digest)
 )
 
-func GetDefaultSignatureAnnotations(imageRef string) map[string]interface{} {
+func GetDefaultSignatureAnnotations(imageName string) map[string]interface{} {
 	return map[string]interface{}{
-		SignatureAnnotationSignedRef: imageRef,
+		SignatureAnnotationSignedName: imageName,
 	}
 }
--- a/pkg/cosign/types.go
+++ b/pkg/cosign/types.go
@@ -1,7 +1,7 @@
 package cosign
 
 const (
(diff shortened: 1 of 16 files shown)


pkg/imageallowrules/selector/selector.go
```diff
@@ -0,0 +1,93 @@
...
+		case metav1.LabelSelectorOpNotIn:
+			op = selection.NotIn
+		case metav1.LabelSelectorOpExists:
+			op = selection.Exists
+		case metav1.LabelSelectorOpDoesNotExist:
+			op = selection.DoesNotExist
+		default:
+			return nil, fmt.Errorf("%q is not a valid label selector operator", expr.Operator)
+		}
+		r, err := labels.NewRequirement(expr.Key, op, append([]string(nil), expr.Values...))
```
g-linville, line 63: The `[]string(nil)` bit seems weird to me. Is this equivalent to `[]string{}`? I have never seen this before.

pkg/imageallowrules/selector/selector.go
```diff
@@ -0,0 +1,93 @@
...
+	return selector, nil
+}
+
+var LabelValueMaxLengthErrMsg string = validation.MaxLenError(validation.LabelValueMaxLength)
+
+const LabelValueRegexpErrMsg string = "a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character"
+
+func IgnoreInvalidFieldErrors(detailsPrefixes ...string) func(error) bool {
+	return func(err error) bool {
+		if ferr := err.(*field.Error); errors.As(err, &ferr) && ferr.Type == field.ErrorTypeInvalid {
```
g-linville, line 80: nit: I think you can do this instead:

```suggestion
		if ferr, ok := err.(*field.Error); ok && ferr.Type == field.ErrorTypeInvalid {
```

g-linville (approved):

pkg/imageallowrules/selector/selector.go
iwilltry42: Heh, I mixed up two variations there:
```suggestion
		if ferr := (*field.Error)(nil); errors.As(err, &ferr) && ferr.Type == field.ErrorTypeInvalid {
```
and
```suggestion
                var ferr *field.Error
		if errors.As(err, &ferr) && ferr.Type == field.ErrorTypeInvalid {
```

... not sure which one I prefer. Just type-casting bit me a few times before.

pkg/cosign/types.go
```diff
@@ -0,0 +1,11 @@
+package cosign
+
+const (
+	SignatureAnnotationSignedName = "acorn.io/signedName" // If an image was signed by `acorn image sign foo/bar:v1`, this annotation should be set to `foo/bar:v1` (the payload usually only includes the image digest)
```
njhale, line 4: Looks like the k8s naming convention for annotation keys calls for `lower-kebab-case` (see https://github.com/kubernetes/api/blob/master/core/v1/annotation_key_constants.go for precedent). Could we change this to align before merging?

```suggestion
	SignatureAnnotationSignedName = "acorn.io/signed-name" // If an image was signed by `acorn image sign foo/bar:v1`, this annotation should be set to `foo/bar:v1` (the payload usually only includes the image digest)
```

njhale (changes requested):

pkg/cosign/types.go
iwilltry42: It's not a k8s annotation, but I don't mind the case here :man_shrugging:

njhale (approved):

Issue #2153, opened 2023-09-08
fix: enable cert-manager for custom cluster-domains


iwilltry42:
As reported by Beta User.

To replicate:
1. Install cluster with AcornDNS disabled (optional) and custom cluster-domain and a cert-manager issue defined: `acorn install --acorn-dns=disabled --cluster-domain my.domain.com --cert-manager-issuer=letsencrypt-prod`
2. Run an app and set a custom endpoint:: `acorn run -p foo.my.domain.com:webapp ghcr.io/acorn-io/hello-world:v1`
3. Observe, how `foo.my.domain.com` gets an https endpoint and the generated endpoint does not

Investigation:
- Inspecting the Ingress objects shows  that the cert-manager annotation is not set on the generated endpoint
- https://github.com/acorn-io/runtime/blob/87e1aef3da18c125d34faa6631f8147d61645203/pkg/publish/cert.go#L151 only sets the annotation if it's a custom domain

sangee2004:
Able to reproduce the issue with acorn version -`v0.8.0+89036514`

1. Install cluster with AcornDNS disabled (optional) and custom cluster-domain and a cert-manager issue defined: acorn install --acorn-dns=disabled --cluster-domain my.domain.com --cert-manager-issuer=letsencrypt-prod
2. Run an app and set a custom endpoint  -  `acorn run -p foo.my.domain.com:webapp ghcr.io/acorn-io/hello-world:v1`

```
acorn apps
NAME             IMAGE                             COMMIT         CREATED     ENDPOINTS                                                                        MESSAGE
dark-waterfall   ghcr.io/acorn-io/hello-world:v1   eab5f8c77bd6   4m51s ago   https://foo.my.domain.com, http://webapp-dark-waterfall-11d87400.my.domain.com   OK
```

Tested with `acorn version v0.8.0-26-g6dfde7eb+6dfde7eb`

Following the same set of steps , we get `https` for generated endpoint as well.

```
acorn apps
NAME           IMAGE                             COMMIT         CREATED   ENDPOINTS                                                                       MESSAGE
polished-dew   ghcr.io/acorn-io/hello-world:v1   eab5f8c77bd6   7s ago    https://foo.my.domain.com, https://webapp-polished-dew-4fbaa3dc.my.domain.com   OK
```

iwilltry42:
So this is OK to close @sangee2004 ? :+1:

PR #2154, opened 2023-09-08
change: only exclude AcornDNS cluster domains from cert-manager annotations (#2153)

iwilltry42:
Ref #2153

Before, we excluded **all** cluster-domains, including custom cluster-domains, meaning those endpoints wouldn't get certificates and you had to explicitly specify a domain via `-p` to get https on the custom domain.

Additionally: 

- fix: do not try to provision wildcard cert if no domain/token is present

### Checklist

### DIFF
--- a/pkg/publish/ingress.go
+++ b/pkg/publish/ingress.go
@@ -222,7 +222,7 @@ func Ingress(req router.Request, svc *v1.ServiceInstance) (result []kclient.Obje
 					if err != nil {
 						return nil, err
 					}
-					if domain == acornDomain {
+					if domain == acornDomain || strings.HasSuffix(domain, profiles.ClusterDomainDefault) {
 						acornDomainTargets[hostname] = Target{Port: port.TargetPort, Service: svc.Name}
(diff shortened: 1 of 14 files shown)


pkg/publish/ingress.go
```diff
@@ -186,14 +188,21 @@ func Ingress(req router.Request, svc *v1.ServiceInstance) (result []kclient.Obje
 		}
 	}
 
+	dnsSecret := &corev1.Secret{}
+	err = req.Client.Get(req.Ctx, router.Key(system.Namespace, system.DNSSecretName), dnsSecret)
+	if err != nil && !apierrors.IsNotFound(err) {
```
g-linville, line 193: Why are we ignoring the not found error? It looks like the code after this assumes that the secret exists, so I feel like that error should be returned if we get it here.

iwilltry42: The secret won't be there, if Runtime was installed with `--acorn-dns=disabled`, in which case we're ok comparing against an empty string for `acornDNSDomain`

g-linville (approved):

StrongMonkey (approved):

Issue #1834, opened 2023-06-23
Add ability to add additional k8s roles to acorn controller


drpebcak:
For example, when deploying acorn into a cluster that has rancher installed, an additional permission is required in order to deploy apps

```
- apiGroups:
  - management.cattle.io
  resources:
  - projects
  verbs:
  - updatepsa
  ````

This could be expanded for other usecases potentially such as openshift, which often requires additional permissions for things to operate. Ideally we could pass a list of roles when installing acorn, and the rolebindings will be created and managed by the installation process.

PR #2158, opened 2023-09-12
Adjust RBAC to run on OpenShift

sam6258:
### Checklist

This PR is related to issue #1834 in that it adds the ability to run acorn runtime on OCP. 

This PR does the following:

    - Add acorn-image-system ClusterRole, ClusterRoleBinding, and SA
    - Bind nonroot-v2 SCC to acorn-image-system ClusterRole
    - Bind nonroot-v2 SCC to acorn-system ClusterRole

As a side note, the registry and controller pods seems to work fine as any UID, but apiserver requires UID 1000 for some stuff at the moment. Later, should evaluate if the apiserver deployment can move to any UID. I hit permission errors with the `apiserver.local.config` certs directory when trying to move to any UID.

This is my first PR to this repo, so if I'm missing anything here, please feel free to correct me. I tried running `make test` but started hanging partly through the integration suite, so I might need to configure something else for OCP here. I still want to get this PR open/reviewed for general feedback, so I'll work on getting my test setup going in the background. I didn't add unit tests as I think this is covered by the repo's existing integration tests/deployments being functional.

I don't think there are upgrade impacts here as both the registry and controller deployments seem to function as any UID already.

### DIFF
--- a/pkg/imagesystem/registrytemplate.go
+++ b/pkg/imagesystem/registrytemplate.go
@@ -39,7 +39,16 @@ func registryService(namespace string) []client.Object {
 	}
 }
 
-func registryDeployment(namespace, registryImage string, requirements corev1.ResourceRequirements, volumeSource corev1.VolumeSource) []client.Object {
+func registryServiceAccount(namespace string) client.Object {
+	return &corev1.ServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      system.RegistryServiceAccountName,
+			Namespace: system.ImagesNamespace,
+		},
+	}
+}
+
+func registryDeployment(namespace, serviceAccountName, registryImage string, requirements corev1.ResourceRequirements, volumeSource corev1.VolumeSource) []client.Object {
 	deployment := &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      system.RegistryName,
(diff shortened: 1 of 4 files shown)


pkg/imagesystem/registrytemplate.go
```diff
@@ -102,7 +102,6 @@ func registryDeployment(namespace, registryImage string, requirements corev1.Res
 								FailureThreshold:    3,
 							},
 							SecurityContext: &corev1.SecurityContext{
-								RunAsUser:                z.Pointer[int64](1000),
```
sam6258, line 105: This probably breaks non-ocp k8s deploys. Ill probably have to create nonroot rbac for this and controller deployment.

sam6258: Made this change.

ibuildthecloud:
The changes here seem simple enough.  The only issue is that if we do accept these changes we really have no idea going forward if openshift will work. We don't test on OpenShift. So I have no problem accepting this PR, just that we are basically just expecting the community to tell us if we happen to break OpenShift.

@thedadams When you get a second can you review this one?

sam6258:
Sure thing, thanks @ibuildthecloud. I was just trying out the project and figured I would help out since it was fairly easy.

thedadams (approved):
Thanks for the contribution here. This looks good to me. As @ibuildthecloud said, if this breaks at some point, please let us know/submit another change.

iwilltry42 (approved):
LGTM

Issue #2148, opened 2023-09-07
Feature: ImageRoleAuthorization + ClusterImageRoleAuthorization resources


iwilltry42:
- [x]  Introduce `ImageRoleAuthorization` and `ClusterImageRoleAuthorization` CRDs in the `[admin.acorn.io](http://admin.acorn.io)` API-Group
- [x]  Introduce a feature flag which enables “deny-by-default” for all permissions (unless allowed by one of the above)

## How to Test

1. Install acorn with `acorn install --features image-role-authorizations=true`
2. Have an Acorn ready that requests some permissions (you have to know them)
3. Run the Acorn
4. Ensure that the Acorn won't start properly and show an `permissions denied` error in the status output
5. Create an IRA that covers the image and references a role that covers the permissions - if you don't have a role like that, you can create it as well
6. Run the Acorn again
7. Verify that it now starts successfully

Here are some example files:

Acornfile:
```
containers: web: {
    image: "nginx"
    permissions: {
		// These are permissions that will only be granted for this container in its namespace.
		rules: [
      	  {
            verbs: ["*"],
        	apiGroups: ["foo.bar.com"],
            resources: ["*"]
     	   },
         ]
}
```

Role + IRA that just allow everything:
```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: acorn:system:cluster-admin
rules:
- apiGroups:
  - '*'
  resources:
  - '*'
  verbs: ["*"]
---
apiVersion: admin.acorn.io/v1
kind: ImageRoleAuthorization
metadata:
  name: image-role-authz-superadmintest
  namespace: acorn
spec:
  imageSelector:
    namePatterns:
      - "**"
  roles:
    scopes:
      - cluster
    roleRefs:
      - name: acorn:system:cluster-admin
        kind: ClusterRole

```

sangee2004:
Tested with acorn version - v0.9.0-rc1-20-g1fc37ec2+1fc37ec2
This feature is available and the workflow mentioned in the issue works as expected.

Following are the issues found when testing this feature:

https://github.com/acorn-io/runtime/issues/2233
https://github.com/acorn-io/runtime/issues/2234

PR #2160, opened 2023-09-14
add: ImageRoleAuthorizations and improve shared ImageSelector logic (#2148)

iwilltry42:
Ref #2148 

## Added

- Feature flag `image-role-authorizations` which enables a "deny-by-default" policy for all permissions requested by images
- `ClusterImageRoleAuthorization` + `ImageRoleAuthorization` resources in the Admin API to allow permissions (via roles) for images matching the imageselector (name pattern + signature selector)

## Changed

- Breaking: Refactored common logic used by IRBs, IARs and IRAs into the new `ImageSelector` type consisting of name patterns and signature matchers
- This also changes the way we evaluate IARs -> From now on, they only act on the staged appImage, not on the currently running one. Since it's fairly expensive to re-evaluate the rules, we decided to stop doing this for now and only notify on changes (i.e. when there is a new staged appImage).

### Checklist


pkg/controller/permissions/permissions_check.go
```diff
@@ -67,6 +72,40 @@ func CheckImagePermissions(req router.Request, resp router.Response) error {
...
+	if enabled, err := config.GetFeature(req.Ctx, req.Client, profiles.FeatureImageRoleAuthorizations); err != nil {
+		return err
+	} else if enabled {
+		imageName := appImage.Name
+
+		// E.g. for child Acorns, the appImage.Name is the image ID, but we need the original image name (with registry/repo)
+		// to check for the signatures
+		if oi, ok := app.GetAnnotations()[labels.AcornOriginalImage]; ok {
+			imageName = oi
+		}
```
iwilltry42, line 85: Just realized that this may be bad, since we're looking at the staged appImage which may or may not match that annotation - does anyone have another idea?

integration/client/imagerules/imageroleauthorizations_test.go
```diff
@@ -0,0 +1,358 @@
...
+					Rules: []internalv1.PolicyRule{{
+						PolicyRule: rbacv1.PolicyRule{
+							APIGroups: []string{"aws.acorn.io"},
+							Verbs:     []string{"get"},
+						},
+					}},
+				},
+			},
+		},
+		Status: internalv1.AppInstanceStatus{},
```
g-linville, line 158: nit: unneeded

```suggestion
```

integration/client/imagerules/imageroleauthorizations_test.go
```diff
@@ -0,0 +1,358 @@
...
+					PolicyRule: rbacv1.PolicyRule{
+						APIGroups: []string{"foo.bar.com"},
+						Verbs:     []string{"get"},
+					},
+				},
+			},
+		},
+	}
+	require.Equal(t, expectedDeniedPermissionsRootapp, app.Status.Staged.ImagePermissionsDenied, "should have denied permissions for foo.bar.com on rootapp")
+	rmapp(ctx, app)
```
g-linville, line 230: I think this `rmapp` call should be done in a `t.Cleanup` call prior to these `require.Equal` calls. That way the app gets cleaned up if the test fails here.

integration/client/imagerules/imageroleauthorizations_test.go
```diff
@@ -0,0 +1,358 @@
...
+						Verbs:     []string{"get"},
+					},
+				},
+			},
+		},
+	}
+
+	require.Equal(t, expectedDeniedPermissionsFooAwsapp, nestedApp.Status.Staged.ImagePermissionsDenied, "should have denied permissions for aws.acorn.io on foo.awsapp")
+
+	rmapp(ctx, app)
```
g-linville, line 324: same here

integration/client/imagerules/imageroleauthorizations_test.go
```diff
@@ -0,0 +1,358 @@
...
+
+	// --------------------------------------------------------------------
+	// Run #4 - Expect no denied permissions since we have an IRA that covers the image
+	// --------------------------------------------------------------------
+
+	// Add the missing api group to the IRA
+	ira.ImageSelector.NamePatterns = []string{tagName, nestedImageTagName}
```
g-linville, line 331: I thought the image selector was already set up for the parent image and the nested image. Why is it being modified here?

pkg/apis/internal.acorn.io/v1/appspec.go
```diff
@@ -577,6 +565,7 @@ func canCombine(left, right PolicyRule) (combined PolicyRule, ok bool) {
 }
 
 func GrantsAll(currentNamespace string, requestedPerms, grantedUserPerms []Permissions) (missing []Permissions, granted bool) {
+	logrus.Errorf("R: %#v\nG: %#v\n", requestedPerms, grantedUserPerms)
```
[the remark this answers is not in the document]
iwilltry42: Oh yeah.. removed :heavy_check_mark:

pkg/controller/appstatus/appstatus.go
```diff
@@ -124,6 +124,10 @@ func setPermissionCondition(app *v1.AppInstance) {
 		cond.Error(fmt.Errorf("can not run new image due to missing permissions: %w", &client2.ErrRulesNeeded{
 			Permissions: app.Status.Staged.PermissionsMissing,
 		}))
+	} else if len(app.Status.Staged.ImagePermissionsDenied) > 0 {
+		cond.Error(fmt.Errorf("can not run new image due to denied permissions: %w", &client2.ErrRulesNeeded{
```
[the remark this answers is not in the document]
iwilltry42: Changed in both :heavy_check_mark:

pkg/imagerules/imageroleauthorizations.go
```diff
@@ -0,0 +1,144 @@
...
+			if roleRef.Kind == "ClusterRole" {
+				roleName = "cluster/" + roleName
+			}
+			if eRole, ok := existingRoles[roleName]; ok {
+				perms = append(perms, resolveGenericRole(eRole, imageName, ar.Scopes))
+			} else {
+				logrus.Warnf("RoleRef references non-existent role [%s] in namespace: [%s]", roleName, namespace)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
iwilltry42: Yeah I may have focused too much on reducing API calls here - I assume it'd be unusual for an IRA to have more than just a few roles :grimacing:

pkg/imageselector/name/name.go
```diff
@@ -0,0 +1,78 @@
...
+func ImageCovered(image name.Reference, digest string, patterns []string) bool {
+	for _, pattern := range patterns {
+		// empty pattern? skip (should've been caught by IAR validation already)
+		if strings.TrimSpace(pattern) == "" {
+			continue
+		}
+
+		// not a pattern? must be exact match then.
+		if !imagepattern.IsImagePattern(pattern) {
+			if strings.TrimSuffix(image.Name(), ":") != pattern && digest != pattern {
```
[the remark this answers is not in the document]
iwilltry42: I've seen things.. 
Actually there are cases where the image reference gets parsed as a repository-only reference, i.e. without a tag or digest attached. Since there's [no validation that the identifier is non-empty in ggcr,](https://github.com/google/go-containerregistry/blob/main/pkg/name/tag.go#L53-L55) in those cases we'd end up with a `:` followed by nothing.

integration/client/imagerules/imageroleauthorizations_test.go
iwilltry42: Removed

pkg/imageselector/name/name.go
```diff
@@ -0,0 +1,78 @@
...
+		parts := strings.Split(pattern, ":")
+		contextPattern := parts[0]
+		tagPattern := ""
+		if len(parts) > 1 {
+			if !strings.Contains(parts[len(parts)-1], "/") {
+				tagPattern = parts[len(parts)-1] // last part is tag
+			} else {
+				contextPattern = pattern // : was part of the context pattern (port)
+			}
+		}
```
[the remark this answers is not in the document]
iwilltry42: Oh good catch! I forgot a line in the first part!

integration/client/imagerules/imageroleauthorizations_test.go
iwilltry42: Yeah, I just wanted to remove them as early as possible instead of deferring it to the cleanup.
However, shouldn't this also be handled by the TempProject cleanup?

integration/client/imagerules/imageroleauthorizations_test.go
iwilltry42: Because I didn't check what I set it up for before :)
Changed.

pkg/cosign/cosign_test.go
```diff
@@ -219,7 +220,11 @@ func TestVerifySignature(t *testing.T) {
 		}
 
 		opts.Key = tc.key
-		opts.AnnotationRules = tc.annotationrules
+		sel, err := signatureannotations.GenerateSelector(tc.annotationrules, signatureannotations.LabelSelectorOpts{})
```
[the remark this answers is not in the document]
iwilltry42: You mean the `LabelRequirementErrorFilters`?
They're not required in this test case. They're a requirement for the `acorn.io/signed-name` annotation, which we don't use in this test (and also no other annotations that could break this).

pkg/imagerules/imageroleauthorizations.go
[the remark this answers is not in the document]
g-linville: Oh that makes sense. It is fine either way

integration/client/imagerules/imageroleauthorizations_test.go
g-linville: Oh probably, yeah. Good point

pkg/imageselector/name/name.go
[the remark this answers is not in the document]
g-linville: Oh yeah, I didn't think about not having `latest` as a default tag. That makes sense

g-linville (approved):

pkg/controller/permissions/permissions_check.go
iwilltry42: Figured out that this is actually OK, since the label is always up-to-date with the staged appImage :heavy_check_mark:

g-linville: Awesome! I couldn't think of any other ideas for this so I am glad that this will work

pkg/imagerules/imageallowrules.go
```diff
@@ -0,0 +1,87 @@
...
+// @param imageName: the image how it was called (e.g. how it was specified by the user in `acorn run`)
+// @param resolvedName: the image name after resolution (e.g. resolved to an internal image ID)
+// @param digest: the digest of the image
+// We will use all of those to check if an image is covered by an IAR.
+// We will prefer resolvedName to find signature artifacts (potentially in the internal registry)
+func CheckImageAgainstRules(ctx context.Context, c client.Reader, namespace, imageName, resolvedName, digest string, imageAllowRules []v1.ImageAllowRuleInstance, opts ...remote.Option) error {
+	// No rules? Deny all images.
+	if len(imageAllowRules) == 0 {
+		return &ErrImageNotAllowed{Image: imageName}
+	}
```
thedadams, line 70: nit: not necessary.

pkg/imagerules/imageroleauthorizations.go
```diff
@@ -0,0 +1,144 @@
...
+
+	return resolveAuthorizedRoles(ctx, c, namespace, imageName, authorizedRoles)
+}
+
+func CheckRoleAuthorizations(ctx context.Context, c client.Reader, namespace, imageName, resolvedName, digest string, iras []internaladminv1.ImageRoleAuthorizationInstance, opts ...remote.Option) ([]internaladminv1.RoleAuthorizations, error) {
+	// No rules? Deny all images.
+	if len(iras) == 0 {
+		return nil, &ErrImageNotAllowed{Image: imageName}
+	}
```
thedadams, line 62: nit: not necessary

thedadams (approved):

pkg/imagerules/imageallowrules.go
iwilltry42: Yup, refactoring leftover.

pkg/imagerules/imageroleauthorizations.go
iwilltry42: Removed.

pkg/controller/permissions/permissions_check.go
```diff
@@ -3,74 +3,134 @@ package permissions
...
-		return nil
-	}
+// CheckPermissions checks various things related to permissions
+// a) if the image is allowed by the image allow rules (if enabled)
+// b) if the permissions requested by all images in the app are
+//
+//	b.1) granted by the user (as set in the app spec)
+//	b.2) authorized by the image role authorizations (if enabled)
+func CheckPermissions(transport http.RoundTripper) router.HandlerFunc {
+	return func(req router.Request, _ router.Response) error {
```
g-linville, line 48: Looks like this transport argument is unused.

pkg/roles/roles.go
```diff
@@ -15,9 +16,25 @@ const (
...
 var (
+	awsRoles = map[string][]rbacv1.PolicyRule{
+		AWSAdmin: {
+			{
+				APIGroups: []string{awspermissions.AWSAPIGroup, awspermissions.AWSRoleAPIGroup},
+				Verbs:     []string{"*"},
+				Resources: []string{"*"},
+			},
+		},
+	}
```
g-linville, line 36: Do we have a practical use case/need for this? This seems really scary to me.

pkg/apis/internal.acorn.io/v1/imageselector.go
```diff
@@ -0,0 +1,25 @@
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type SignedBy struct {
```
tylerslaton, line 7: General comment for all the structs here, but given that this feature has a lot of moving parts can we leave GoDoc comments explaining each of these? That should make our future-selves quite happy.

pkg/controller/permissions/permissions_check.go
iwilltry42: Oh yeah, removed and put back in place during my debugging sessions - will use it again.

pkg/roles/roles.go
iwilltry42: Yeah, basically for all allowed aws service acorns.
The role will be referenced by IRAs in Runtime and IRBs in Manager.
We'll have the almighty admin role until we figure out more granular roles that make sense for the services.
So far we had auto-approvals in Manager.

tylerslaton (approved):
Small nit, focusing on the Manager PR as this has been reviewed.

pkg/roles/roles.go
tylerslaton: I'm sure that this has already been discussed, but do these roles need to live in Runtime? They feel more Manager oriented to me.

PR #2164, opened 2023-09-15
Fix app validation when computeclasses change on stop

njhale:
App validation prevents Apps from being stopped when the available
ComputeClasses have changed. Change validation so that ComputeClasses
are ignored when stopping an App.

### DIFF
--- a/pkg/server/registry/apigroups/acorn/apps/validator.go
+++ b/pkg/server/registry/apigroups/acorn/apps/validator.go
@@ -171,35 +171,37 @@ func (s *Validator) Validate(ctx context.Context, obj runtime.Object) (result fi
 			return
 		}
 
-		disableCheckImageAllowRules := false
-		if app.Spec.Stop != nil && *app.Spec.Stop {
-			// app was stopped, so we don't need to check image allow rules (this could prevent stopping an app if the image allow rules changed)
-			disableCheckImageAllowRules = true
-		}
-
-		if !disableCheckImageAllowRules {
+		if !z.Dereference(app.Spec.Stop) {
+			// App was stopped, so we don't need to check image allow rules or compute classes
+			// (this could prevent stopping an app if either of these have changed)
 			if err := imageallowrules.CheckImageAllowed(ctx, s.client, app.Namespace, checkImage, image, imageDetails.AppImage.Digest, remote.WithTransport(s.transport)); err != nil {
 				result = append(result, field.Forbidden(field.NewPath("spec", "image"), fmt.Sprintf("%s not allowed to run: %s", app.Spec.Image, err.Error())))
 				return
 			}
-		}
 
-		workloadsFromImage, err := s.getWorkloads(imageDetails)
-		if err != nil {
-			result = append(result, field.Invalid(field.NewPath("spec", "image"), app.Spec.Image, err.Error()))
-			return
-		}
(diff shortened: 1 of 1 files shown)


pkg/server/registry/apigroups/acorn/apps/validator.go
```diff
@@ -171,35 +171,37 @@ func (s *Validator) Validate(ctx context.Context, obj runtime.Object) (result fi
...
 		}
 
-		disableCheckImageAllowRules := false
-		if app.Spec.Stop != nil && *app.Spec.Stop {
-			// app was stopped, so we don't need to check image allow rules (this could prevent stopping an app if the image allow rules changed)
-			disableCheckImageAllowRules = true
-		}
-
-		if !disableCheckImageAllowRules {
+		if !z.Dereference(app.Spec.Stop) {
```
njhale, line 174: For the stop case, is this good enough? or do we want to validate that _nothing else_ has changed but the `spec.stop` field? I'm worried about the case where other fields are changed on stop; e.g. I switch to a bad compute class.

pkg/server/registry/apigroups/acorn/apps/validator.go
```diff
@@ -171,35 +171,37 @@ func (s *Validator) Validate(ctx context.Context, obj runtime.Object) (result fi
...
+				workloadsFromImage,
+				apiv1cfg.WorkloadMemoryDefault,
+				apiv1cfg.WorkloadMemoryMaximum)
+			if len(errs) != 0 {
+				result = append(result, errs...)
+				return
+			}
 		}
 
 		if err := validateVolumeClasses(ctx, s.client, app.Namespace, app.Spec, imageDetails.AppSpec, project); err != nil {
```
tylerslaton, line 207: Do we want to also validateVolumeClasses when stopping an app? My thought is yes, but just want to check.

pkg/server/registry/apigroups/acorn/apps/validator.go
tylerslaton: So with your code, we stop an App and skip validation for imageAllowRules (previous behavior) and computeClasses (new). If I swap to a bad ComputeClass at the same time nothing will be validated. If I then start the App again it will go through validation for imageAllowRules and computeClasses and fail the computeClass validation. 

That behavior could be fine but could also be confusing if a user stopped an App a long time ago and starts an App up again to be met with a "invalid compute class" error. However, the behavior was the same for imageAllowRules before this PR, so its likely fine.

tylerslaton (approved):

pkg/server/registry/apigroups/acorn/apps/validator.go
njhale: >  However, the behavior was the same for imageAllowRules before this PR, so its likely fine

yeah, this was my sentiment exactly

pkg/server/registry/apigroups/acorn/apps/validator.go
njhale: I think so. If it become apparent it's an issue, then we can always preclude them during stop later.

pkg/server/registry/apigroups/acorn/apps/validator.go
njhale: I'm going to consider this resolved for now.

thedadams (approved):

PR #2165, opened 2023-09-15
Account for scale of a container when creating a QuotaRequest

tylerslaton:
Prior to this PR, container scale was not reflected in the QuotaRequest.

### Checklist

### DIFF
--- a/integration/run/run_test.go
+++ b/integration/run/run_test.go
@@ -1638,9 +1638,9 @@ func TestEnforcedQuota(t *testing.T) {
 		return kclient.Update(ctx, obj) == nil
 	})
 
-	// Run a simple app.
-	image, err := c.AcornImageBuild(ctx, "./testdata/simple/Acornfile", &client.AcornImageBuildOptions{
-		Cwd: "./testdata/volume",
+	// Run a scaled app.
+	image, err := c.AcornImageBuild(ctx, "./testdata/scaled/Acornfile", &client.AcornImageBuildOptions{
+		Cwd: "./testdata/scaled",
 	})
 	if err != nil {
 		t.Fatal(err)
@@ -1668,7 +1668,7 @@ func TestEnforcedQuota(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	assert.Equal(t, quotaRequest.Spec.Resources.Containers, 1)
+	assert.Equal(t, 2, quotaRequest.Spec.Resources.Containers)
 
 	// Update the status of the QuotaRequest to communicate readiness.
 	quotaRequest.Status = adminv1.QuotaRequestInstanceStatus{
--- a/pkg/controller/quota/quota.go
+++ b/pkg/controller/quota/quota.go
@@ -80,10 +80,9 @@ func EnsureQuotaRequest(req router.Request, resp router.Response) error {
 		ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: namespace},
 		Spec: adminv1.QuotaRequestInstanceSpec{
 			Resources: adminv1.Resources{
-				Containers: len(app.Containers),
-				Jobs:       len(app.Jobs),
-				Volumes:    len(app.Volumes),
-				Images:     len(app.Images),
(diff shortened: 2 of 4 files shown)


integration/run/run_test.go
```diff
@@ -1668,7 +1668,7 @@ func TestEnforcedQuota(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	assert.Equal(t, quotaRequest.Spec.Resources.Containers, 1)
+	assert.Equal(t, 1, quotaRequest.Spec.Resources.Containers)
```
njhale, line 1671: 👍

pkg/controller/quota/quota.go
```diff
@@ -102,6 +103,17 @@ func EnsureQuotaRequest(req router.Request, resp router.Response) error {
...
 }
 
+// addContainers adds the number of containers and accounts for the scale of each container.
+func addContainers(containers map[string]apiv1.Container, quotaRequest *adminv1.QuotaRequestInstance) {
+	for _, container := range containers {
+		scale := 1
+		if container.Scale != nil {
+			scale = int(*container.Scale)
+		}
+		quotaRequest.Spec.Resources.Containers += scale
```
njhale, line 113: So if scale is unset, we assume scale = 1? Is that the default?

njhale (approved):
I left a question, but I don't think it's blocking.

pkg/controller/quota/quota.go
tylerslaton: Yeah, it operates under the same rules as the [`Replicas` section of a DeploymentSpec](https://pkg.go.dev/k8s.io/api@v0.27.3/apps/v1#DeploymentSpec.Replicas:~:text=//%20Number%20of%20desired%20pods.%20This%20is%20a%20pointer%20to%20distinguish%20between%20explicit%0A%09//%20zero%20and%20not%20specified.%20Defaults%20to%201.%0A%09//%20%2Boptional%0A%09Replicas%20*int32%20%60json%3A%22replicas%2Comitempty%22%20protobuf%3A%22varint%2C1%2Copt%2Cname%3Dreplicas%22%60). We use it [here](https://github.com/tylerslaton/runtime/blob/07f8216fc68a5ab3c9e979e0f217a0c9c5225c6c/pkg/controller/appdefinition/deploy.go#L675-L676), to be more specific.

thedadams (approved):

tylerslaton:
@thedadams @njhale Addressing re-review as I caught a bug. I accounted for a nil scale when adding containers but not when adding compute reserved by containers. this has been fixed in the latest push.

pkg/controller/quota/quota.go
```diff
@@ -186,3 +196,13 @@ func isEnforced(req router.Request, namespace string) (bool, error) {
...
 }
+
+// replicas returns the number of replicas based on an int32 pointer. If the
+// pointer is nil, it is assumed to be 1.
+func replicas(s *int32) int {
+	scale := 1
+	if s != nil {
+		scale = int(*s)
+	}
+	return scale
```
thedadams, line 207: nit:
```suggestion
	if s != nil {
		return int(*s)
	}
	return 1
```

thedadams (approved):

pkg/controller/quota/quota.go
```diff
@@ -186,3 +196,12 @@ func isEnforced(req router.Request, namespace string) (bool, error) {
...
 }
+
+// replicas returns the number of replicas based on an int32 pointer. If the
+// pointer is nil, it is assumed to be 1.
+func replicas(s *int32) int {
+	if s != nil {
+		return int(*s)
+	}
+	return 1
+}
```
njhale, line 207: nit: This makes me want `z.DereferenceOrDefault`.

njhale (approved):

pkg/controller/quota/quota.go
tylerslaton: That could be nice, yeah. There have definitely been times I've wanted something like that. Its probably borderline though.

PR #2174, opened 2023-09-22
Ensure image cache purging respects context canceling

thedadams:
A previous change implemented image details caching and purging. However, the purge goroutine didn't respect any cancellation of the context passed to it. This change implements this behavior.

### DIFF
--- a/pkg/imagemetadatacache/imagemetadatacache.go
+++ b/pkg/imagemetadatacache/imagemetadatacache.go
@@ -13,12 +13,19 @@ import (
 var interval = 24 * time.Hour
 
 func Purge(ctx context.Context, c kclient.Client) {
+	timer := time.NewTimer(interval)
 	for {
-		err := doPurge(ctx, c)
-		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/imagemetadatacache/imagemetadatacache.go
```diff
@@ -13,12 +13,19 @@ import (
...
+	timer := time.NewTimer(interval)
 	for {
-		err := doPurge(ctx, c)
-		if err != nil {
-			logrus.Errorf("Failed to purge imagemetadatacache: %v", err)
+		select {
+		case <-ctx.Done():
+			timer.Stop()
+			return
+		case <-timer.C:
```
ibuildthecloud, line 22: In general there no advantage to using a timer over time.After() IF it is expected that the normal condition is to always hit the timeout.  The downside of time.After is that if you normally don't hit the timeout you end up with a lot of goroutines doing nothing.

The down side of the timer is you have to ensure you call Reset or Stop, or whatever.  I personally don't like timers or tickers unless I really need them.

PR #2176, opened 2023-09-22
Add support for interpolations to nested Acorns and services

g-linville:
This will be very helpful for me with some other things I am working on.

This is somewhat of a breaking change. Previously, if you did something like this:

```cue
acorns: myacorn: {
  // ...
  deployArgs: myarg: "@{service.someservice.data.datapoint}"
}
```

then it would be kept as that literal string, and be available for interpolation inside the nested acorn. But now, that will be interpolated at the top level, and the interpolated value will be passed into the nested acorn.

To keep the literal string in the nested acorn, you can do this:

```cue
acorns: myacorn: {
  // ...
  deployArgs: myarg: "@{@{service}.someservice.data.datapoint}"
}
```

In this case, the inner set of `@{}` will basically just disappear.

### Checklist

### DIFF
--- a/pkg/controller/appdefinition/acorn.go
+++ b/pkg/controller/appdefinition/acorn.go
@@ -15,14 +15,15 @@ import (
 	"github.com/acorn-io/runtime/pkg/labels"
 	"github.com/acorn-io/runtime/pkg/ports"
 	"github.com/acorn-io/runtime/pkg/publicname"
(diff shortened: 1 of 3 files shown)


g-linville:
Unassigning reviewers and putting this into draft. I think I might know a way to accomplish this without this change.

PR #2182, opened 2023-09-25
Add service container labels

thedadams:
The `containerLabels` option is available in the Acornfile, but it is not piped through the controllers correctly. This change ensures that it is.


g-linville (approved):

njhale (approved):
Lgtm

PR #2183, opened 2023-09-26
change: proper error handling for fetching pubkeys from GitHub API (#2185)

iwilltry42:
Ref #2185 

The error will be given to the user via CLI when doing `acorn image verify` and to the admin via controller logs (since we do the verification server-side, `acorn image verify` will also count against the rate limit of the Runtime installation's egress IP).

For the future we **could** think about an option for administrators to add GitHub credentials to the controller (e.g. a GitHub App has a rate limit of 15k requests per 60 minutes).
Alternatively we could figure out a smart way of caching public keys for GitHub users, not sure about the cache timeout there to pick up new keys or dismiss removed keys :thinking: 

### Checklist

### DIFF
--- a/pkg/cosign/github.go
+++ b/pkg/cosign/github.go
@@ -17,11 +17,21 @@ type GitHubPublicKey struct {
 func getGitHubPublicKeys(username string) ([]GitHubPublicKey, error) {
 	logrus.Debugf("Getting public keys for GitHub user %s", username)
 	url := fmt.Sprintf("https://api.github.com/users/%s/keys", username)
-	resp, err := http.Get(url)
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Accept", "application/json")
+	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
 		return nil, err
 	}
 	defer resp.Body.Close()
+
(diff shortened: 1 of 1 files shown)


pkg/cosign/github.go
```diff
@@ -17,12 +17,22 @@ type GitHubPublicKey struct {
...
+	}
+	req.Header.Set("Accept", "application/json")
+	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
 		return nil, err
 	}
 	defer resp.Body.Close()
 
+	if resp.StatusCode != http.StatusOK {
+		body, _ := io.ReadAll(resp.Body)
```
keyallis, line 32: No error handling on this read? Not sure if that would cause issues since we are trying to include the body in the error message. Also since we use the body whether we're erroring out or continuing the function we could refactor this body read out of the if before the check maybe

iwilltry42: `io.ReadAll` always returns a byte slice, even on failure, so I think this should not error out.
You're right though, that we could refactor this out since we use the body anyway, but I thought it might look weird since we do care about the read error if we need to use the body... within the status code handling, the body is just a nice to have :man_shrugging:

njhale (approved):

g-linville (approved):

PR #2206, opened 2023-10-03
Use context.AfterFunc instead of goroutine approach

thedadams:
Note that the stop function returned by context.AfterFunc doesn't run 
the function that is passed to context.AfterFunc. It disassociates the 
function from the context (i.e. releases the resources). Calling stop 
after the context is cancelled is valid, so stop should be called to 
ensure that the resources are released.

### DIFF
--- a/pkg/client/app.go
+++ b/pkg/client/app.go
@@ -316,20 +316,20 @@ func (c *DefaultClient) AppLog(ctx context.Context, name string, opts *LogOption
 		return nil, err
 	}
 
-	go func() {
-		<-ctx.Done()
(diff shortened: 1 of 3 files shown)


g-linville (approved):

njhale (approved):

thedadams:
Re-requesting reviews because I changed this. I misunderstood that `stop` will disassociate the passed function from the context, but not call the function. The function is only called if the context is cancelled.

g-linville (approved):

thedadams:
Closing because this isn't important and I haven't received the reviews necessary to merge.

PR #2215, opened 2023-10-03
Add new --volume-size-default install flag

tylerslaton:
This adds a new flag, `volume-size-default` that allows a user to specify what the default size for a Volume should be, agnostic of Volume Classes. It also sets the default for the `prod` profile to be `1Gi`.

### Checklist


pkg/apis/api.acorn.io/v1/types.go
```diff
@@ -546,6 +546,7 @@ type Config struct {
 	PropagateProjectAnnotations    []string        `json:"propagateProjectAnnotations" name:"propagate-project-annotation" usage:"The list of keys of annotations to propagate from acorn project to app namespaces"`
 	PropagateProjectLabels         []string        `json:"propagateProjectLabels" name:"propagate-project-label" usage:"The list of keys of labels to propagate from acorn project to app namespaces"`
 	ManageVolumeClasses            *bool           `json:"manageVolumeClasses" name:"manage-volume-classes" usage:"Manually manage volume classes rather than sync with storage classes, setting to 'true' will delete Acorn-created volume classes"`
+	VolumeSizeDefault              string          `json:"volumeSizeDefault" name:"volume-size-default" usage:"Set the default size for acorn volumes. Accepts storage suffixes (K, M, G, etc) and \".\" and \"_\" seperators (default 0)" short:"v"`
```
g-linville, line 549: Imo we should not have `-v` as short for this. I'm so used to thinking that `-v` means verbose, and for a very specific configuration parameter like this, I don't think it really makes sense to have a short version.

Also, this is more of a nit, but I would also add to the list of accepted suffixes `Mi` and `Gi` to indicate that the base 2 suffixes are supported as well.

tylerslaton: The -v was not intentional. That was copilot 😅

tylerslaton: Updated!

pkg/apis/api.acorn.io/v1/types.go
```diff
@@ -546,6 +546,7 @@ type Config struct {
 	PropagateProjectAnnotations    []string        `json:"propagateProjectAnnotations" name:"propagate-project-annotation" usage:"The list of keys of annotations to propagate from acorn project to app namespaces"`
 	PropagateProjectLabels         []string        `json:"propagateProjectLabels" name:"propagate-project-label" usage:"The list of keys of labels to propagate from acorn project to app namespaces"`
 	ManageVolumeClasses            *bool           `json:"manageVolumeClasses" name:"manage-volume-classes" usage:"Manually manage volume classes rather than sync with storage classes, setting to 'true' will delete Acorn-created volume classes"`
+	VolumeSizeDefault              string          `json:"volumeSizeDefault" name:"volume-size-default" usage:"Set the default size for acorn volumes. Accepts storage suffixes (K, M, G, Ki, Mi, Gi, etc) and \".\" and \"_\" seperators (default 0)"`
```
g-linville, line 549:
```suggestion
	VolumeSizeDefault              string          `json:"volumeSizeDefault" name:"volume-size-default" usage:"Set the default size for acorn volumes. Accepts storage suffixes (K, M, G, Ki, Mi, Gi, etc) and \".\" and \"_\" separators (default 0)"`
```

There are a couple other instance of this misspelling in the same file that should also be fixed. (It matters imo since this is user-facing documentation.)

g-linville (approved):

pkg/apis/api.acorn.io/v1/types.go
tylerslaton: I can fix it now, good catch :)

g-linville (approved):
Lgtm

pkg/controller/appdefinition/volume.go
```diff
@@ -200,7 +200,8 @@ func toPVCs(req router.Request, appInstance *v1.AppInstance) (result []kclient.O
 			pvc.Spec.VolumeName = pvName
 
 			if volumeRequest.Size == "" {
-				pvc.Spec.Resources.Requests[corev1.ResourceStorage] = *v1.DefaultSize
+				// TODO - should we do some sort of nil pointer validation here?
+				pvc.Spec.Resources.Requests[corev1.ResourceStorage] = *appInstance.Status.Defaults.VolumeSize
```
tylerslaton, line 204: Want to put a special highlight out for this change, thoughts?

g-linville: Like in the changelog? I'm not sure it's something that users would understand.

g-linville (approved):

pkg/controller/appdefinition/volume.go
tylerslaton: I meant that I want to talk about this line of code in GitHub. The code has actually changed since I wrote this comment since I had to update it to fix some tests. I'll resolve this and open another comment.

pkg/controller/appdefinition/volume.go
```diff
@@ -200,7 +201,14 @@ func toPVCs(req router.Request, appInstance *v1.AppInstance) (result []kclient.O
...
 			if volumeRequest.Size == "" {
-				pvc.Spec.Resources.Requests[corev1.ResourceStorage] = *v1.DefaultSize
+				// This shouldn't happen because we set a default in the status. However
+				// if this situation does occur, we'll just use the default size for runtime
+				// and log a warning. As an example, our unit tests hit this case.
+				if appInstance.Status.Defaults.VolumeSize == nil {
+					logrus.Debugf("no default volume size found in status, using static default size of %v", v1.DefaultSize)
+					appInstance.Status.Defaults.VolumeSize = v1.DefaultSize
+				}
+				pvc.Spec.Resources.Requests[corev1.ResourceStorage] = *appInstance.Status.Defaults.VolumeSize
```
tylerslaton, line 211: Does this make sense? Essentially if the `VolumeSize` is ever nil, we either have to error out or choose a default. In this case I chose to have a default default which feels strange 😓 

@g-linville @njhale

g-linville: Imo this is fine and makes sense. The comment helps a lot

pkg/controller/appdefinition/volume.go
```diff
@@ -200,7 +201,14 @@ func toPVCs(req router.Request, appInstance *v1.AppInstance) (result []kclient.O
 			pvc.Spec.VolumeName = pvName
 
 			if volumeRequest.Size == "" {
-				pvc.Spec.Resources.Requests[corev1.ResourceStorage] = *v1.DefaultSize
+				// This shouldn't happen because we set a default in the status. However
+				// if this situation does occur, we'll just use the default size for runtime
+				// and log a warning. As an example, our unit tests hit this case.
+				if appInstance.Status.Defaults.VolumeSize == nil {
+					logrus.Debugf("no default volume size found in status, using static default size of %v", v1.DefaultSize)
```
cjellick, line 208: comment says log a warning. you are logging at debug. just wanted to check which one you want to do

tylerslaton: Oops. Definitely debug. Will update.

cjellick (changes requested):
just one small comment

cjellick (approved):

PR #2218, opened 2023-10-04
Re-evaluate image permissions for failing apps

thedadams:
When an IRA is added or updated, this change will allow apps with failing image permissions to be re-evaluated in case the IRA now allows the app to run.


njhale (approved):

pkg/controller/permissions/permissions_check.go
```diff
@@ -44,105 +43,103 @@ func CopyPromoteStagedAppImage(req router.Request, resp router.Response) error {
...
-		app := req.Object.(*v1.AppInstance)
-
-		iraEnabled, err := config.GetFeature(req.Ctx, req.Client, profiles.FeatureImageRoleAuthorizations)
-		if err != nil {
-			return err
-		}
-		if !iraEnabled {
-			app.Status.Staged.ImagePermissionsDenied = nil
-		}
+func CheckPermissions(req router.Request, _ router.Response) error {
```
iwilltry42, line 46: I think we may have kept the transport here and passed it on to checkImageAllowed and also to getAuthorizedPermissions, since it's useful as a remoteOpts for the signature checks.
Just seeing that you dropped it reminded me, that I didn't use it here :roll_eyes:

thedadams: Please add it back if you think it is necessary. Since things were working and we weren't using it, I removed it.

pkg/controller/routes.go
```diff
@@ -134,6 +135,9 @@ func routes(router *router.Router, cfg *rest.Config, registryTransport http.Roun
 	router.Type(&appsv1.Deployment{}).Namespace(system.ImagesNamespace).HandlerFunc(networkpolicy.ForBuilder)
 	router.Type(&netv1.NetworkPolicy{}).Selector(managedSelector).HandlerFunc(gc.GCOrphans)
 
+	router.Type(&internaladminv1.ImageRoleAuthorizationInstance{}).HandlerFunc(permissions.BumpImageRoleAuthorizations)
+	router.Type(&internaladminv1.ClusterImageRoleAuthorizationInstance{}).HandlerFunc(permissions.BumpClusterImageRoleAuthorizations)
```
iwilltry42, line 139: Will those only get triggered upon create/update/delete?

thedadams: The handler will get triggered when an IRA is created/updated/deleted or when the controller starts up. However, if the IRA hasn't changed (the observed generation is the same as the generation) or the app doesn't have any denied permissions, then we won't do anything.

PR #2226, opened 2023-10-06
change: move actual permissions to .status + apply IRAs to consumed permissions

iwilltry42:
This PR introduces quite a few changes around permissions:

- Some comments and renaming to make permissions stuff clearer
- The actually granted and authorized permissions for every app will now go to AppInstances' `.status.permissions` field and contain **only** those permissions required for this specific AppInstance, **not** for nested Acorns
- Permissions are promoted from `.spec` to `.status` along with the staged AppImage when all checks (including ImageRoleAuthorizations) pass
- Permissions consumed from nested Services are added after all that and thus we have a new handler for them that only allows the consuming App to run if it's authorized to hold the permissions consumed from the producing service
  - Those will be merged into the `.status.permissions` field such that this field reflects truly all permissions given to that AppInstance

### Checklist


integration/client/imagerules/imageroleauthorizations_test.go
```diff
@@ -363,4 +364,227 @@ func TestImageRoleAuthorizations(t *testing.T) {
...
+	})
+
+	serviceAcorn, err := c.AppGet(ctx, app.Status.AppStatus.Services["producer"].ServiceAcornName)
+	require.NoError(t, err, "should not error while getting service acorn")
+
+	serviceAcorn = helper.WaitForObject(t, helper.Watcher(t, c), &apiv1.AppList{}, serviceAcorn, func(obj *apiv1.App) bool {
+		return obj.Status.Staged.PermissionsChecked
+	})
+
+	require.Equal(t, 1, len(serviceAcorn.Status.Staged.ImagePermissionsDenied), "should have 0 denied permissions")
```
g-linville, line 479: Bit confused by this. Shouldn't it be requiring that the len of ImagePermissionsDenied is 0, not 1? Or is the error message incorrect?

pkg/controller/appdefinition/depends.go
```diff
@@ -56,7 +56,9 @@ func getDependencyAnnotations(app *v1.AppInstance, containerOrJobName string, de
 		}
 	}
 
-	if !allReady {
+	consumerpermsok := len(app.Status.DeniedConsumerPermissions) == 0
+
+	if !allReady || !consumerpermsok {
```
g-linville, line 61: nit:

```suggestion
	consumerPermsOk := len(app.Status.DeniedConsumerPermissions) == 0

	if !allReady || !consumerPermsOk {
```

pkg/controller/appdefinition/pullappimage.go
```diff
@@ -29,6 +29,7 @@ type pullClient struct {
 	pull pullImageFunc
 }
 
+// pullAppImage
```
g-linville, line 32: nit: remove if not needed

pkg/controller/permissions/consumerpermissions.go
```diff
@@ -0,0 +1,157 @@
...
+	for jobName, jobDef := range app.Status.AppSpec.Jobs {
+		consumerPerms, err := GetConsumerPermissions(ctx, client, app, jobName, jobDef)
+		if err != nil {
+			return nil, err
+		}
+		result = append(result, consumerPerms)
+	}
+
+	return result, nil
+}
```
g-linville, line 71: Have you verified that sidecars are also covered by this? It is possible that the Dependencies field on a container object will include the deps of its sidecars, but idk for sure whether this is the case.

pkg/controller/permissions/consumerpermissions.go
```diff
@@ -0,0 +1,157 @@
...
+
+		if len(denied) > 0 {
+			app.Status.DeniedConsumerPermissions = denied
+			return nil
+		}
+	}
+
+	app.Status.DeniedConsumerPermissions = nil
+	app.Status.Permissions = v1.SimplifySet(append(app.Status.Permissions, newPerms...))
+	return nil
```
g-linville, line 156: In a few spots here, the app status is modified and then we `return nil`. Don't we need to pass this to `resp.Objects` or update it with the client or something in order for the changes to stick?

pkg/controller/permissions/permissions_check.go
```diff
@@ -104,6 +114,22 @@ func CheckPermissions(req router.Request, _ router.Response) error {
 			details.AppImage.Digest, appImage.Digest)
 	}
 
+	// ServiceNames of the current app level (i.e. not nested Acorns/Services)
+	scvnames := maps.Keys(details.AppSpec.Containers)
+	scvnames = append(scvnames, maps.Keys(details.AppSpec.Jobs)...)
+	scvnames = append(scvnames, maps.Keys(details.AppSpec.Services)...)
```
g-linville, line 120: Do we need to add sidecar names to this as well? I believe the "service name" (as far as permissions are concerned) for sidecars is in the format of `<parent container name>.<sidecar name>`.

g-linville (commented):
This is my first pass. Gonna have to wrap my head around this some more, but these are my initial thoughts.

integration/client/imagerules/imageroleauthorizations_test.go
iwilltry42: Yeah, the error message was wrong :D

pkg/controller/permissions/consumerpermissions.go
iwilltry42: Good point!
I was checking sidecars explicitly before, but they may have been dropped during another refactoring...
I will double check this and add a test case :+1:

pkg/controller/permissions/consumerpermissions.go
iwilltry42: They just stick without anything else to be done :grin: 
I'm not knees deep in the details there, as that's just what I did before without questioning it, but I assume that the request object is always part of the response :+1:

pkg/controller/permissions/consumerpermissions.go
iwilltry42: EDIT: Actually there's no change to the old logic here. We always only considered `app.Status.AppSpec.Jobs|Containers` - still I'll verify that again.

iwilltry42: Checked again and apparently, `dependsOn` in the sidecar scope within the Acornfile doesn't have any effect :thinking:

pkg/controller/permissions/permissions_check.go
iwilltry42: Permissions set on sidecars are "assigned" to the parent container :+1:

pkg/controller/permissions/permissions_check.go
```diff
@@ -130,13 +156,18 @@ func CheckPermissions(req router.Request, _ router.Response) error {
 			return nperms
 		}
 
-		denied, _ := v1.GrantsAll(app.Namespace, copyWithName(details.Permissions, imageName), authzPerms)
+		deniedPerms := []v1.Permissions{}
```
thedadams, line 159: nit:
```suggestion
		var deniedPerms []v1.Permissions
```

pkg/controller/permissions/consumerpermissions.go
```diff
@@ -0,0 +1,157 @@
...
+		if svc.Spec.Consumer.Permissions != nil {
+			result.Rules = append(result.Rules, svc.Spec.Consumer.Permissions.GetRules()...)
+		}
+	}
+
+	return
+}
+
+func collectConsumerPermissions(ctx context.Context, client kclient.Client, app *v1.AppInstance) ([]v1.Permissions, error) {
+	result := []v1.Permissions{}
```
thedadams, line 52: same nit:
```suggestion
	var result []v1.Permissions
```

pkg/controller/permissions/consumerpermissions.go
```diff
@@ -0,0 +1,157 @@
...
+	app := req.Object.(*v1.AppInstance)
+
+	iraEnabled, err := config.GetFeature(req.Ctx, req.Client, profiles.FeatureImageRoleAuthorizations)
+	if err != nil {
+		return err
+	}
+	if !iraEnabled {
+		// Reset status field if IRAs are disabled
+		app.Status.DeniedConsumerPermissions = nil
+	}
```
thedadams, line 120: Looks like this may be unnecessary and is handled by the code below.

thedadams (approved):

pkg/controller/permissions/consumerpermissions.go
thedadams: I remember a conversation where the fact that `dependsOn` in sidecars were ignored.

g-linville (approved):

PR #2240, opened 2023-10-12
Fix an issue with default VolmeStorage not being accounted for in QuotaRequests

tylerslaton:
Before this PR, we were not accounting for the scenario that the Volume will receive a default size. This caused a number of issues regarding approval of resources.

### Checklist

### DIFF
--- a/pkg/controller/quota/quota.go
+++ b/pkg/controller/quota/quota.go
@@ -146,16 +146,25 @@ func addStorage(appInstance *v1.AppInstance, quotaRequest *adminv1.QuotaRequestI
 			size = boundSize
 		}
 
-		// No need to proceed if the size is empty
-		if size == "" || size == "0" {
+		// Handle three cases:
+		// 1. The volume's size is explicitly set to 0. This means there is nothing to count.
+		// 2. The volume's size is not set. This means we should assume the default size.
+		// 3. The volume's size is set to a specific value. This means we should use that value.
+		var sizeQuantity resource.Quantity
+		switch size {
+		case "0":
 			continue
+		case "":
+			sizeQuantity = defaultVolumeSize(appInstance, name)
+		default:
+			parsedQuantity, err := resource.ParseQuantity(string(size))
+			if err != nil {
+				return err
+			}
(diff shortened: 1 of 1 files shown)


pkg/controller/quota/quota.go
```diff
@@ -169,6 +178,25 @@ func addStorage(appInstance *v1.AppInstance, quotaRequest *adminv1.QuotaRequestI
...
+	result := *v1.DefaultSize // Safe to dereference because it is statically set in the v1 package.
+
+	// If the volume has a default size set, use that on status.Defaults, use that. Otherwise, if the
+	// VolumeSize default is set on status.Defaults, use that.
+	if defaultVolume, set := appInstance.Status.Defaults.Volumes[name]; set {
+		// We do not expect this to ever fail because VolumeClasses have their sizes validated. However,
+		// if it does fail, we'll just return the default size.
+		parsedQuantity, err := resource.ParseQuantity(string(defaultVolume.Size))
+		if err == nil {
+			result = parsedQuantity
```
tylerslaton, line 191: This felt like the right behavior to me if a size failed to parse, but let me know if something else makes more sense.

pkg/controller/quota/quota.go
```diff
@@ -146,16 +146,25 @@ func addStorage(appInstance *v1.AppInstance, quotaRequest *adminv1.QuotaRequestI
...
 		}
 
-		// No need to proceed if the size is empty
-		if size == "" || size == "0" {
+		// Handle three cases:
+		// 1. The volume's size is explicitly set to 0. This means there is nothing to count.
+		// 2. The volume's size is not set. This means we should assume the default size.
+		// 3. The volume's size is set to a specific value. This means we should use that value.
+		var sizeQuantity resource.Quantity
+		switch size {
```
thedadams, line 154: It looks like the size of bound volumes are counted here. Is there any risk that we are double-counting bound volumes? Or are we saying you can have as many volumes as you want as long as none of them are bound?

thedadams (approved):

njhale (approved):

Issue #2192, opened 2023-09-29
Limit list of authorized characters in basic secret


lucj:
Currently we can specify the characters we want to be used in a secret of type token, the example below is from the doc:

```
secret: {
    password: {
        type: "token"
        params: {
            length: 16
            characters: "abcdedfhifj01234567890"
        }
        data: token: ""
    }
}
```

That should be great to have the same behavior for secret of type basic as well. Basic secret are great to create database username/password on the fly but because the chars list does not prevent special chars it cannot be used to generate postgres username/password for instance.

I think that could be great if the chars restriction is available for basic secret as well.

lucj:
Just an additional note ;)
Could the list of characters be also represented with a-z, A-Z, 0-9 so we can use "a-zA-Z0-9" for aphanumerical characters ?

njhale:
Merged the change to parameterize basic auth secrets, will be adding the `"a-zA-Z0-9"` enhancement in a follow-up!

sangee2004:
Tested with acorn version - `v0.9.0-rc1-31-gbdd8d407+bdd8d407`

Able to limit list of authorized character for basic secret's username and password.

1. Deployed secrets with type `basic`:
```
secrets: {
    password: {
        type: "basic"
        params: {
            usernameLength: 16
            usernameCharacters: "abcdedfhifj01234567890"
            passwordLength: 16
            passwordCharacters: "abcdedfhifj01234567890"
        }
    }
}
```

This results in secrets being created with the provided size and character set for both username and password fields.
```
acorn secrets reveal testnew.password 
NAME               TYPE      KEY        VALUE
testnew.password   basic     password   0eb0fa93fehhf7e8
testnew.password   basic     username   04cadd2d5dd6fead
```

2. Another example with more restricted character set and different ones for username and password fields also work as expected.
```
secrets: {
    password: {
        type: "basic"
        params: {
            usernameLength: 10
            usernameCharacters: "xyz123"
            passwordLength: 7
            passwordCharacters: "atyu9"
        }
    }
}
```

```
acorn secret reveal testres.password  
NAME               TYPE      KEY        VALUE
testres.password   basic     password   a9tytyy
testres.password   basic     username   313z2323yz
```

njhale:
Just merged support for "character range" syntax sugar: #2250

sangee2004:
Tested with Acorn version - `v0.9.0-rc1-37-g2be8a309+2be8a309`

[...]

PR #2242, opened 2023-10-12
Parameterize basic secret generation (#2192)

njhale:
Enable users to specify character set and length for username
and password fields of generated basic auth secrets.

Addresses #2192


pkg/secrets/generateSecret.go
```diff
@@ -5,27 +5,31 @@ import (
...
-func GenerateRandomSecret(length int) (string, error) {
-	const (
-		uppercase = "ABCDEFGHJKLMNPQRSTUVWXYZ"
-		lowercase = "abcdefghijkmnopqrstuvwxyz"
-		numbers   = "23456789"
-		special   = "!#$%^&*_-=+"
-	)
+const (
+	defaultLength       = 54
+	defaultCharacterSet = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789!#$%^&*_-=+"
```
g-linville, line 10: nit: do we want to remove the following characters from this list?

`IO01`

In the old code, they were not present, probably due to their similarity.

pkg/controller/secrets/secret_test.go
```diff
@@ -108,8 +110,8 @@ func TestBasic_Gen(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	assert.Len(t, resp.Client.Created, 2)
-	assert.Len(t, resp.Collected, 2)
+	require.Len(t, resp.Client.Created, 2)
```
thedadams, line 113: Asking because I don't know: why is this better?

pkg/secrets/secret.go
```diff
@@ -249,15 +249,40 @@ func generateBasic(req router.Request, appInstance *v1.AppInstance, secretName s
...
+			charactersKey: "passwordCharacters",
+		},
+	} {
+		if len(secretRef.Data[keys.dataKey]) > 0 {
+			// Explicitly set by user, don't generate
+			continue
+		}
+
+		var length int64
+		if lengthParam, ok := secretRef.Params.GetData()[keys.lengthKey]; ok {
```
g-linville, line 272: This doesn't handle the `!ok` case. Is it possible for that to happen?

pkg/secrets/generateSecret.go
thedadams, line 1: nit: the name of this file could be updated.

thedadams (approved):

pkg/controller/secrets/secret_test.go
njhale: This fails the testcase before it gets into an index out-of-bounds issue in a check below.

pkg/secrets/generateSecret.go
njhale: good call, changing

pkg/secrets/generateSecret.go
njhale: Nice eye, copilot decided it would fill in the gaps for me I guess. I'll drop them

pkg/secrets/secret.go
njhale: yes it is. In the case of `!ok`, the length is `0`, which triggers the default value to be used in `GenerateRandomSecret`

pkg/secrets/generateSecret.go
njhale: On second thought, there seem to be more letters missing than that; the omissions are also not consistent between lower and upper case. I wonder if this was just an oversight initially

g-linville (approved):

PR #2250, opened 2023-10-16
Add range syntax sugar for secret generation

njhale:
Add the ability to specify ranges "A-Z", "a-z", and "0-9" when providing
arguments for basic and token type secrets.

### DIFF
--- a/pkg/secrets/generate_test.go
+++ b/pkg/secrets/generate_test.go
@@ -1,0 +1,83 @@
+package secrets
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestInflateRanges(t *testing.T) {
+	for _, tt := range []struct {
+		name string
+		want string
+	}{
+		{
+			name: "",
+			want: "",
+		},
+		{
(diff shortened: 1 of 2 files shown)


pkg/secrets/generate_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+func TestInflateRanges(t *testing.T) {
+	for _, tt := range []struct {
+		name string
+		want string
+	}{
+		{
+			name: "A-Z",
+			want: "ABCDEFGHIJKLMNOPQRSTUVWXYZ",
+		},
```
tylerslaton, line 17: Can we add a test case for empty ranges being passed too? Mainly to make certain we don't have any chance of hitting an out of bounds panic.

tylerslaton (approved):

pkg/secrets/generate_test.go
njhale: done

StrongMonkey (approved):

tylerslaton (approved):

PR #2254, opened 2023-10-17
Ensure that the status.Defaults.VolumeSize is used if a binding has no size specified

tylerslaton:
This fixes an issue where, if a binding had nothing set in `appSpec.Volumes` the static size would always be used instead of `status.Defaults.VolumeSize`.

### Checklist

### DIFF
--- a/pkg/controller/quota/quota.go
+++ b/pkg/controller/quota/quota.go
@@ -193,20 +193,23 @@ func addStorage(appInstance *v1.AppInstance, quotaRequest *adminv1.QuotaRequestI
 	return nil
 }
 
+// defaultVolumeSize determines the default size of the specified volume. If the volume has a default size set
+// on the status.Defaults.Volumes, it uses that. Otherwise, it uses the default size set on the status.Defaults.VolumeSize.
 func defaultVolumeSize(appInstance *v1.AppInstance, name string) resource.Quantity {
+	// Figure out what the default volume size should be. If the status has a default volume size set, use that. Otherwise,
+	// use the default size set in the v1 package.
 	result := *v1.DefaultSize // Safe to dereference because it is statically set in the v1 package.
+	if appInstance.Status.Defaults.VolumeSize != nil {
+		result = *appInstance.Status.Defaults.VolumeSize
+	}
 
-	// If the volume has a default size set, use that on status.Defaults, use that. Otherwise, if the
-	// VolumeSize default is set on status.Defaults, use that.
+	// If the volume has a default size set on status.Defaults.Volumes, use that.
 	if defaultVolume, set := appInstance.Status.Defaults.Volumes[name]; set {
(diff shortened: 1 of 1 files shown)


pkg/controller/quota/quota.go
```diff
@@ -193,20 +193,23 @@ func addStorage(appInstance *v1.AppInstance, quotaRequest *adminv1.QuotaRequestI
...
 }
 
+// defaultVolumeSize determines the default size of the specified volume. If the volume has a default size set
+// on the status.Defaults.Volumes, it uses that. Otherwise, it uses the default size set on the status.Defaults.VolumeSize.
 func defaultVolumeSize(appInstance *v1.AppInstance, name string) resource.Quantity {
+	// Figure out what the default volume size should be. If the status has a default volume size set, use that. Otherwise,
+	// use the default size set in the v1 package.
 	result := *v1.DefaultSize // Safe to dereference because it is statically set in the v1 package.
+	if appInstance.Status.Defaults.VolumeSize != nil {
+		result = *appInstance.Status.Defaults.VolumeSize
```
thedadams, line 203: I would like to see this included as part of `appInstance.Status.Defaults.Volumes` instead of it being its own field. Otherwise, this field is going to need to be checked everywhere we are trying to determine the default.

tylerslaton: That would require some migration of apps going to `Defaults.VolumeSize` => `Defaults.Volumes[volume].Size`. Do you want that to start with this PR?

pkg/controller/quota/quota.go
```diff
@@ -193,20 +193,23 @@ func addStorage(appInstance *v1.AppInstance, quotaRequest *adminv1.QuotaRequestI
...
 	if defaultVolume, set := appInstance.Status.Defaults.Volumes[name]; set {
 		// We do not expect this to ever fail because VolumeClasses have their sizes validated. However,
-		// if it does fail, we'll just return the default size.
-		parsedQuantity, err := resource.ParseQuantity(string(defaultVolume.Size))
-		if err == nil {
+		// if it does fail, we'll just use the default size instead.
+		if parsedQuantity, err := resource.ParseQuantity(string(defaultVolume.Size)); err == nil {
 			result = parsedQuantity
 		}
-	} else if appInstance.Status.Defaults.VolumeSize != nil {
```
thedadams, line 208: Why didn't this logic work? Is it because `set` was `true`, but `defaultVolume` was `""`?

tylerslaton: Yeah, exactly. In that case we wouldn't hit the next else if statement which in turn wouldn't set the default from Defaults.VolumSize. This PR changes that so we calculate the "default default" before checking what's written in `Defaults.Volumes`.

drpebcak (approved):

njhale (approved):

PR #2255, opened 2023-10-17
Fix edge case in range syntax sugar

njhale:
Treat mixed case and backwards ranges (e.g. `A-z`, `0-A`, `z-a`) as
individual characters in the given set. This simplifies the UX and
makes ranges easier to reason about.

### DIFF
--- a/pkg/secrets/generate.go
+++ b/pkg/secrets/generate.go
@@ -4,6 +4,7 @@ import (
 	"crypto/rand"
 	"math/big"
 	"sort"
+	"unicode"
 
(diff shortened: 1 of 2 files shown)


njhale:
Just thought of some problems with this fix. Fixing the fix now.

StrongMonkey (approved):

tylerslaton (approved):

keyallis (approved):
Would love to have a chat about the idea of a removelist for runes rather than defining the specific ranges we'd like to support

PR #2261, opened 2023-10-18
Fix issue where implicitly bound volumes were counted twice for QuotaRequests

tylerslaton:
If an incoming app's name matches both a volume and that volume's creating app, they will automatically bind. This logic was not previously accounted for in the current QuotaRequest counting logic. This PR makes the `LookupExistingPV` function public and utilizes it to determine if a Volume will go through that implicit binding process. In addition I've added a unit test that explicitly tests this situation. Since there weren't tests for other basic cases, I bundled those into this PR as well.

### Checklist


pkg/controller/quota/quota.go
```diff
@@ -161,6 +162,12 @@ func addStorage(appInstance *v1.AppInstance, quotaRequest *adminv1.QuotaRequestI
 			size = boundSize
 		}
 
+		// If the volume will be bound to an existing PV implicitly, then we don't need to count it. This will happen
+		// when the incoming app has the same name as the app that created an existing released PV.
+		if pvName, err := appdefinition.LookupExistingPV(req, appInstance, name); err == nil && pvName != "" {
+			continue
+		}
```
thedadams, line 169: If `LookupExistingPV` errors, would you want to error here? I am guessing that if `LookupExistingPV` returns an error, then we aren't able to determine if we should count the volume again the quota, so may want to retry?

tylerslaton: You bring up a good point. I did this intentionally but looking back, the current behavior is wrong. Good catch.

thedadams (approved):

njhale (approved):

PR #2262, opened 2023-10-18
Do not allow a user to exec into a container unless they have granted perms

ibuildthecloud:
We check the permissions granted to this acorn and ensure that the user
that is running exec has all of the permissions granted to the image. If
they do not, we return error. This is intended to prevent privilege
escalation of users by means of using exec and service accounts.

### DIFF
--- a/pkg/server/registry/apigroups/acorn/apps/validator.go
+++ b/pkg/server/registry/apigroups/acorn/apps/validator.go
@@ -51,17 +51,29 @@ var (
 type Validator struct {
 	client            kclient.Client
 	clientFactory     *client.Factory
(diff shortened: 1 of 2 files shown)


pkg/server/registry/apigroups/acorn/apps/validator.go
```diff
@@ -51,17 +51,29 @@ var (
 type Validator struct {
 	client            kclient.Client
 	clientFactory     *client.Factory
-	deleter           strategy.Deleter
+	getter            strategy.Getter
```
ibuildthecloud, line 54: I have no idea why this was deleter, it was used as a getter.

PR #2282, opened 2023-10-24
fix: close orphaned grpc connections in dev mode and silence related logs

iwilltry42:
Ref: https://github.com/acorn-io/manager/issues/1211

Upstream fix: https://github.com/devspace-sh/devspace/pull/2739

The upstream fix resolves the issue of having lots of dangling, unused grpc connections.
But even with the fix, we'd see at least 2 lines of the annoying logs coming from the PING connection check right before closing the connection - that's why we silence grpc info/warn logs as well.

### Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -41,7 +41,7 @@ require (
 	github.com/hexops/autogold/v2 v2.2.1
 	github.com/hexops/valast v1.4.4
 	github.com/liggitt/tabwriter v0.0.0-20181228230101-89fcab3d43de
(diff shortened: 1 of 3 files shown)


thedadams (approved):

iwilltry42:
> Does this suppress all of the logs from the package? Are we losing any logs that we intend on being able to view by making this change?

Warning and Info only. Darren just wanted to not see any of those, but I left the error/fatal logs on, since we didn't face those so far.

PR #2297, opened 2023-10-26
Only set Status.Defaults.Volumes if it hasn't been set

tylerslaton:
If the Volume already has defaults, skip it. We don't want to overwrite defaults for volumes values as it can lead to unexpected behavior when volume classes are updated. One example is a volume class going down in size, which in turn will cause the volume to be sized down and likely go into an error state on the next app update.

This also includes a migration step away from Status.Defaults.VolumeSize. It removes the logic that originally set the VolumeSize field and instead hands that logic off to the `Status.Defaults.Volumes` map field. If VolumeSize was previously set, then we will set that in the Volumes map field as the default. This prevents any issues from occuring with existing volumes when the default VolumeClass is updated.

### Checklist


pkg/controller/defaults/testdata/volumeclass/volume-class-defaults-set/input.yaml
```diff
@@ -0,0 +1,26 @@
...
+      container-name:
+        image: "image-name"
+        dirs:
+          "/var/tmp":
+            volume: foo
+  defaults:
+    volumes:
+      foo:
+        class: test-volume-class
+        size: 4Gi
```
tylerslaton, line 26: In this particular test instance, we expect this to remain 4Gi after going through a reconciliation.

g-linville (approved):

g-linville (approved):

pkg/controller/defaults/volumeclass.go
```diff
@@ -54,3 +76,24 @@ func addVolumeClassDefaults(ctx context.Context, c kclient.Client, app *v1.AppIn
 
 	return nil
 }
+
+func getDefaultVolumeSize(ctx context.Context, c kclient.Client, appInstance *v1.AppInstance) (v1.Quantity, error) {
+	// If the Status.Defaults.VolumeSize has been set, use that.
+	if appInstance.Status.Defaults.VolumeSize != nil {
+		return v1.Quantity(appInstance.Status.Defaults.VolumeSize.String()), nil
+	}
+
```
thedadams, line 85: nit: this function is only used when we know that the `VolumeSize == nil`. Therefore, this check is redundant. Maybe not that big of a deal because, based on other comments here, `VolumeSize` is going to be removed.

tylerslaton: Yep, that's the idea. Right now it is not longer being set in this PR. I still want it in the schema to allow old apps to move over and then I'll remove it at that point.

thedadams (approved):

pkg/controller/defaults/volumeclass.go
tylerslaton: Update - I did go ahead and remove this check though as I agree with your assessment.

tylerslaton:
@thedadams @g-linville I was giving this a test before merging and noticed some behavior that was not quite right, mainly around the migration. I've gone ahead and updated this PR to fix that issue. Please let me know if I can answer any questions. Re-requesting review for the time being.

g-linville (approved):

pkg/controller/defaults/volumeclass.go
```diff
@@ -36,21 +37,58 @@ func addVolumeClassDefaults(ctx context.Context, c kclient.Client, app *v1.AppIn
...
+			}
 		}
 
 		app.Status.Defaults.Volumes[name] = volDefaults
 	}
 
 	return nil
 }
+
+func getDefaultVolumeSize(ctx context.Context, c kclient.Client, appInstance *v1.AppInstance) (v1.Quantity, error) {
```
g-linville, line 80: nit: no need for the `appInstance` parameter here

pkg/controller/defaults/testdata/volumeclass/volume-class-defaults-same-gen/expected.golden
```diff
@@ -30,13 +30,14 @@ status:
   appStatus: {}
   columns: {}
   conditions:
+  - error: true
+    message: 'cannot establish defaults because two defaults volume classes exist:
+      test-volume-class and test-volume-class-1'
```
thedadams, line 35: Is this the desired outcome for this test?

pkg/controller/defaults/volumeclass.go
```diff
@@ -36,21 +37,58 @@ func addVolumeClassDefaults(ctx context.Context, c kclient.Client, app *v1.AppIn
...
 	}
 
 	return nil
 }
+
+func getDefaultVolumeSize(ctx context.Context, c kclient.Client) (v1.Quantity, error) {
+	cfg, err := config.Get(ctx, c)
+	if err != nil {
+		return "", err
+	}
```
thedadams, line 84: nit: don't get the config unless you have to.

pkg/controller/defaults/volumeclass.go
```diff
@@ -36,21 +37,58 @@ func addVolumeClassDefaults(ctx context.Context, c kclient.Client, app *v1.AppIn
...
+			if vol.Size == "" {
+				if volumeClasses[vol.Class].Size.Default == "" {
+					defaultSize, err := getDefaultVolumeSize(ctx, c)
+					if err != nil {
+						return err
+					}
+					volDefaults.Size = defaultSize
+				} else {
+					volDefaults.Size = volumeClasses[vol.Class].Size.Default
+				}
```
thedadams, line 70: nit: just to dry it up.
```suggestion
				volDefaults.Size = volumeClasses[vol.Class].Size.Default
				if volDefaults.Size == "" {
					defaultSize, err := getDefaultVolumeSize(ctx, c)
					if err != nil {
						return err
					}
					volDefaults.Size = defaultSize
				}
```

pkg/controller/defaults/testdata/volumeclass/volume-class-defaults-same-gen/expected.golden
tylerslaton: In this case, yes. We are not longer only processing the defaults when the generation is different. Instead we are processing it always if a default has not been set for this Volume. As such this test is working as expected given that the input had two default VolumeClasses set this causing a conflict. That might be too confusing of a test though and this could instead be replaced with something more inline with what we expect to happen

thedadams: Is that the desired behavior? My understanding was we don't want to potentially break apps unless a user took an explicit action.

In other words, I understand that we only want to set the volume default once, but should that only happen once an existing app is updated?

pkg/controller/defaults/volumeclass.go
tylerslaton: I might be missing something, let me know. However, I am using this config in the if statement below to determine if a default was set on the Acorn config.

pkg/controller/defaults/testdata/volumeclass/volume-class-defaults-same-gen/expected.golden
tylerslaton: Talked offline but to recap - this is a temporary thing that is necessary while we're migrating away from VolumeSize. The logic is essentially: if VolumeSize is set, make sure it is set on the Volumes map. This needs to happen at least once and with it being behind the generation would mean a user would have to explicitly update their app to get that Default set on the Volumes map. Once this merges and propagates we will be able to put it back behind the generation like we'd expect.

thedadams (approved):

PR #2298, opened 2023-10-26
Add pre-stop hooks for containers that expose ports

thedadams:
This pre-stop hook will simply wait for 5 seconds. In order to achieve this in the most general way, a configmap is mounted as a volume that has a very simply binary that sleeps for 5 seconds. It is done this way because this should work for any container images, including scratch.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,7 @@ FROM ghcr.io/acorn-io/images-mirror/tonistiigi/binfmt:qemu-v6.2.0 AS binfmt
 FROM ghcr.io/acorn-io/images-mirror/moby/buildkit:v0.11.6 AS buildkit
 FROM ghcr.io/acorn-io/images-mirror/registry:2.8.1 AS registry
 FROM ghcr.io/acorn-io/images-mirror/rancher/klipper-lb:v0.3.5 AS klipper-lb
+FROM ghcr.io/acorn-io/sleep:latest AS sleep
 
 FROM ghcr.io/acorn-io/images-mirror/golang:1.21-alpine AS helper
 WORKDIR /usr/src
@@ -20,7 +21,8 @@ RUN --mount=type=cache,target=/go/pkg --mount=type=cache,target=/root/.cache/go-
 FROM ghcr.io/acorn-io/images-mirror/golang:1.21 AS build
 COPY / /src
 WORKDIR /src
-RUN --mount=type=cache,target=/go/pkg --mount=type=cache,target=/root/.cache/go-build make build
+COPY --from=sleep /sleep /src/pkg/controller/appdefinition/embed/acorn-sleep
+RUN --mount=type=cache,target=/go/pkg --mount=type=cache,target=/root/.cache/go-build GO_TAGS=netgo,image make build
 
 FROM ghcr.io/acorn-io/images-mirror/nginx:1.23.2-alpine AS base
 RUN apk add --no-cache ca-certificates iptables ip6tables fuse3 git openssh pigz xz \
--- a/pkg/controller/appdefinition/deploy.go
+++ b/pkg/controller/appdefinition/deploy.go
@@ -2,6 +2,7 @@ package appdefinition
 
 import (
 	"crypto/sha256"
+	_ "embed"
 	"encoding/hex"
 	"encoding/json"
 	"errors"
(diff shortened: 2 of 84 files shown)


Dockerfile
```diff
@@ -4,6 +4,7 @@ FROM ghcr.io/acorn-io/images-mirror/tonistiigi/binfmt:qemu-v6.2.0 AS binfmt
 FROM ghcr.io/acorn-io/images-mirror/moby/buildkit:v0.11.6 AS buildkit
 FROM ghcr.io/acorn-io/images-mirror/registry:2.8.1 AS registry
 FROM ghcr.io/acorn-io/images-mirror/rancher/klipper-lb:v0.3.5 AS klipper-lb
+FROM docker.io/thedadams/sleep:five AS sleep
```
thedadams, line 7: This needs to be mirrored, at least.

thedadams:
Putting this up for review. I tested it with a `scratch` image and it worked. I'm still in shock.

thedadams:
https://github.com/acorn-io/manager/pull/1517 should be merged first so we can use the mirrored image.

pkg/controller/appdefinition/deploy.go
```diff
@@ -636,18 +657,19 @@ func getSecretAnnotations(req router.Request, appInstance *v1.AppInstance, conta
 func toDeployment(req router.Request, appInstance *v1.AppInstance, tag name.Reference, name string, container v1.Container, pullSecrets *PullSecrets, interpolator *secrets.Interpolator) (*appsv1.Deployment, error) {
 	var (
 		stateful = isStateful(appInstance, container)
+		addWait  = len(acornSleepBinary) > 0 && len(container.Ports) > 0 && z.Dereference(container.Scale) > 1 && !stateful && !appInstance.Status.GetDevMode()
```
thedadams, line 660: We also want to consider sidecar ports here. I will make an adjustment.

Dockerfile
njhale: Should we be maintaining a GitHub repo and build action for this image? That's assuming we're building this sleep binary from source somewhere.

pkg/controller/appdefinition/sleep_image.go
```diff
@@ -0,0 +1,8 @@
+//go:build image
+
+package appdefinition
+
+import _ "embed"
+
+//go:embed embed/acorn-sleep
```
njhale, line 7: Is it possible for an app's underlying pods to get scheduled to a node with a different arch than the one the runtime controller is running on? I'm not sure if we even support multi-arch setups, but if we do, this could be an issue.

g-linville (approved):

pkg/controller/appdefinition/sleep_image.go
thedadams: Yeah, we currently don't support multiple architectures. We'll have to cross that bridge when we come to it.

pkg/controller/appdefinition/testdata/deployspec/pre-stop/stateful/input.yaml
```diff
@@ -0,0 +1,27 @@
...
+  namespace: app-created-namespace
+  appImage:
+    id: test
+  appSpec:
+    containers:
+      container-name:
+        image: "image-name"
+        dirs:
+          "/var/tmp":
+            volume: foo
```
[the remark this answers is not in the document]
thedadams: Good call. I updated the test.

Dockerfile
thedadams: I made a repo for this.

thedadams:
In addition to the above mentioned changes, I:
1. Made it so that the `ConfigMap` is only stored in memory once for apps with multiple containers that needed it. That is, the `ConfigMap` was only created once, but it was stored in memory for the apply multiple times.
2. Changed the location of the binary from `/usr/local/bin` to `/.acorn/sleep`
3. The repo I created for the sleep binary gets the size of the binary down to 5K for amd64 and 7k for arm64.

Re-requesting reviews for these reasons.

g-linville (approved):

njhale (approved):

Dockerfile
njhale: Thank you!

PR #2305, opened 2023-10-30
Refactor DeploySpec so that secrets are first

thedadams:
Creating/updating secrets first is important because the deployments rely on them. If the deployment gets created/updated first, then secret may not have the information it needs to deploy, causing errors.


njhale (approved):

pkg/controller/appdefinition/deploy.go
```diff
@@ -90,62 +92,44 @@ func DeploySpec(req router.Request, resp router.Response) (err error) {
 		return err
 	}
 
-	if err := addDeployments(req, appInstance, tag, pullSecrets, interpolator, resp); err != nil {
+	if objs, err := ToDeployments(req, appInstance, tag, pullSecrets, interpolator); err != nil {
```
iwilltry42, line 95: Nit: this could be unexported, if wanted

iwilltry42 (approved):
Nice to have this less nested :)

PR #2306, opened 2023-10-30
support http2 protocol (#1523)

keyallis:
acorn-io/manager#1523

### Checklist

### DIFF
--- a/pkg/controller/service/service.go
+++ b/pkg/controller/service/service.go
@@ -1,8 +1,7 @@
 package service
 
 import (
-	"fmt"
-
+	"github.com/acorn-io/baaah/pkg/name"
 	"github.com/acorn-io/baaah/pkg/router"
 	v1 "github.com/acorn-io/runtime/pkg/apis/internal.acorn.io/v1"
 	"github.com/acorn-io/runtime/pkg/ports"
@@ -22,7 +21,8 @@ func RenderServices(req router.Request, resp router.Response) error {
 	if len(http2Ports) > 0 {
 		svcCopy := svcInstance.DeepCopy()
 		svcCopy.Spec.Ports = http2Ports
-		svcCopy.Name = fmt.Sprintf("%s-%s", svcInstance.Name, v1.ProtocolHTTP2)
+		// append uuid to reduce the chance of clash
+		svcCopy.Name = name.SafeConcatName(svcInstance.Name, string(v1.ProtocolHTTP2), string(svcInstance.UID))
 		if svcCopy.Spec.Annotations == nil {
 			svcCopy.Spec.Annotations = map[string]string{}
 		}
--- a/pkg/controller/service/service.go
+++ b/pkg/controller/service/service.go
@@ -18,30 +18,21 @@ func RenderServices(req router.Request, resp router.Response) error {
 
 	var svcList []*v1.ServiceInstance
 	http2Ports := ports.ByProtocol(svcInstance.Spec.Ports, true, v1.ProtocolHTTP2)
-	httpPorts := ports.ByProtocol(svcInstance.Spec.Ports, true, v1.ProtocolHTTP)
-	// If we have both types of ports we need to create a separate service with the annotation
-	if len(http2Ports) > 0 && len(httpPorts) > 0 {
+	otherPorts := ports.ByProtocol(svcInstance.Spec.Ports, false, v1.ProtocolHTTP2)
+	if len(http2Ports) > 0 {
 		svcCopy := svcInstance.DeepCopy()
-		svcCopy.Spec.Ports = ports.ByProtocol(svcInstance.Spec.Ports, false, v1.ProtocolHTTP2) // filter out http2 ports
+		svcCopy.Spec.Ports = http2Ports
+		svcCopy.Name = fmt.Sprintf("%s-%s", svcInstance.Name, v1.ProtocolHTTP2)
+		if svcCopy.Spec.Annotations == nil {
+			svcCopy.Spec.Annotations = map[string]string{}
+		}
+		svcCopy.Spec.Annotations["traefik.ingress.kubernetes.io/service.serversscheme"] = "h2c"
 		svcList = append(svcList, svcCopy)
-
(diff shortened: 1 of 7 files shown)


pkg/controller/service/service.go
```diff
@@ -13,25 +16,58 @@ func RenderServices(req router.Request, resp router.Response) error {
...
-	if err != nil {
-		return err
-	}
-	resp.Objects(objs...)
-	svcInstance.Status.HasService = len(objs) > 0
+	var svcList []*v1.ServiceInstance
+	http2Ports := ports.ByProtocol(svcInstance.Spec.Ports, true, v1.ProtocolHTTP2)
+	httpPorts := ports.ByProtocol(svcInstance.Spec.Ports, true, v1.ProtocolHTTP)
+	// If we have both types of ports we need to create a separate service with the annotation
+	if len(http2Ports) > 0 && len(httpPorts) > 0 {
```
iwilltry42, line 23: I'll put some explanation on this, since we paired up to work on this yesterday and I still feel like this is a good and simple solution that can be a little confusing.

The background is that we have to put a traefik annotation on the K8s Service for http2 support (h2c). With that service we'd break the traffic for "normal" http traefik to other ports listed in that service though.
E.g. if someone uses http2 for the app server, but metrics are served via http.
That means, if we have both http/http2 mixed in a single ServiceInstance, we have to split it up to generate two K8s services.

However, all the code further down the stream is specific to a single service and even re-uses the ServiceInstance name in a lot of places.
So we need a "virtual" ServiceInstance with a different name, holding only the http2 ports to be able to leave lots of code untouched.

The DeepCopies are necessary to leave the original serviceInstance object untouched during that process and we only have to populated actively modified status fields on it.

pkg/controller/service/service.go
```diff
@@ -13,25 +16,49 @@ func RenderServices(req router.Request, resp router.Response) error {
 	// reset, should get repopulated
 	svcInstance.Status.Endpoints = nil
 
-	objs, _, err := services.ToK8sService(req, svcInstance)
-	if err != nil {
-		return err
+	var svcList []*v1.ServiceInstance
+	http2Ports := ports.ByProtocol(svcInstance.Spec.Ports, true, v1.ProtocolHTTP2)
```
StrongMonkey, line 20: Nit: Would you be able to use this function like 
```
http2Ports := ports.ByProtocol(svcInstance.Spec.Ports, v1.ProtocolHTTP2)
otherProtocol := []v1.Protocol{v1.ProtocolHTTP, v1.ProtocolTCP, v1.ProtocolUDP}
otherPorts := ports.ByProtocol(svcInstance.Spec.Ports, otherProtocols)
```
Making the function a bit easy to read

pkg/controller/service/service.go
```diff
@@ -13,25 +16,49 @@ func RenderServices(req router.Request, resp router.Response) error {
...
+	otherPorts := ports.ByProtocol(svcInstance.Spec.Ports, false, v1.ProtocolHTTP2)
+	if len(http2Ports) > 0 {
+		svcCopy := svcInstance.DeepCopy()
+		svcCopy.Spec.Ports = http2Ports
+		svcCopy.Name = fmt.Sprintf("%s-%s", svcInstance.Name, v1.ProtocolHTTP2)
+		if svcCopy.Spec.Annotations == nil {
+			svcCopy.Spec.Annotations = map[string]string{}
+		}
+		svcCopy.Spec.Annotations["traefik.ingress.kubernetes.io/service.serversscheme"] = "h2c"
+		svcList = append(svcList, svcCopy)
```
StrongMonkey, line 30: Do you have to actually create svcInstance or can we just create separate services with special annotations? Seems like we should not be creating svcinstances here.

StrongMonkey: So I wonder if you should be able to create separate corev1.service in `services.ToK8sService(req, svc)`. So for http2 port it will be a separate service with special annotation

pkg/controller/service/service.go
keyallis: I think my big issue with doing that is if we ever add a new protocol then this logic needs to get updated, whereas I'd rather have it naturally handle a situation like that

StrongMonkey: yeah that's true

pkg/controller/service/service.go
StrongMonkey: Discussed offline, this is doing the right thing to create separate service by tweaking serviceInstance that is passed to `ToK8sService`

StrongMonkey (approved):
LGTM, would you be able to add some intergation tests around it

pkg/controller/service/service.go
```diff
@@ -13,25 +16,48 @@ func RenderServices(req router.Request, resp router.Response) error {
...
-	objs, _, err := services.ToK8sService(req, svcInstance)
-	if err != nil {
-		return err
+	var svcList []*v1.ServiceInstance
+	http2Ports := ports.ByProtocol(svcInstance.Spec.Ports, true, v1.ProtocolHTTP2)
+	otherPorts := ports.ByProtocol(svcInstance.Spec.Ports, false, v1.ProtocolHTTP2)
+	if len(http2Ports) > 0 {
+		svcCopy := svcInstance.DeepCopy()
+		svcCopy.Spec.Ports = http2Ports
+		svcCopy.Name = fmt.Sprintf("%s-%s", svcInstance.Name, v1.ProtocolHTTP2)
```
thedadams, line 25: I am agreeing with @StrongMonkey here. Creating an additional `ServiceInstance` with a different name may cause issues with other parts of the system. For example, if something depends on the service, but the service has a different name. Or, the user may create a service that results in a `ServiceInstance` being created with the same name you are creating here.

Would it be possible to render the `corev1.Service` with the special annotations instead of creating an additional `ServiceInstance`?

keyallis: The `ServiceInstance` is only created as a temporary copy for our purposes so that we can pass the same value into the function logic. We don't actually return the temporarily created `ServiceInstance` object, we just use it to create a new `corev1.Service`.

njhale (approved):

PR #2310, opened 2023-11-02
Adjust status calculations

thedadams:
There was an issue where app status was being used to process deployments, jobs, etc but the app status wasn't being persisted. This change should give us the status updates we are looking forward without the "act before persist" issues.


thedadams:
I had to move the `isBlocked` messaging back to where it was previously.

iwilltry42 (approved):

PR #2321, opened 2023-11-04
Deduplicate endpoints for ingress that has port defined

StrongMonkey:
### Checklist

In case of NLB port allocation, the service.status.ingress already have ports allocated. So we shouldn't need to read ports from service.spec.ports, as it will cause duplicated endpoints. So, deduplicate the endpoints to make sure each endpoint is unique 

https://github.com/acorn-io/aws-load-balancer-controller/pull/2

Related issue: https://github.com/acorn-io/manager/issues/1568 https://github.com/acorn-io/manager/issues/1567

### DIFF
--- a/pkg/controller/appstatus/endpoints.go
+++ b/pkg/controller/appstatus/endpoints.go
@@ -30,6 +30,7 @@ func serviceEndpoints(ctx context.Context, c kclient.Client, app *v1.AppInstance
 		return nil, err
 	}
 
+	endpointSet := map[string]v1.Endpoint{}
 	for _, service := range serviceList.Items {
 		containerName := service.Labels[labels.AcornContainerName]
 		if containerName == "" {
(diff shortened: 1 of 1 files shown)


pkg/controller/appstatus/endpoints.go
```diff
@@ -90,7 +94,17 @@ func serviceEndpoints(ctx context.Context, c kclient.Client, app *v1.AppInstance
 		}
 	}
 
-	return
+	set := map[string]v1.Endpoint{}
+	for _, ep := range endpoints {
+		set[fmt.Sprintf("%v://%v", ep.Protocol, ep.Address)] = ep
+	}
+
+	var result []v1.Endpoint
```
thedadams, line 102: nit to avoid multiple allocations:
```suggestion
	result := make([]v1.Endpoint, 0, len(set))
```

pkg/controller/appstatus/endpoints.go
```diff
@@ -90,7 +94,17 @@ func serviceEndpoints(ctx context.Context, c kclient.Client, app *v1.AppInstance
 		}
 	}
 
-	return
+	set := map[string]v1.Endpoint{}
+	for _, ep := range endpoints {
+		set[fmt.Sprintf("%v://%v", ep.Protocol, ep.Address)] = ep
+	}
```
thedadams, line 100: Instead of constructing a slice, then constructing a map, and then constructing a slice again, can the main for-loop above construct the map? That way we dedup as we go?

StrongMonkey: Great call. I can do that.

thedadams (approved):

iwilltry42 (approved):

PR #2351, opened 2023-11-21
add: volume preloading via `?preload=true` directive in Acornfile (manager#1598)

iwilltry42:
Ref https://github.com/acorn-io/manager/issues/1598

Defining a volume, e.g. `dirs: "/usr/share/nginx/html": "volume://myvol?preload=true"` will pre-populate the volume with the data that the used container image has at the defined place.
Usually the directory would be empty (overlayed with the empty volume), so we're mimicking the Docker copy-up behavior here using initContainers.
Note: The data lands in the `data/` subdirectory of the volume.

**Note 2:** There's no validation if ?preload is put on e.g. `secret://` - didn't see that anywhere else, so we just do nothing in that case

## What is this?

First off: Darren wants it. It's a pre-requisite for an upcoming dev flow.
The example above  `dirs: "/usr/share/nginx/html": "volume://myvol?preload=true"` will do the following:

- Add an initContainer to the resulting Deployment which copies the static busybox binary from the Acorn runtime base image to a shared ephemeral volume
- Add another initContainer that's running the target app's image and uses the shared busybox binary to copy the contents of `/usr/share/nginx/html` within the image to the volume myvol's subPath `data/`
    - this is also placing a file `.preload-done` into `myvol`'s root, that we use to avoid doing the process twice (i.e. preloading happens only once in the app lifecycle)
- The `data/` subPath is then mounted into the actual application container on `/usr/share/nginx/html`, so that the image contents are there

- This is mimicking the [default Docker behavior](https://docs.docker.com/storage/volumes/#populate-a-volume-using-a-container), which is not present in Kubernetes. Without this, the mountPath will just be empty if the volume is fresh.

### Checklist
- [x] Commits follow [contributing guidance](https://github.com/acorn-io/runtime/blob/main/CONTRIBUTING.md#commits) -> will be squashed

### DIFF
--- a/pkg/apis/internal.acorn.io/v1/unmarshal.go
+++ b/pkg/apis/internal.acorn.io/v1/unmarshal.go
@@ -1429,9 +1429,13 @@ func parseVolumeReference(s string) (string, string, bool, error) {
 		subPath = u.Query().Get("sub-path")
 	}
 
-	preload, err := strconv.ParseBool(u.Query().Get("preload"))
-	if err != nil {
-		return "", "", false, fmt.Errorf("malformed ?preload value: %v", err)
+	preload := false
+	preloadStr := u.Query().Get("preload")
+	if preloadStr != "" {
+		preload, err = strconv.ParseBool(preloadStr)
+		if err != nil {
+			return "", "", false, fmt.Errorf("malformed ?preload value: %v", err)
+		}
 	}
 
 	return s, subPath, preload, nil
--- a/pkg/apis/internal.acorn.io/v1/unmarshal.go
+++ b/pkg/apis/internal.acorn.io/v1/unmarshal.go
@@ -1434,7 +1434,7 @@ func parseVolumeReference(s string) (string, string, bool, error) {
 	if preloadStr != "" {
 		preload, err = strconv.ParseBool(preloadStr)
 		if err != nil {
-			return "", "", false, fmt.Errorf("malformed ?preload value: %v", err)
+			return "", "", false, fmt.Errorf("malformed ?preload value %q: %v", preloadStr, err)
 		}
 	}
 
--- a/pkg/apis/internal.acorn.io/v1/unmarshal.go
+++ b/pkg/apis/internal.acorn.io/v1/unmarshal.go
@@ -974,7 +974,7 @@ func (in *VolumeMount) UnmarshalJSON(data []byte) error {
 	} else if strings.HasPrefix(s, "./") {
 		in.ContextDir = s
 	} else {
-		in.Volume, in.SubPath, err = parseVolumeReference(s)
+		in.Volume, in.SubPath, in.Preload, err = parseVolumeReference(s)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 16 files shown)


pkg/controller/appdefinition/deploy.go
```diff
@@ -234,7 +234,50 @@ func toContainers(app *v1.AppInstance, tag name.Reference, name string, containe
 		})
 	}
 
+	if addBusybox {
+		// Drop the static busybox binary into a shared volume so that we can use it in initContainers.
+		initContainers = append(initContainers, corev1.Container{
+			Name:            "acorn-helper-busybox",
+			Image:           system.DefaultImage(),
+			Command:         []string{"acorn-busybox-init"},
+			ImagePullPolicy: corev1.PullIfNotPresent,
```
thedadams, line 243: nit: I think this is the default, so it can be dropped.

pkg/controller/appdefinition/deploy.go
```diff
@@ -234,7 +234,50 @@ func toContainers(app *v1.AppInstance, tag name.Reference, name string, containe
...
+	}
+
 	newContainer := toContainer(app, tag, name, container, interpolator, addWait && len(container.Ports) > 0)
+	for src, dir := range container.Dirs {
+		if dir.Preload {
+			// If a directory is marked for preload, then add an initContainer that copies the directory into the shared volume.
+			// Data will be copied to the data/ subdirectory and we'll drop a .preload-done file in the root to indicate that
+			// the copy has been completed (and should not be repeated).
+			initContainers = append(initContainers, corev1.Container{
+				Name:            "acorn-preload-dir-" + sanitizeVolumeName(src),
```
thedadams, line 261: It is possible, though unlikely, for a user to define an init container with this name. Should we add some randomness to it?

thedadams (approved):

pkg/controller/appdefinition/deploy.go
iwilltry42: Fair. Dropped.

pkg/controller/appdefinition/deploy.go
iwilltry42: I'll just add the app UID to it :)
It's already unlikely enough, given that we're using the path hash, so this should do it.

pkg/apis/internal.acorn.io/v1/unmarshal.go
```diff
@@ -1411,14 +1411,14 @@ func parseVolumeDefinition(anonName, s string) (VolumeBinding, error) {
 	return result, nil
 }
 
-func parseVolumeReference(s string) (string, string, error) {
+func parseVolumeReference(s string) (string, string, bool, error) {
```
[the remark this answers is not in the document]
iwilltry42: I added some explanation :)

pkg/appdefinition/acornfile-schema.acorn
```diff
@@ -225,8 +225,8 @@ let types: {
 	User: (int > 0) || (string =~ "^[0-9]+(:[0-9]+)?$")
 
 	ShortVolumeRef: string =~ "^[a-z][-a-z0-9]*$"
-	VolumeRef:      string =~ "^volume://.+$"
-	EphemeralRef:   string =~ "^ephemeral://.*$|^$"
+	VolumeRef:      string =~ "^volume://.+(.preload=(true|false))?$"
+	EphemeralRef:   string =~ "^ephemeral://.*(.preload=(true|false))?$|^$"
```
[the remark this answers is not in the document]
iwilltry42: > I guess I just don't get what this accomplishes.

Nothing. It's a leftover from when I was first checking out the query params here. Removed.

pkg/controller/appdefinition/deploy_test.go
```diff
@@ -354,3 +356,48 @@ func TestUserContext(t *testing.T) {
 	require.Equal(t, int64(3000), *dep.Spec.Template.Spec.Containers[1].SecurityContext.RunAsUser)
 	require.Equal(t, int64(4000), *dep.Spec.Template.Spec.Containers[1].SecurityContext.RunAsGroup)
 }
+
+func TestVolumePreload(t *testing.T) {
```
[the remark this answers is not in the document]
iwilltry42: Good point. Moved it to the superior appdefinition autogold test suite.

tylerslaton (approved):

iwilltry42:
@g-linville I addressed your comments, please have another look whenever you have some time :)

g-linville (approved):

PR #2360, opened 2023-11-22
Add feature flag to control whether we should add don't evict annotation

StrongMonkey:
### Checklist

We should add runtime configuration to control whether we should add karpenter do-not-evict annotation for any deploy that has 1 replica. Ideally, we should only add the annotation when cluster is in production use. For dev purpose, it is better to not have the annotation so that these apps/nodes can be consolidated.


g-linville (approved):

tylerslaton (approved):

pkg/profiles/default.go
```diff
@@ -73,5 +73,6 @@ func defaultProfile() apiv1.Config {
 		ControllerCPU:                  new(string),
 		APIServerMemory:                new(string),
 		APIServerCPU:                   new(string),
+		AutoConfigureKarpenterDontEvictAnnotations: z.Pointer(true),
```
thedadams, line 76: nit: `go fmt`

thedadams (approved):

pkg/profiles/default.go
StrongMonkey: Nope, still give me the same ouput 🤷‍♂️

PR #2361, opened 2023-11-23
change: allow different protocols on the same port (manager#1566)

iwilltry42:
Ref https://github.com/acorn-io/manager/issues/1566

With this, it's OK to expose the same port/targetPort combination as long as they're using different protocols (which are non-blocking).

### Checklist

### DIFF
--- a/pkg/ports/ports_test.go
+++ b/pkg/ports/ports_test.go
@@ -132,6 +132,30 @@ func TestCollectPorts(t *testing.T) {
 				{TargetPort: 7070, Port: 7000, Hostname: "myapp3.local"},
 			},
 		},
+		// same port mappings on different hostnames
+		{
+			name: "same target ports, same ports, different hostnames",
+			ports: []v1.PortDef{
+				{TargetPort: 8080, Port: 8080, Hostname: "myapp.local"},
+				{TargetPort: 8080, Port: 8080, Hostname: "myapp2.local"},
+			},
+			expected: []v1.PortDef{
+				{TargetPort: 8080, Port: 8080, Hostname: "myapp.local"},
+				{TargetPort: 8080, Port: 8080, Hostname: "myapp2.local"},
+			},
+		},
+		{
+			name: "same target ports, same ports, different protocols, different hostnames - tcp twice ",
+			ports: []v1.PortDef{
+				{TargetPort: 8080, Port: 8080, Protocol: v1.ProtocolTCP, Hostname: "myapp.local"},
+				{TargetPort: 8080, Port: 8080, Protocol: v1.ProtocolUDP, Hostname: "myapp2.local"},
+				{TargetPort: 8080, Port: 8080, Protocol: v1.ProtocolHTTP2, Hostname: "myapp3.local"},
+			},
+			expected: []v1.PortDef{
+				{TargetPort: 8080, Port: 8080, Protocol: v1.ProtocolTCP, Hostname: "myapp.local"},
+				{TargetPort: 8080, Port: 8080, Protocol: v1.ProtocolUDP, Hostname: "myapp2.local"},
+			},
+		},
 		{
 			name: "same target ports, same ports, different protocols - tcp twice",
 			ports: []v1.PortDef{
--- a/pkg/ports/ports_test.go
+++ b/pkg/ports/ports_test.go
@@ -137,24 +137,24 @@ func TestCollectPorts(t *testing.T) {
 			ports: []v1.PortDef{
 				{TargetPort: 8080, Port: 8080, Protocol: v1.ProtocolTCP},
 				{TargetPort: 8080, Port: 8080, Protocol: v1.ProtocolUDP},
-				{TargetPort: 8080, Port: 8080, Protocol: v1.ProtocolHTTP},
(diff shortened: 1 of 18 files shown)


pkg/ports/publish.go
```diff
@@ -216,23 +216,43 @@ func collectPorts(seen map[int32][]int32, seenHostnames map[string]struct{}, por
...
-					if t == port.TargetPort {
-						// Same port and target port but different hostnames, so keep both
-						seen[port.Port] = append(targets, port.TargetPort)
-						seenHostnames[port.Hostname] = struct{}{}
-						result = append(result, port)
+		protocolMatch := func(a, b v1.Protocol) bool {
+			return a == b ||
+				(a == v1.ProtocolHTTP && b == v1.ProtocolTCP) ||
+				(a == v1.ProtocolTCP && b == v1.ProtocolHTTP)
+		}
```
iwilltry42, line 223: This is something I don't like and I think we should discuss.
I feel like we should be smarter here.
I.e. HTTP should supersede TCP, as both create a TCP service, right? HTTP just adds an Ingress (if publish is true).
However, both should be able to co-exist if both a `publish: true`.
Any opinions?

vincent99: What about http2?  Or the next protocol that gets added?

Asking for multiple variants of tcp doesn't really make sense; e.g. tcp means the data within is opaque and untouched, http is changing it to add headers.  

The user should pick which one they want and each acorn port definition corresponds directly to exactly one layer-4 port in the container. 

Since we have various kinds of TCP, one UDP and no others, that check could currently be just "a is UDP and b is not".

iwilltry42: I agree that it doesn't make sense, but even http/tcp could be fine. Both result in a k8s service with tcp defined as protocol.
But: http *may* add Ingress. TCP *may* set type LoadBalancer. Http2 will create a separate k8s service with h2c annotation set.

I'm fine with being restrictive/opinionated here though just as you suggest 👍

g-linville: Yeah I'm not sure what I really think here either. In the issue that this PR links to, Bill has an example Acornfile provided by a user and they are defining TCP and UDP on the same port number. Is that something we want to support, and is it even possible?

iwilltry42: I guess at least for plain old tcp/udp via the same port we should have support, since that's used by DNS, RTP services, even OpenVPN or Minecraft Server.
That obviously depends on the configuration, but I think we should support it in general.
Not sure about all the other specific protocols though 🤷‍♂️

iwilltry42: I feel like multiple combinations are possible but potentially unlikely to be used at all.
For now I'll roll with Vince's suggestion of only allowing UDP + 1 flavor of TCP (tcp/http/http2) in general.

iwilltry42:
I feel like this should throw an error/warning way earlier in the process instead of just dropping some ports without further notice. Any opinions on this?

pkg/ports/publish.go
```diff
@@ -216,23 +216,39 @@ func collectPorts(seen map[int32][]int32, seenHostnames map[string]struct{}, por
...
+		if seenPortDefs, ok := seen[port.Port]; ok {
+			discard := false
+			for _, p := range seenPortDefs {
+				if port.TargetPort == p.TargetPort {
+					if port.Hostname != "" {
+						// OK: Same port and target port (and potentially protocol) but different hostnames, so keep both
 						break
 					}
+					if !(port.Protocol == v1.ProtocolUDP && p.Protocol != v1.ProtocolUDP ||
+						port.Protocol != v1.ProtocolUDP && p.Protocol == v1.ProtocolUDP) {
```
keyallis, line 228: This could be logically cleaned up, I'd also recommend putting the `&` blocks in parens so it's easier to understand how it gets interpreted. It should be:
```
if (port.Protocol != v1.ProtocolUDP && p.Protocol != v1.ProtocolUDP) || (port.Protocol == v1.ProtocolUDP && p.Protocol == v1.ProtocolUDP) {
```

keyallis: To explain this further, you can view `port = udp` as a logical boolean value `A` (thus `port != udp` is `!A`), and `p = udp` as `B` (thus `p != udp` is `!B`). Your initial logical statement is `!(A!B|!AB)`, which simplifies to `!A!B|AB`

pkg/ports/ports_test.go
```diff
@@ -132,11 +132,45 @@ func TestCollectPorts(t *testing.T) {
 				{TargetPort: 7070, Port: 7000, Hostname: "myapp3.local"},
 			},
 		},
+		{
```
keyallis, line 135: Would it be possible to add a test around the same port mappings happening on different hostnames?

keyallis (approved):
Generally looks good, requested a logic change and potentially another test case

tylerslaton (approved):
Approving with the assumption that @keyallis's feedback is addressed. That's the only thing blocking in my mind here.

pkg/ports/ports_test.go
iwilltry42: Done :+1:

iwilltry42:
Fixed the logic to match what we discussed before - added related comments.
I also had to update quite a bunch of other files to make the `revive` linter happy :grimacing:

PR #2362, opened 2023-11-23
fix: replace / with - in anonymous ephemeral volume names (manager-specific) (manager#1598)

iwilltry42:
`dirs: "/foo/bar": "ephemeral://"` stores a volume with name `/foo/bar`, which is OK in runtime, but is actually used as a resource name in Manager, which cannot contain `/`, so we're replacing this here (instead of adapting the store strategy in Manager)

I'm not sure if we'll need more sophisticated replaces here to make it a DNS name :thinking: 

Discovered as part of [ #1598](https://github.com/acorn-io/manager/issues/1598)

### Checklist


g-linville (approved):
This looks fine to me, but how badly will this disrupt existing volumes with `/` in the name from the user's perspective?

iwilltry42:
@g-linville this only affects `ephemeral://` (no name given) auto-generated volume names, so the impact will probably be fairly low, especially since they apparently didn't work in Manager so far.

thedadams (approved):

PR #2363, opened 2023-11-26
Add support for runtime classes on Acorn computeclasses 

cloudnautique:
Runtime classes are added as in internal value on Acorn compute class objects. This allows for using things like Kata Containers.

### DIFF
--- a/pkg/controller/appdefinition/deploy.go
+++ b/pkg/controller/appdefinition/deploy.go
@@ -778,7 +778,7 @@ func toDeployment(req router.Request, appInstance *v1.AppInstance, tag name.Refe
 					Affinity:                      appInstance.Status.Scheduling[name].Affinity,
 					Tolerations:                   appInstance.Status.Scheduling[name].Tolerations,
 					PriorityClassName:             appInstance.Status.Scheduling[name].PriorityClassName,
-					RuntimeClassName:              z.Pointer[string](appInstance.Status.Scheduling[name].RuntimeClassName),
+					RuntimeClassName:              stringOrNilPtr(appInstance.Status.Scheduling[name].RuntimeClassName),
 					TerminationGracePeriodSeconds: z.Pointer[int64](10),
 					ImagePullSecrets:              pullSecrets.ForContainer(name, append(containers, initContainers...)),
 					EnableServiceLinks:            new(bool),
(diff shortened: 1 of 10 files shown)


pkg/controller/appdefinition/deploy.go
```diff
@@ -867,3 +868,10 @@ func ToDeployments(req router.Request, appInstance *v1.AppInstance, tag name.Ref
 
 	return result, nil
 }
+
+func stringOrNilPtr(str string) *string {
+	if str != "" {
+		return z.Pointer[string](str)
```
tylerslaton, line 874: nit: necessary?
```suggestion
		return z.Pointer(str)
```

pkg/install/role.yaml
```diff
@@ -131,4 +135,4 @@ roleRef:
 subjects:
   - kind: ServiceAccount
     namespace: acorn-image-system
-    name: acorn-image-system
```
tylerslaton, line 134: nit: Was this intentional?

pkg/controller/appdefinition/deploy.go
cloudnautique: If I understand the signature: https://github.com/acorn-io/z/blob/main/z.go#L21 and the usage on line 782:                                         
```
...
TerminationGracePeriodSeconds: z.Pointer[int64](10),
...
```

I believe it is necessary.

pkg/install/role.yaml
cloudnautique: I believe the YAML file was auto formatted. I don't think it hurts anything.

pkg/controller/scheduling/scheduling.go
```diff
@@ -125,6 +132,21 @@ func PriorityClassName(req router.Request, computeClass *adminv1.ProjectComputeC
 	return computeClass.PriorityClassName, nil
 }
 
+// RuntimeClassName checks that a defined RuntimeClass exists and returns the name of it
+func RuntimeClassName(req router.Request, computeClass *adminv1.ProjectComputeClassInstance) (string, error) {
```
cjellick, line 136: You are exporting this function. I don't think it is used outside this package. If that's true, can you unexport (lowercase) it?

cjellick: Seems like you were copying the pattern of PriorityClassName. If the same is true for that one, please unexport it as well

cjellick:
@cloudnautique can you give a practicall example here? I understand this is related to the kata POC. But can you:
1. show me in yaml what the compute class will look like
2. explain what that translates to at a k8s infra level. Like, where does that runtime class get set?

cjellick (changes requested):

thedadams (approved):

cjellick (approved):

PR #2366, opened 2023-11-30
Ignore auth failures when trying to pull apps during cli login

keyallis:
This has been causing failures in e2e

### Checklist


g-linville:
I'm a bit confused; why are we getting auth failures in the first place?

keyallis:
If we are logged in as a user that doesn't have permissions to pull apps in a given project and then we try to log in as another user. What we're seeing in e2e is that when we log in as a non-permissioned user for rbac tests and then try to log back in as the default e2e user we fail to pull app information and then never actually log in because the cli never attempts to do the auth login flow once we fail to pull apps

keyallis:
That would be a question for @ibuildthecloud

g-linville (approved):

thedadams (approved):

PR #2368, opened 2023-12-05
Ensure that nginx does DNS lookups

thedadams:
When using static hostnames, nginx will lookup the DNS for these hostnames on startup/config reload and won't do it again until the container is restarted or the config is reloaded. This means that if the service IPs change, then nginx won't have the correct IPs until one of these things happen.

This change uses a hack to ensure that nginx will look up for the DNS for services at least every 15 seconds. This ensures that if the service IPs change for some reason, then nginx will get them.

### DIFF
--- a/pkg/controller/appdefinition/router.go
+++ b/pkg/controller/appdefinition/router.go
@@ -1,6 +1,7 @@
 package appdefinition
 
 import (
+	"context"
 	"crypto/sha256"
 	"encoding/hex"
 	"strconv"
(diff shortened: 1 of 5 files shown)


pkg/controller/appdefinition/router.go
```diff
@@ -165,9 +172,11 @@ func toRouter(appInstance *v1.AppInstance, routerName string, router v1.Router)
 	}, nil
 }
 
-func toNginxConf(routerName string, router v1.Router) (string, string) {
+func toNginxConf(internalClusterDomain, namespace, routerName string, router v1.Router) (string, string) {
 	buf := &strings.Builder{}
-	buf.WriteString("server {\nlisten 8080;\n")
+	buf.WriteString("resolver kube-dns.kube-system.")
```
thedadams, line 177: @ibuildthecloud This is the only part of this that I am hesitant about.

g-linville (approved):

iwilltry42 (approved):

StrongMonkey:
In which case, service IP could change and that would require nginx to look up again?

thedadams:
@StrongMonkey A change like [this one](https://github.com/acorn-io/manager/pull/1705) where the linked services are changing.

StrongMonkey (approved):

PR #2369, opened 2023-12-05
Resolved offerings

g-linville:
### Overview

This change refactors `AppInstance.Status.Defaults` to `AppInstance.Status.ResolvedOfferings`. This field used to store the default values for container memory requirements, the region, and volumeclass information for the AppInstance. With these changes, the ResolvedOfferings stores the actual value of the app's region, computeclass information for each container in the app (rather than just memory information), and volumeclass information for each volume. Now it is the source of absolute truth, and there is no need to compare `AppInstance.Spec` (overrides from the user), `AppInstance.Status.AppSpec` (defaults from the acorn image), and `AppInstance.Status.Defaults` (defaults from volume and compute classes) like we used to.

This is a **breaking API change**. I have updated all of the unit tests in here to reflect that change. (Sorry, there are a lot of them.) I believe this will not cause deployed apps to restart, but they will not get the status updates until they are processed again by the controller (which requires their generation to increase). If we want them to all restart and get the status updates as soon as this runtime change rolls out, I can do that pretty easily as well.

### Checklist


pkg/controller/appdefinition/testdata/volumes/cluster-default-with-bind-combos/expected.golden
```diff
@@ -140,10 +140,10 @@ metadata:
   namespace: app-created-namespace
 spec:
   accessModes:
-  - ReadWriteMany
+  - ReadWriteOnce
   resources:
     requests:
-      storage: 6Gi
+      storage: 3Gi
```
g-linville, line 146: note to self: this looks incorrect; need to investigate

pkg/volume/volume.go
```diff
@@ -151,24 +151,30 @@ func getVolumeClassNames(volumeClasses map[string]adminv1.ProjectVolumeClassInst
 	return typed.SortedKeys(storageClassName)
 }
 
-func CopyVolumeDefaults(volumeRequest v1.VolumeRequest, volumeBinding v1.VolumeBinding, volumeDefaults v1.VolumeDefault) v1.VolumeRequest {
+func ResolveVolumeRequest(volumeRequest v1.VolumeRequest, volumeBinding v1.VolumeBinding, resolvedVolume v1.VolumeResolvedOffering) v1.VolumeRequest {
+	// Order of precedence:
+	// 1. VolumeResolvedOffering - the previously resolved values for Class, Size, and AccessModes
+	// 2. VolumeBinding - the values set by the user in the app spec
+	// 3. VolumeRequest - the defaults as defined in the acorn image
```
thedadams, line 158: Volume bindings are the ones set by the user at runtime (with the `-v` flag). I think these are the top precedence.

pkg/controller/appdefinition/testdata/volumes/cluster-default-with-bind-combos/input.yaml
```diff
@@ -34,12 +34,13 @@ status:
...
-  defaults:
+  resolvedOfferings:
     volumes:
       foo:
-        class: test-custom-class-default
-        size: 2Gi
-        accessModes: [ "readWriteMany" ]
+        class: test-custom-class
+        size: 3Gi
+        accessModes: [ "readWriteOnce" ]
```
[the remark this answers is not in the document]
g-linville: So that it matches what is defined in `spec.volumes` higher up in the file. By the time we reach this unit test, offerings should already be resolved, so it doesn't make sense for our input file to have them resolved incorrectly.

pkg/controller/resolvedofferings/volumeclass.go
```diff
@@ -0,0 +1,112 @@
...
+		// field has been set in the past, we want to migrate that over to be set on app.Status.Volumes[name].
+		// There is another edge case where the Size field was set by a VolumeClass's default size. In this
+		// case we want to leave the Size field alone.
+		if app.Status.ResolvedOfferings.VolumeSize != nil && resolvedVolume.Size == "" {
+			resolvedVolume.Size = internalv1.Quantity(app.Status.ResolvedOfferings.VolumeSize.String())
+		}
+
+		bound := volumeBindings[name].Volume != ""
+
+		if _, alreadySet := app.Status.ResolvedOfferings.Volumes[name]; !alreadySet && !bound {
```
thedadams, line 55: This makes it seem like it is not possible to change a volume's settings. For example, if a volume as been resolved and is `alreadySet`, then would it be possible to change its size or access modes?

The difference of what this did before was that it was only calculating the defaults, and we decided that the defaults for a volume shouldn't change because it causes too many issues. However, now this is the completely "resolved" volume information, which should be allowed to change if the user tries to change it via an Acornfile update (I think your logic allows the user to change volume information via the command line).

pkg/volume/volume.go
```diff
@@ -151,24 +151,30 @@ func getVolumeClassNames(volumeClasses map[string]adminv1.ProjectVolumeClassInst
 	return typed.SortedKeys(storageClassName)
 }
 
-func CopyVolumeDefaults(volumeRequest v1.VolumeRequest, volumeBinding v1.VolumeBinding, volumeDefaults v1.VolumeDefault) v1.VolumeRequest {
+func ResolveVolumeRequest(volumeRequest v1.VolumeRequest, volumeBinding v1.VolumeBinding, resolvedVolume v1.VolumeResolvedOffering) v1.VolumeRequest {
```
thedadams, line 154: The previous iteration of this function had one responsibility: to put the request, binding and default together to get a merged volume request for the volume to process.

It seems that now this doesn't completely "resolve" the volume. It resolves the request without the defaults and the default information is added later, if needed. In other words, it kind of only does have the job now.

My opinion, which can be completely disregarded, is that passing the default information here and using it to complete "resolve" the volume would be better. Some care would need to be taken so that this function can be uses in the various places it is being used, but I think it would make it more complete.

pkg/controller/resolvedofferings/volumeclass.go
g-linville: Oh, good point. I will work on fixing this.

pkg/apis/internal.acorn.io/v1/appinstance.go
```diff
@@ -52,24 +52,18 @@ type AppInstance struct {
...
 
 func (in *AppInstance) SetDefaultRegion(region string) {
-	if in.Spec.Region == "" {
-		if in.Status.Defaults.Region == "" {
-			in.Status.Defaults.Region = region
-		}
-	} else {
-		in.Status.Defaults.Region = ""
-	}
+	in.Status.ResolvedOfferings.Region = region
```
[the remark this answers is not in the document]
g-linville: @thedadams let me know if this function looks good to you now. I am still a little bit confused as to how it is supposed to work to be honest.

pkg/controller/appdefinition/deploy.go
```diff
@@ -597,12 +597,25 @@ func containerAnnotation(container v1.Container) string {
 	return string(json)
 }
 
+func resolvedOfferingsAnnotation(appInstance *v1.AppInstance, container v1.Container) string {
+	if resolved, exists := appInstance.Status.ResolvedOfferings.Containers[container.Name]; exists {
+		data, _ := convert.EncodeToMap(resolved)
+		j, _ := json.Marshal(data)
```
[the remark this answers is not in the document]
g-linville: Fixed

pkg/controller/scheduling/scheduling.go
```diff
@@ -157,10 +157,10 @@ func ResourceRequirements(req router.Request, app *v1.AppInstance, containerName
...
 
 	var memDefault *int64
-	if val, ok := app.Status.Defaults.Memory[containerName]; ok && val != nil {
-		memDefault = val
-	} else if val, ok := app.Status.Defaults.Memory[""]; ok && val != nil {
-		memDefault = val
+	if val, ok := app.Status.ResolvedOfferings.Containers[containerName]; ok && val.Memory != nil {
+		memDefault = val.Memory
+	} else if val, ok := app.Status.ResolvedOfferings.Containers[""]; ok && val.Memory != nil {
+		memDefault = val.Memory
```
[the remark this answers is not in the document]
g-linville: Fixed

pkg/server/registry/apigroups/acorn/apps/validator.go
```diff
@@ -258,7 +258,7 @@ func (s *Validator) ValidateUpdate(ctx context.Context, obj, old runtime.Object)
 		}
 	}
 
-	if newParams.Spec.Region != oldParams.Spec.Region && newParams.Spec.Region != oldParams.Status.Defaults.Region {
+	if newParams.Spec.Region != oldParams.Spec.Region && newParams.Spec.Region != oldParams.Status.ResolvedOfferings.Region {
```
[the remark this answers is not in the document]
g-linville: Didn't go with the suggestion because it caused update validation issues, but I did modify this line. Let me know what you think.

pkg/apis/api.acorn.io/v1/types.go
```diff
@@ -23,10 +23,7 @@ func (in *App) GetStopped() bool {
 }
 
 func (in *App) GetRegion() string {
-	if in.Spec.Region != "" {
-		return in.Spec.Region
-	}
-	return in.Status.Defaults.Region
+	return in.Status.ResolvedOfferings.Region
```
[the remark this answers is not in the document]
g-linville: Good question. I will test updating to my new runtime changes with existing apps installed and make sure that the offerings get resolved properly.

pkg/controller/appdefinition/volume.go
```diff
@@ -197,16 +196,7 @@ func toPVCs(req router.Request, appInstance *v1.AppInstance) (result []kclient.O
...
-				// volume set, then we use that. Otherwise, we use v1 package level default size.
-				defaultSize := *v1.DefaultSize
-				if defaultVolume, hasDefaultSet := appInstance.Status.Defaults.Volumes[vol]; hasDefaultSet {
-					defaultSize, err = resource.ParseQuantity(string(defaultVolume.Size))
-					if err != nil {
-						return nil, fmt.Errorf("failed to parse default volume size %q for volume %q: %w", defaultVolume.Size, vol, err)
-					}
-				}
-				pvc.Spec.Resources.Requests[corev1.ResourceStorage] = defaultSize
+				pvc.Spec.Resources.Requests[corev1.ResourceStorage] = *v1.DefaultSize
```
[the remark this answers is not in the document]
g-linville: Not sure why I removed that, lol. I'll add it back.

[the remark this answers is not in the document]
g-linville: Oh wait, nvm. The reason I removed that is because the refactored ResolveVolumeRequest function (called earlier here) will handle setting the default size and checking to see if there is already a size that was previously resolved. So by the time we get down here, if there is still no determined size, the only thing left to use is `v1.DefaultSize`.

pkg/apis/internal.acorn.io/v1/appinstance.go
```diff
@@ -52,23 +52,23 @@ type AppInstance struct {
...
 
 func (in *AppInstance) SetDefaultRegion(region string) {
-	if in.Spec.Region == "" {
-		if in.Status.Defaults.Region == "" {
-			in.Status.Defaults.Region = region
+	if in.Status.ResolvedOfferings.Region == "" {
+		if in.Spec.Region != "" && in.Status.ResolvedOfferings.Region == "" {
+			in.Status.ResolvedOfferings.Region = in.Spec.Region
+		} else {
+			in.Status.ResolvedOfferings.Region = region
```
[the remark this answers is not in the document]
g-linville: I reworked it a little bit. Let me know if you think it looks good. I'm still kinda confused about this.

pkg/controller/resolvedofferings/volumeclass.go
g-linville: So, from what I understand, you cannot change the access modes of an existing PVC in Kubernetes, but you can increase its size. I heavily modified this function so that the size can be changed even if the volume offering has been previously resolved, but the volume class and access modes cannot be changed after the volume is created. Let me know if it looks good to you. From my testing, everything seems to me to be working properly.

pkg/volume/volume.go
g-linville: I reworked the function to fully resolve the volume based on the previously resolved values, all compute class information, bindings specified by the user, etc. Thankfully it didn't require too many modifications in the other places where this function is called.

pkg/apis/api.acorn.io/v1/types.go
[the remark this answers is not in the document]
g-linville: Just verified. The resolution occurs as soon as the upgrade happens, so there is no concern about offerings not being calculated yet.

pkg/apis/internal.acorn.io/v1/appinstance.go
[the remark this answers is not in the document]
thedadams: Honestly, I think what I suggested is simpler. However, what you have does the job.

pkg/apis/internal.acorn.io/v1/appinstance.go
```diff
@@ -52,23 +52,23 @@ type AppInstance struct {
 }
 
 func (in *AppInstance) HasRegion(region string) bool {
-	return in.Status.Defaults.Region == region || in.Spec.Region == region
+	return in.Status.ResolvedOfferings.Region == region || in.Spec.Region == region
```
thedadams, line 55:
```suggestion
	return in.Status.ResolvedOfferings.Region == region
```

pkg/apis/internal.acorn.io/v1/appinstance.go
```diff
@@ -52,23 +52,23 @@ type AppInstance struct {
...
 }
 
 func (in *AppInstance) GetRegion() string {
-	if in.Spec.Region != "" {
-		return in.Spec.Region
+	if in.Status.ResolvedOfferings.Region != "" {
+		return in.Status.ResolvedOfferings.Region
 	}
-	return in.Status.Defaults.Region
+	return in.Spec.Region
```
thedadams, line 62: I think we can just `return in.Status.ResolvedOfferings.Region` here.

pkg/controller/appdefinition/deploy.go
```diff
@@ -597,12 +597,34 @@ func containerAnnotation(container v1.Container) string {
...
+	return "", nil
+}
+
 func podAnnotations(appInstance *v1.AppInstance, container v1.Container) map[string]string {
 	annotations := map[string]string{
 		labels.AcornContainerSpec: containerAnnotation(container),
 	}
 	addPrometheusAnnotations(annotations, container)
 
+	offerings, err := resolvedOfferingsAnnotation(appInstance, container)
```
StrongMonkey, line 623: Looks like this error is still not handled...

StrongMonkey (approved):

pkg/controller/resolvedofferings/volumeclass.go
thedadams: It doesn't seem to set a different size if a volume is `alreadySet` and not `bound`.

thedadams (approved):

pkg/controller/appdefinition/deploy.go
g-linville: Below this line there is a check for `if err == nil`, and then we attach the annotation. I don't actually want to return this error if it happens, since we can safely just ignore it and keep going.

Issue #2380, opened 2023-12-15
Add "resources" field to the computeClass definition


dciangot:
Introducing the field resources to the [ProjectComputeClassInstance](https://github.com/acorn-io/runtime/blob/main/pkg/apis/internal.admin.acorn.io/v1/computeclasses.go#L44) and, as a consequence, to [ClusterComputeClassInstance](https://github.com/acorn-io/runtime/blob/main/pkg/apis/internal.admin.acorn.io/v1/computeclasses.go#L13) will enable cluster admins to add GPU and any custom hardware to the "offering" for the users.

Note from Darren: _"One catch is that the resources can only be applied to the "main container" not any sidecars, as you really don't know which one to apply to."_
Complete discussion ref on [slack](https://acorn-users.slack.com/archives/C03R9ME0SKC/p1702567598185389)

sangee2004:
Tested with acorn version - `v0.10.0-rc2-9-g43dbcbf4+43dbcbf4`

1. Able to create a computeclass with `resources` field containing `requests` and `limits` using following yaml :
```
kind: ClusterComputeClass
apiVersion: admin.acorn.io/v1
default: false
metadata:
  name: cc-res
description: Large compute for linux on arm64
cpuScaler: 0.75
supportedRegions:
 - local
memory:
  default: 20M
  min: 10M
  max: 100M
resources:
  limits:
    gpu-vendor.example/example-gpu: 2
  requests:
    gpu-vendor.example/example-gpu: 2
``` 

2. When deploying app using this compute class , I am able to see the pods having the resources set in the container spec as follows:
```
                        "resources": {
                            "limits": {
                                "gpu-vendor.example/example-gpu": "2",
                                "memory": "20M"
                            },
                            "requests": {
                                "cpu": "14m",
                                "gpu-vendor.example/example-gpu": "2",
                                "memory": "20M"
                            }
                        },
```

The app deployment itself fails in my case with following error since I dont have an environment that has nodes that support gpu which is as expected.
  
```
0/1 nodes are available: 1 Insufficient gpu-vendor.example/example-gpu. preemption: 0/1 nodes are available: 1 No preemption victims found for incoming pod
```

PR #2384, opened 2023-12-19
Add resources field to computeclass #2380

dciangot:
The edited version of the computeclass will allow to specify any kind of resources allowed by K8s corev1.ResourceRequirements. 
In principle that should be enough to enable admins to create computeclass with GPU and accelarators in general. 

### NOTES/QUESTIONS

- The quota management of these resources is postponed for later times as per @cjellick suggestion, since it will take some more thoughts.
- To avoid conflicts with the "first class" fields for memory and cpu scaling the client should rise an error if the admin tries to push a complute class where memory/cpu limits/requests are passed into the "resources" field. I have an integration test ready for this, but commented out since I'm failing to find the proper spot in the codebase where to implement such a check.

### Checklist


cjellick:
Context is this slack thread: https://acorn-users.slack.com/archives/C03R9ME0SKC/p1702567598185389

@ibuildthecloud effectively blessed the approach conceptually

integration/client/computeclass/computeclass_test.go
```diff
@@ -144,6 +162,26 @@ func TestCreatingComputeClasses(t *testing.T) {
...
+		// },
+		// {
+		// 	name: "invalid-custom-resources-requests",
+		// 	resources: corev1.ResourceRequirements{
+		// 		Requests: corev1.ResourceList{
+		// 			"memory": resource.MustParse("1"),
+		// 		},
+		// 	},
+		// 	fail: true,
+		// },
```
thedadams, line 184: Can this be removed?

pkg/controller/scheduling/scheduling.go
```diff
@@ -70,6 +70,12 @@ func addScheduling(req router.Request, appInstance *v1.AppInstance, workloads ma
 		}
 
 		for sidecarName, sidecarContainer := range container.Sidecars {
+			// disable extra resource indication (GPU et al. for sidecars)
+			if computeClass != nil {
+				if computeClass.Resources != nil {
+					computeClass.Resources = &corev1.ResourceRequirements{Limits: corev1.ResourceList{}, Requests: corev1.ResourceList{}}
+				}
+			}
```
thedadams, line 78:
```suggestion
			if computeClass != nil && computeClass.Resources != nil {
				computeClass.Resources = &corev1.ResourceRequirements{}
			}
```

pkg/controller/scheduling/scheduling.go
```diff
@@ -155,6 +161,16 @@ func ResourceRequirements(req router.Request, app *v1.AppInstance, containerName
...
+	if computeClass != nil {
+		if computeClass.Resources != nil {
+			if computeClass.Resources.Requests != nil {
+				requirements.Requests = computeClass.Resources.Requests
+			}
+			if computeClass.Resources.Limits != nil {
+				requirements.Limits = computeClass.Resources.Limits
+			}
+		}
+	}
```
thedadams, line 173:
```suggestion
	if computeClass != nil && computeClass.Resources != nil {
		if computeClass.Resources.Requests != nil {
			requirements.Requests = computeClass.Resources.Requests
		}
		if computeClass.Resources.Limits != nil {
			requirements.Limits = computeClass.Resources.Limits
		}
	}
```

thedadams (changes requested):
Thank you for your contribution!

Additional to my comments, I believe you also need to check in the changes generated from running `make generate`. The CI would have caught that had it run.

pkg/controller/appdefinition/testdata/computeclass/generic-resources/existing.yaml
```diff
@@ -0,0 +1,25 @@
...
+cpuScaler: 0.25
+memory:
+  min: 1Mi
+  max: 2Mi
+  default: 1Mi
+resources:
+  limits:
+    gpu-vendor.example/example-limit: 1
+  requests:
+    gpu-vendor.example/example-request: 1
```
tylerslaton, line 16: I don't actually think this test is working properly. If you look at the `expected.golden` that got generated, it doesn't actually have these limits in it.

integration/client/computeclass/computeclass_test.go
dciangot: sure, it was there because I tried to raise errors for the admin when specifying CPU and Memory in the resources field instead that on the Memory and CPU Scale fields. I don't see an easy/useful way to enforce it, since the definition of compute classes does not pass through the acorn client afaiu.

If you have any suggestion I'll give it a try, otherwise I'll go ahead removing this

pkg/controller/appdefinition/testdata/computeclass/generic-resources/existing.yaml
dciangot: right, good catch thanks @tylerslaton 
Do I understand correctly that in this test I have to set in input.yaml the AppInstance expected after the scheduling? Meaning that `status.scheduling.oneimage.requirements` in input.yaml should already contain the gpu information? That makes me wonder if existing.yaml has any effect other than reference.

While the scheduling part is validated a part in the scheduling folder.

dciangot:
> Thank you for your contribution!
> 
> Additional to my comments, I believe you also need to check in the changes generated from running `make generate`. The CI would have caught that had it run.

Thank you @thedadams , indeed I missed that make step, sorry about that. 

I have ready a new commit for the rest of the comments, but I'd wait for answers to a couple of questions/proposal/clarification I put inline before pushing it

pkg/controller/scheduling/scheduling.go
dciangot: fixed. Don't know how you prefer me to click on resolve or you'll do it.

pkg/controller/scheduling/scheduling.go
dciangot: fixed.

dciangot:
the recent commit broke the new tests for computeclass, I have rebased on the current main, hope it is ok for you

integration/client/computeclass/computeclass_test.go
dciangot: removed, and put into the docs the description of the behavior in case of memory/cpu passed in this field

pkg/controller/appdefinition/testdata/computeclass/generic-resources/existing.yaml
tylerslaton: You are correct in your understanding that `input.yaml` is expected to be the AppInstance after the `scheduling` handler has run. You are also correct in that `existing.yaml` has no effect other than reference. The fix here is to add your `resources` section into the `input.yaml` so that we can validate it actually appears in the underlying Deployment, which is what the `computeclass_test.go` cases are testing for.

dciangot: awesome, thank you. In that case the commits I pushed yesterday are in line with what you just said.

thedadams (approved):

tylerslaton (approved):
Thank you for the contribution! Super excited to see this get in.

integration/client/computeclass/computeclass_test.go
thedadams: Hello @dciangot. Back from the holiday break. Now that I am properly rested, I better understand what you are saying here. If you would like to add validation for this scenario such that the above (now removed) integration test passes, you would need to add some validation logic [here](https://github.com/acorn-io/runtime/blob/85cfe8c01b57609ae5213adbed04c9e2611e4fb8/pkg/server/registry/apigroups/admin/computeclass/validator.go#L63) and [here](https://github.com/acorn-io/runtime/blob/85cfe8c01b57609ae5213adbed04c9e2611e4fb8/pkg/server/registry/apigroups/admin/computeclass/validator.go#L26).

Adding validation logic in both places would ensure that an admin wouldn't be able to set the memory and cpu on the resources field at the API level.

Would you like to take the time to add this validation now?

dciangot: Sure no problem. I'll add the validation then.

cjellick:
Looks like a new commit came in. WIll need rereviewed @thedadams @tylerslaton 

Tyler, once this is good, can you own getting it merged and shepherding it along internally (whatever that may mean...actually dont think theres much to do there other than merge)

dciangot:
Yeah, it was on the latest @thedadams request. Now integration is failing, not related to the changes I made though. I can take a look later, meanwhile any hint is welcome.

pkg/server/registry/apigroups/admin/computeclass/validator.go
```diff
@@ -39,6 +41,22 @@ func (s *ProjectValidator) Validate(ctx context.Context, obj runtime.Object) (re
...
+				result,
+				field.Invalid(field.NewPath(fmt.Sprintf("spec.resources.requests%s", string(resource))), cc.Default, fmt.Sprintf("Cannot specifiy spec.resources.requests.%s. Use explicit spec.cpuScaler and spec.memory instead", string(resource))))
+		}
+	}
+
+	for _, resource := range invalidResources {
+		if _, specified := cc.Resources.Limits[resource]; specified {
+			return append(result, field.Invalid(field.NewPath(fmt.Sprintf("spec.resources.limits%s", string(resource))), cc.Default, fmt.Sprintf("Cannot specifiy spec.resources.limits.%s. Use explicit spec.cpuScaler and spec.memory instead", string(resource))))
+		}
+	}
```
tylerslaton, line 58: Any reason to not do this in one iteration?

```suggestion
	for _, resource := range invalidResources {
		if _, specified := cc.Resources.Requests[resource]; specified {
			return append(
				result,
				field.Invalid(field.NewPath(fmt.Sprintf("spec.resources.requests%s", string(resource))), cc.Default, fmt.Sprintf("Cannot specifiy spec.resources.requests.%s. Use explicit spec.cpuScaler and spec.memory instead", string(resource))))

		if _, specified := cc.Resources.Limits[resource]; specified {
			return append(result, field.Invalid(field.NewPath(fmt.Sprintf("spec.resources.limits%s", string(resource))), cc.Default, fmt.Sprintf("Cannot specifiy spec.resources.limits.%s. Use explicit spec.cpuScaler and spec.memory instead", string(resource))))
		}
	}
```

cjellick:
we'll rekick the tests after it officially fails to see if it was a flake

pkg/server/registry/apigroups/admin/computeclass/validator.go
```diff
@@ -76,6 +94,20 @@ func (s *ClusterValidator) Validate(ctx context.Context, obj runtime.Object) (re
...
+		if _, specified := cc.Resources.Requests[resource]; specified {
+			return append(result, field.Invalid(field.NewPath(fmt.Sprintf("spec.resources.requests%s", string(resource))), cc.Default, fmt.Sprintf("Cannot specifiy spec.resources.requests.%s. Use explicit spec.cpuScaler and spec.memory instead", string(resource))))
+		}
+	}
+
+	for _, resource := range invalidResources {
+		if _, specified := cc.Resources.Limits[resource]; specified {
+			return append(result, field.Invalid(field.NewPath(fmt.Sprintf("spec.resources.limits%s", string(resource))), cc.Default, fmt.Sprintf("Cannot specifiy spec.resources.limits.%s. Use explicit spec.cpuScaler and spec.memory instead", string(resource))))
+		}
+	}
```
tylerslaton, line 109: Same question here, as well.

tylerslaton (commented):
@cjellick Yep, I'll own getting this merged.

@dciangot Just one small comment and I'll reapprove once it is addressed. I believe the failure you were seeing was flake and I've restarted CI to check. If it fails we can look further.

pkg/server/registry/apigroups/admin/computeclass/validator.go
thedadams: Yeah, let's do this in one loop.

thedadams (changes requested):

pkg/server/registry/apigroups/admin/computeclass/validator.go
dciangot: absolutely! done

tylerslaton (approved):

thedadams (approved):

tylerslaton:
Merging now, thanks for the contribution @dciangot!

PR #2391, opened 2023-12-21
Add support for acorn.region interpolation in Acornfiles

thedadams:


pkg/secrets/interpolation.go
```diff
@@ -241,6 +241,19 @@ func (i *Interpolator) GetProjectName() (string, error) {
...
+	if region := i.app.GetRegion(); region != "" {
+		return region, nil
+	}
+
+	project := &v1.ProjectInstance{}
+	if err := i.client.Get(i.ctx, router.Key("", i.app.Namespace), project); err != nil {
+		return "", err
+	}
+
+	return project.Status.DefaultRegion, nil
```
njhale, line 254: Is there an assumption that `project.Status.DefaultRegion != ""` here?

I believe there are scenarios in which it's possible (albeit unlikely) that the `DefaultRegion` field hasn't been set by the time this is executed. IIUC, when that happens, `@{acorn.region}` is initially resolved to `""`.

njhale (approved):
I left a comment, but it's not a blocker.

LGTM

iwilltry42 (approved):

pkg/secrets/interpolation.go
thedadams: Yes, but this code runs in the controller. So, as soon as the project changes, this run again.

Additionally, in the SaaS where regions are more important, this is less likely to happen because syncing only happens after these things are set.

thedadams: Now that you've made me think about this a little more, I think I can make it better. I will push an update.

thedadams: After looking at this again upon my return, I don't believe I can make this better.

While it is true that it is possible for `project.Status.DefaultRegion` to be empty for a very short period of time, once it is non-empty, any controllers that rely on it will run again. Additionally, as I stated previously, this will not happen in the SaaS because of syncing.

Re-requesting your review to ensure we are on the same page.

njhale: SGTM, thanks for considering it so carefully

njhale (approved):
lgtm

Issue #2381, opened 2023-12-16
Support over-provisioning with computeClasses


cjellick:
This feature will allow  us to configure a compute class such that we can "overproivsion" pods to a node, based on memory. Meaning, schedule more pods to it than we currently can.

### Background

ComputeClasses are acorn's abstraction around scheduling pods. They primarily handle a few things:
- schedule rules such as tolerations, affinity, runtimeClass, and priorityClass
- resource requests and limits. Right now this is limited to memory and CPU.
The intention is that administrators configure these nitty gritty details behind the scenes and end users just need to say they want the "memory-optimized" or "gpu" compute class.

For the context of this feature, we are only concerned with the resource request/limit aspect of computeClasses, not the scheduling aspect.

When you create an acorn, you do two things related to resource consumption:
1. You say how much memory each container should get (or you get the default)
2. You say which computeClass it should use (or you get the default)

Our controllers translate that into resource requests and limits for memory and cpu. The user never directly specifies CPU. The computeClass has a CPUScaler field which basically says "for X memory give Y millicpu." We do this because we think users more easily understand the implications of memory than cpu scheduling and because in clouds like AWS, the ratio is always predictable, like with M5s you always get 1 cpu for every 4 GB of memory, so by controlling the ratio, we are scheduling more efficiently.

### Actual feature

Right now if you say your acorn gets 1GB of memory, that's exactly what gets set for the limit and request for memory. For cpu, we take 1G and math it (I forget how exactly) against the cpuScaler to calculate your cpu request. We do not set a cpu limit.

To handle over provisioning, we want to add a new memoryScaler field (name subject to change) that will change what is set for memory and cpu requests. It should be a decimal representing the percentage of the original value. For example, if my computeClass has `memoryScaler: .25` and I set `memory` on my acorn to `1G`, then the actual memory limit on the pod should be `.25G` (in whatever the right unit should be). Additionally, we should use the *scaled* value to calculate cpu request. So, with the previous example, if my cpuScaler was `.25`, then the cpu request for my pod would be `1000millicpu * .25 * .25` or 62.5 millicpu. Memory limit should remain exactly what the user requested (1G in this case).

When this lands, we should be able to handle "Old" computeClasses that didnt have the field by assuming they have a memoryScaler of 1.

[...]

PR #2394, opened 2024-01-02
ComputeClass request downscaling for overprovisioning (#2381)

keyallis:
#2381 

Add support for scaling down memory requests for acorns in relation to memory limits. We handle ComputeClasses that do not have this field set (ie legacy ComputeClasses) as having a 1:1 mapping. This means all existing tests should continue to work and behave as expected, so all new tests are only used to cover the new field itself.

### Checklist


thedadams (approved):
One nit. Otherwise, looks good to me.

tylerslaton (approved):
This looks good, I'm trying to imagine a situation where we want to scale up *above* the defined memory such that `1.25` would be a valid scalar. The thing I'm imagining is where an admin wants to allocate more memory than the user defined. Is that something we'd want?

If not this is good to go in my mind.

g-linville:
I'm just now starting to look at the code, but one thing I am noticing is that `scalar` needs to be replaced with `scaler` everywhere it appears. We're calling it a "scaler" because it determines the scale of the memory request in comparison to the limit. Mathematically speaking it is also a "scalar" since it is a single numerical value and not a vector, but that is not what we should name it lol.

g-linville (changes requested):
Lgtm other than the naming change that I mentioned in my previous comment.

g-linville (approved):

PR #2396, opened 2024-01-04
ComputeClass overprovisioning request floor ignores minimum (#2381)

keyallis:
#2381 

Remove the min bounding on compute class memory request determination. This was causing most pods to only be able to get scaled down ~50% which seems to not be enough for the dev use case.

### Checklist

### DIFF
--- a/pkg/computeclasses/computeclasses.go
+++ b/pkg/computeclasses/computeclasses.go
@@ -13,7 +13,6 @@ import (
 	internaladminv1 "github.com/acorn-io/runtime/pkg/apis/internal.admin.acorn.io/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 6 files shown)


cjellick (approved):

StrongMonkey (approved):

tylerslaton (approved):

tylerslaton:
Looks like you just need to update the autogold tests.

pkg/computeclasses/computeclasses.go
```diff
@@ -121,11 +120,7 @@ func Validate(cc apiv1.ComputeClass, memory resource.Quantity, memDefault *int64
 	return nil
 }
 
-func CalculateCPU(cc internaladminv1.ProjectComputeClassInstance, memDefault *int64, memory resource.Quantity) (resource.Quantity, error) {
-	if err := ValidateProjectComputeClass(cc, memory, memDefault); err != nil {
-		return resource.Quantity{}, err
-	}
```
keyallis, line 127: Removed this check since it's being called with the memory value associated to the request and not the limit. The request value can be outside of the range of the min-max defined in the compute class, and can also be outside of the defined set of `values` on the compute class if configured

pkg/computeclasses/computeclasses.go
```diff
@@ -134,18 +129,6 @@ func CalculateCPU(cc internaladminv1.ProjectComputeClassInstance, memDefault *in
...
-		ObjectMeta: metav1.ObjectMeta{
-			Name: cc.Name,
-		},
-		Memory:           cc.Memory,
-		Description:      cc.Description,
-		Default:          cc.Default,
-		SupportedRegions: cc.SupportedRegions,
-	}, memory, memDefault)
-}
-
```
keyallis, line 148: This function was only called in the above line that was removed, so I'm taking it out

pkg/controller/scheduling/scheduling.go
```diff
@@ -197,17 +197,6 @@ func ResourceRequirements(req router.Request, app *v1.AppInstance, containerName
...
-		// Never allocate less than the defined minimum of the compute class
-		if computeClass.Memory.Min != "" && computeClass.Memory.Min != "0" {
-			minValue, err := resource.ParseQuantity(computeClass.Memory.Min)
-			if err != nil {
-				return nil, err
-			}
-			if minValue.Cmp(memoryRequest) == 1 {
-				memoryRequest = minValue
-			}
-		}
```
keyallis, line 210: Actual code change to remove the floor for the requested amount

Issue #2411, opened 2024-01-12
Default values set for credential data fields are not usable.


sangee2004:
acorn version - v0.10.0-rc2-9-g43dbcbf4+43dbcbf4

**Steps to reproduce the problem:**
1. Deploy app with credentials that have default values  set for data fields 
```
secrets: config: {
	type: "credential.acorn.io/mytestcred"
	params: {
		promptOrder: ["username", "password"]
	}
	data: {
		username:      "abc"
		password:      "ppp"
	}
}
containers: {
	mytest: {
		image: "nginx"
	}
}
```

2. When user chooses to create a new credential , the default values are not presented when user is prompted to enter values for the fields (it is empty)

```
? Choose an existing credential or enter a new one Enter a new credential
? password 
? username 
```

3. Choosing to enter no values for these fields , does not result in credentials being created with the default values.

**Expected Behavior:**
1. Default values should be presented when user is prompted to enter values for the fields .
2. When user does not do any changes to the existing value, credential should get created with default values set for the data.

g-linville:
Just merged the fix.

g-linville:
Merged the fix. There is now no default displayed if the value is set to `""`.

sangee2004:
Tested with acorn version -  `v0.10.0-rc2-35-g3cf6c107+3cf6c107`

Default label is not presented for fields that do not have default values when prompted to enter credentials.

PR #2420, opened 2024-01-17
fix: credential secrets: use defaults as defined in the Acornfile (#2411)

g-linville:
for #2411

Now when the user is prompted to enter values for a credential secret, the default will be displayed to them, and if they enter nothing into the prompt, then the default will be used.

### Checklist

### DIFF
--- a/pkg/login/login.go
+++ b/pkg/login/login.go
@@ -106,10 +106,13 @@ func createSecret(ctx context.Context, c client.Client, app *apiv1.App, secretNa
 	data := map[string][]byte{}
 	promptOrder, _ := app.Status.AppSpec.Secrets[secretName].Params.GetData()["promptOrder"].([]string)
 	for _, key := range promptOrder {
-		if _, ok := app.Status.AppSpec.Secrets[secretName].Data[key]; ok {
-			value, err := prompt.Password(key)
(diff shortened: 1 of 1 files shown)


pkg/login/login.go
```diff
@@ -106,11 +106,14 @@ func createSecret(ctx context.Context, c client.Client, app *apiv1.App, secretNa
...
 	promptOrder, _ := app.Status.AppSpec.Secrets[secretName].Params.GetData()["promptOrder"].([]string)
 	for _, key := range promptOrder {
-		if _, ok := app.Status.AppSpec.Secrets[secretName].Data[key]; ok {
-			value, err := prompt.Password(key)
+		if def, ok := app.Status.AppSpec.Secrets[secretName].Data[key]; ok {
+			value, err := prompt.Password(fmt.Sprintf("%s (default: %s)", key, def))
 			if err != nil {
 				return err
 			}
+			if len(value) == 0 {
```
thedadams, line 114: Is it worth it to be a `strings.TrimSpace` on `value` here (and below)?

thedadams (approved):

pkg/login/login.go
g-linville: Imo, no. If the user enters spaces, I think those should be used as the value instead of changing it to the default.

iwilltry42 (approved):

Issue #2410, opened 2024-01-12
When user creates a new credentials as part of deploying app/ picks an existing one , he is prompted again to pick a credential.


sangee2004:
acorn version - `v0.10.0-rc2-9-g43dbcbf4+43dbcbf4`

**Steps to reproduce the problem:**

**Scenario 1**:

1. Deploy app with credentials
```
secrets: config: {
	type: "credential.acorn.io/mytestcred"
	params: {
		promptOrder: ["username", "password"]
	}
	data: {
		username:      "abc"
		password:      "ppp"
	}
}
containers: {
	mytest: {
		image: "nginx"
	}
}
```
2. When prompted for credentials , enter new credentials. 
3. After this , user is presented withe list of existing credentials (which also includes the one that was just created) and asked to choose from the list.

```
mytestnew
STATUS: ENDPOINTS[] HEALTHY[] UPTODATE[] 
STATUS: ENDPOINTS[] HEALTHY[0] UPTODATE[0] waiting for quota allocation; (container: mytest): pending; (secret: config): pending
? password **
? username **
STATUS: ENDPOINTS[] HEALTHY[0] UPTODATE[0] "acorn login mytestnew" required
STATUS: ENDPOINTS[] HEALTHY[0/1] UPTODATE[0] "acorn login mytestnew" required
STATUS: ENDPOINTS[] HEALTHY[0/1] UPTODATE[1] "acorn login mytestnew" required
? Choose an existing credential or enter a new one Existing: config-n778n (Keys: [password, username], Created: [2024-01-12 10:55:50 -0800 PST])
STATUS: ENDPOINTS[] HEALTHY[1] UPTODATE[1] "acorn login mytestnew" required

┌───────────────────────────────────────────────┐
| STATUS: ENDPOINTS[] HEALTHY[1] UPTODATE[1] OK |
└───────────────────────────────────────────────┘
```

**Expected Behavior**:
If user is creating a new cred , we should be able to use the created one for the app instead of having to letting him select again.

**Scenario 2:**

1. Deploy app with creds and create new cred.
2. Deploy another app with same credential type as step1.
3. When prompted for credentials , pick the existing one that gets listed.
4. Notice that user will be prompted again to pick a cred.

```
mytestnew5
STATUS: ENDPOINTS[] HEALTHY[] UPTODATE[] 
[...]

PR #2422, opened 2024-01-17
fix: credential secrets: avoid race condition when prompting user (#2410)

g-linville:
for #2410

There is a race condition that occurs only when deploying an acorn to the SaaS that defines a credential secret. The user will be prompted to enter credentials multiple times because the app keeps changing upstream. If we wait for a little bit (until all containers are marked as Defined), the race condition doesn't occur, and the user is only prompted once.

I also tested this with Acornfiles that only define this secret and a nested acorn, as well as an Acornfile that defines just the secret and nothing else, and it still works just fine.

### Checklist

### DIFF
--- a/pkg/log/default_log.go
+++ b/pkg/log/default_log.go
@@ -47,11 +47,26 @@ func (d *DefaultLoggerImpl) AppStatus(ready bool, msg string, app *apiv1.App) {
 		d.lock.Lock()
 		defer d.lock.Unlock()
 		if app.Status.AppStatus.LoginRequired && app.Status.ObservedGeneration > d.lastLogin {
-			err := login.Secrets(d.ctx, d.client, app)
-			if err != nil {
-				go d.Errorf(err.Error())
-			} else {
-				d.lastLogin = app.Generation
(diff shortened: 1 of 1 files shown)


thedadams (approved):

StrongMonkey (approved):

pkg/log/default_log.go
```diff
@@ -47,11 +47,26 @@ func (d *DefaultLoggerImpl) AppStatus(ready bool, msg string, app *apiv1.App) {
...
+				}
+			}
+
+			if allDefined {
+				err := login.Secrets(d.ctx, d.client, app)
+				if err != nil {
+					go d.Errorf(err.Error())
+				} else {
+					d.lastLogin = app.Generation
+				}
```
njhale, line 69: nano-nit:
```suggestion
				if err := login.Secrets(d.ctx, d.client, app); err != nil {
					go d.Errorf(err.Error())
				} else {
					d.lastLogin = app.Generation
				}
```

njhale (approved):
late to the party, but lgtm anyway

PR #2454, opened 2024-01-22
enhancement: add ComputeClasses and VolumeClasses field to BaseResources

tylerslaton:
This commit enhances the BaseResources struct by adding ComputeClasses and VolumeClasses fields. These new fields allow for a more detailed specification of compute and storage resources, categorized by class instead of just by resource type.

This enhancement is crucial for accurately tracking the usage of memory, CPU, and storage for each specific compute or volume class. Consequently, the QuotaRequest logic has been updated to account for these new fields.

These new fields are maps and they introduce a unique approach to handle unlimited resources. They include special keys, `AllComputeClasses` and `AllVolumeClasses`. If a value is assigned to these keys in ComputeClassResources or VolumeClassResources, all compute or volume classes will be evaluated against the value.

### Checklist


g-linville (changes requested):

pkg/apis/internal.admin.acorn.io/v1/computeclassresources.go
```diff
@@ -0,0 +1,144 @@
...
+
+	return nil
+}
+
+// ToString will return a string representation of the ComputeClassResources within the struct.
+func (current ComputeClassResources) ToString() string {
+	var resourceStrings []string
+
+	for computeClass, resources := range current {
+		resourceStrings = append(resourceStrings, fmt.Sprintf("Computeclass \"%s\": { %s }", computeClass, resources.ToString()))
```
StrongMonkey, line 124: Nit: you can use `%q` to add double quote

pkg/apis/internal.admin.acorn.io/v1/computeclassresources.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	return strings.Join(resourceStrings, ", ")
+}
+
+// Equals will check if the current ComputeClassResources struct is equal to another. This is useful
+// to avoid needing to do a deep equal on the entire struct.
```
StrongMonkey, line 131: Just curious why wouldn't you do a deepEqual?

StrongMonkey (approved):
LGTM.

All my comments are questions for my own understanding and some nit code style

pkg/apis/internal.admin.acorn.io/v1/volumeclassresources.go
```diff
@@ -0,0 +1,121 @@
...
+		current[volumeClass] = c
+	}
+}
+
+// Remove will remove the VolumeClassResources of another VolumeClassResources struct from the current one. Calling remove
+// will be a no-op for any resource values that are set to unlimited.
+func (current VolumeClassResources) Remove(incoming VolumeClassResources, all bool) {
+	if !all {
+		return
+	}
```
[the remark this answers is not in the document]
tylerslaton: Yeah, I originally wrote this to be closer to how `Remove` works for other `Resource` structs but upon further inspection, I think you're right.

pkg/apis/internal.admin.acorn.io/v1/computeclassresources.go
tylerslaton: The `DeepEqual` call didn't work well with Quantity resources when I tried in the past. I just got around figuring out why with this.

pkg/apis/internal.admin.acorn.io/v1/computeclassresources.go
tylerslaton: Oh yeah, good point! I forgot that exists.

pkg/apis/internal.admin.acorn.io/v1/computeclassresources.go
```diff
@@ -0,0 +1,144 @@
+package v1
+
+import (
+	"fmt"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+const AllComputeClasses = "*"
```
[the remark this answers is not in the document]
g-linville: We talked about this and I learned a bit more about what it is used for. The implementation makes sense, but Tyler is going to add more comments to the code to explain it.

pkg/apis/internal.admin.acorn.io/v1/baseresources.go
```diff
@@ -16,9 +14,10 @@ type BaseResources struct {
 	Volumes    int `json:"volumes"`
 	Images     int `json:"images"`
 
-	VolumeStorage resource.Quantity `json:"volumeStorage"`
```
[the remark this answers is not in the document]
tylerslaton: Talked offline and this won't be an issue because of the QuotaSyncer in Manager.

pkg/apis/internal.admin.acorn.io/v1/computeclassresources.go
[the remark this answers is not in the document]
tylerslaton: Added a comment to this section explaining it through!

g-linville (approved):

pkg/apis/internal.admin.acorn.io/v1/volumeclassresources.go
```diff
@@ -0,0 +1,117 @@
+package v1
+
+import (
+	"fmt"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+const AllVolumeClasses = "*"
```
keyallis, line 10: This could use a similar comment as the compute class one

pkg/controller/quota/testdata/implicit-pv-bind/input.yaml
```diff
@@ -39,3 +39,4 @@ status:
       volume-name:
         accessModes:
         - readWriteOnce
+
```
keyallis, line 42: nit: whitespace

keyallis (approved):

pkg/apis/internal.admin.acorn.io/v1/baseresources.go
```diff
@@ -75,43 +77,52 @@ func (current *BaseResources) Fits(incoming BaseResources) error {
...
+
+	if err := current.VolumeClasses.Fits(incoming.VolumeClasses); err != nil {
+		errs = append(errs, err)
 	}
 
 	// Build an aggregated error message for the exceeded resources
-	if len(exceededResources) > 0 {
-		return fmt.Errorf("%w: %s", ErrExceededResources, strings.Join(exceededResources, ", "))
+	if len(errs) != 0 {
+		return errors.Join(errs...)
```
thedadams, line 94: `errors.Join` does the right thing if `errs` is empty. You can unconditionally return its result.

pkg/apis/internal.admin.acorn.io/v1/baseresources.go
```diff
@@ -75,43 +77,52 @@ func (current *BaseResources) Fits(incoming BaseResources) error {
...
+	for _, resource := range []struct {
+		name     string
+		asString string
+	}{
+		{"ComputeClasses", current.ComputeClasses.ToString()},
+		{"VolumeClasses", current.VolumeClasses.ToString()},
+	} {
+		if result != "" && resource.asString != "" {
+			result += fmt.Sprintf(", %s: %s", resource.name, resource.asString)
+		}
```
thedadams, line 122: So, if `result` is empty, then nothing will be appended here. If this is correct, can you add a comment explaining why this should be the case?

pkg/apis/internal.admin.acorn.io/v1/computeclassresources.go
```diff
@@ -0,0 +1,155 @@
...
+func (current ComputeClassResources) Remove(incoming ComputeClassResources) {
+	for computeClass, resources := range incoming {
+		if _, ok := current[computeClass]; !ok || current == nil {
+			continue
+		}
+
+		c := current[computeClass]
+		c.Memory = SubQuantity(c.Memory, resources.Memory)
+		c.CPU = SubQuantity(c.CPU, resources.CPU)
+		current[computeClass] = c
```
thedadams, line 74: Based on the logic below, this line can be removed.
```suggestion
```

pkg/apis/internal.admin.acorn.io/v1/computeclassresources.go
```diff
@@ -0,0 +1,155 @@
...
+
+type ComputeClassResources map[string]ComputeResources
+
+// Add will add the ComputeClassResources of another ComputeClassResources struct into the current one.
+func (current ComputeClassResources) Add(incoming ComputeClassResources) {
+	for computeClass, resources := range incoming {
+		if _, ok := current[computeClass]; !ok {
+			current[computeClass] = ComputeResources{}
+		}
+
```
thedadams, line 55: This can be removed without changing the functionality.
```suggestion
```

pkg/apis/internal.admin.acorn.io/v1/volumeclassresources.go
```diff
@@ -0,0 +1,122 @@
...
+
+type VolumeClassResources map[string]VolumeResources
+
+// Add will add the VolumeClassResources of another VolumeClassResources struct into the current one.
+func (current VolumeClassResources) Add(incoming VolumeClassResources) {
+	for volumeClass, resources := range incoming {
+		if _, ok := current[volumeClass]; !ok {
+			current[volumeClass] = VolumeResources{}
+		}
+
```
thedadams, line 38: Same comment here about removing.
```suggestion
```

pkg/apis/internal.admin.acorn.io/v1/volumeclassresources.go
```diff
@@ -0,0 +1,122 @@
...
+		c.VolumeStorage = AddQuantity(c.VolumeStorage, resources.VolumeStorage)
+		current[volumeClass] = c
+	}
+}
+
+// Remove will remove the VolumeClassResources of another VolumeClassResources struct from the current one. Calling remove
+// will be a no-op for any resource values that are set to unlimited.
+func (current VolumeClassResources) Remove(incoming VolumeClassResources) {
+	for volumeClass, resources := range incoming {
+		if _, ok := current[volumeClass]; !ok || current == nil {
```
thedadams, line 49:
```suggestion
		if _, ok := current[volumeClass]; !ok {
```

thedadams (changes requested):

pkg/apis/internal.admin.acorn.io/v1/baseresources.go
tylerslaton: This logic was off, I'm updating it. Sorry for the confusion and thanks for pointing it out.

pkg/apis/internal.admin.acorn.io/v1/volumeclassresources.go
```diff
@@ -0,0 +1,122 @@
...
+	// Check if any of the quantity resources are exceeded
+	for volumeClass, resources := range incoming {
+		// If a specific volume class is defined on current then we check if it will
+		// fit the incoming resources. If is not defined, then we check if the current
+		// resources has AllVolumeClasses defined and if so, we check if the incoming
+		// resources will fit those. If neither are defined, then we deny the request
+		// by appending the volume class to the exceeded resources and continuing.
+		if _, ok := current[volumeClass]; !ok {
+			if _, ok := current[AllVolumeClasses]; ok {
+				volumeClass = AllVolumeClasses
```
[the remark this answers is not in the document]
tylerslaton: That could be pretty useful, I can see it. I'll code something up to see how this feels in execution.

pkg/controller/quota/quota.go
```diff
@@ -139,10 +139,17 @@ func addCompute(containers map[string]v1.Container, appInstance *v1.AppInstance,
...
 		for i := 0; i < replicas(container.Scale); i++ {
-			quotaRequest.Spec.Resources.CPU.Add(requirements.Requests["cpu"])
-			quotaRequest.Spec.Resources.Memory.Add(requirements.Requests["memory"])
+			quotaRequest.Spec.Resources.Add(adminv1.QuotaRequestResources{
+				BaseResources: adminv1.BaseResources{ComputeClasses: adminv1.ComputeClassResources{
+					computeClass: {
+						Memory: requirements.Requests["memory"],
+						CPU:    requirements.Requests["cpu"],
+					},
+				}}})
```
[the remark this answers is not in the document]
tylerslaton: Didn't even know that method existed, thanks Donnie.

pkg/apis/internal.admin.acorn.io/v1/baseresources.go
```diff
@@ -75,43 +77,51 @@ func (current *BaseResources) Fits(incoming BaseResources) error {
...
+	}{
+		{"ComputeClasses", current.ComputeClasses.ToString()},
+		{"VolumeClasses", current.VolumeClasses.ToString()},
+	} {
+		if result != "" && resource.asString != "" {
+			result += ", "
+		}
+		if resource.asString != "" {
+			result += fmt.Sprintf("%s: %s", resource.name, resource.asString)
+		}
```
njhale, line 121: nit: I think you could achieve the same thing with `strings.Join`, right?

pkg/apis/internal.admin.acorn.io/v1/baseresources.go
```diff
@@ -75,43 +77,51 @@ func (current *BaseResources) Fits(incoming BaseResources) error {
...
+		asString string
+	}{
+		{"ComputeClasses", current.ComputeClasses.ToString()},
+		{"VolumeClasses", current.VolumeClasses.ToString()},
+	} {
+		if result != "" && resource.asString != "" {
+			result += ", "
+		}
+		if resource.asString != "" {
+			result += fmt.Sprintf("%s: %s", resource.name, resource.asString)
```
njhale, line 120: nit: any reason for this to not be defined as part of the `ToString` method of the `Volume|ComputClasses` type?

pkg/apis/internal.admin.acorn.io/v1/computeclassresources.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+	}
+
+	return strings.Join(resourceStrings, ", ")
+}
+
+type ComputeClassResources map[string]ComputeResources
+
+// Add will add the ComputeClassResources of another ComputeClassResources struct into the current one.
+func (current ComputeClassResources) Add(incoming ComputeClassResources) {
```
njhale, line 50: nit: if we define these methods on `*ComputeClassResources`, then the structure of the underlying map isn't copied every time they are called.

njhale (approved):
left some nits, but they shouldn't hold this PR up.

pkg/apis/internal.admin.acorn.io/v1/baseresources.go
tylerslaton: When you call `.ToString()` from `Volume/ComputeClassesResources` I didn't want it to output `ComputeClasses: <ToString>`. I instead wanted to handle that explicitly in `BaseResources`. For example, `ToString()` on `BaseResources` doesn't output `BaseResources: <ToString>` when you call it.

pkg/apis/internal.admin.acorn.io/v1/baseresources.go
tylerslaton: `strings.Join()` will still add the separator if the string is empty which is not what I want. Is there another method that would work better here?

pkg/apis/internal.admin.acorn.io/v1/computeclassresources.go
njhale: spoke with tyler offline, but couldn't find supporting docs. what i did find was contrary to my statement; that copying a map is a effectively copying a single pointer instead of the map structure

thedadams (approved):

PR #2459, opened 2024-01-24
fix: handle multiple compute classes in a project

keyallis:
acorn-io/manager#1819

Handle the case of multiple compute classes being configured as default for a project spanning multiple regions.

### Checklist

### DIFF
--- a/integration/run/run_test.go
+++ b/integration/run/run_test.go
@@ -931,6 +931,310 @@ func TestDeployParam(t *testing.T) {
 	assert.Equal(t, "5", appInstance.Status.AppSpec.Containers["foo"].Environment[0].Value)
 }
 
+func TestMultipleDefaultComputeClass(t *testing.T) {
+	helper.StartController(t)
+	cfg := helper.StartAPI(t)
+	project := helper.TempProjectWithRegions(t, helper.MustReturn(kclient.Default), []string{apiv1.LocalRegion, "local2"})
+	kclient := helper.MustReturn(kclient.Default)
+	c, err := client.New(cfg, "", project.Name)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	ctx := helper.GetCTX(t)
+
+	checks := []struct {
+		name              string
+		noComputeClass    bool
+		testDataDirectory string
+		region            string
+		computeClasses    []adminv1.ProjectComputeClassInstance
+		expected          map[string]v1.Scheduling
+		waitFor           func(obj *v1.AppInstance) bool
+		fail              bool
+	}{
+		{
+			name:              "local-default",
+			testDataDirectory: "./testdata/simple",
+			region:            apiv1.LocalRegion,
+			computeClasses: []adminv1.ProjectComputeClassInstance{
+				{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "acorn-test-custom",
+						Namespace: c.GetNamespace(),
+					},
+					Default: true,
+					Memory: adminv1.ComputeClassMemory{
+						Default: "512Mi",
+						Max:     "1Gi",
+						Min:     "512Mi",
+					},
+					SupportedRegions: []string{apiv1.LocalRegion},
+				},
+				{ObjectMeta: metav1.ObjectMeta{
+					Name:      "acorn-test-custom-2",
+					Namespace: c.GetNamespace(),
+				},
+					Default: true,
+					Memory: adminv1.ComputeClassMemory{
+						Default: "513Mi",
+						Max:     "1Gi",
+						Min:     "512Mi",
+					},
+					SupportedRegions: []string{"local2"},
+				},
+			},
+			expected: map[string]v1.Scheduling{"simple": {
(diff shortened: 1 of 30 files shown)


pkg/server/registry/apigroups/acorn/apps/validator.go
```diff
@@ -584,13 +584,20 @@ func (s *Validator) checkScheduling(ctx context.Context, params *apiv1.App, proj
 		return append(validationErrors, field.Invalid(field.NewPath("spec", "image"), params.Spec.Image, fmt.Sprintf("error listing compute classes: %v", err)))
 	}
 
+	filteredComputeClasses := new(apiv1.ComputeClassList)
+	for _, cc := range computeClasses.Items {
+		if slices.Contains(cc.SupportedRegions, params.Spec.Region) || slices.Contains(cc.SupportedRegions, defaultRegion) {
```
thedadams, line 589: If the `params.Spec.Region != ""`, then we don't want to check the `defaultRegion` here.

However, we also need to collect all the compute classes that may be specified by a user or acornfile. In order of precidence:
1. `params.Spec.ComputeClasses`
2. `app.Status.AppSpec.{Containers,Jobs}[].ComputeClass` as well as `app.Status.AppSpec.{Containers,Jobs}[].Sidecars[].ComputeClass`
3. The default for the region.

By including all of these here, the checks below will catch if an app is trying to use a compute class that is not for the region.

In order to check 2, you would need the `imageDetails.AppSpec` from where this method is called.

thedadams (changes requested):
The two failing tests here are testing that an app cannot use a compute class that is not for its region. The reason they are failing is that the app is not specifying a region, so the "wrong" compute class is no longer being included.

The tests should be updated such that the app specifically tries to use the compute class that is not for its region. That should fail (which would mean the test would pass).

pkg/server/registry/apigroups/acorn/apps/validator.go
tylerslaton: ➕

integration/run/run_test.go
```diff
@@ -932,6 +932,304 @@ func TestDeployParam(t *testing.T) {
...
+			projectObjects = append(projectObjects, cc.DeepCopy())
+		}
+		// Perform the same test cases on both Project and Cluster ComputeClasses
+		for kind, computeClasses := range map[string][]crClient.Object{"projectcomputeclass": projectObjects, "clustercomputeclass": clusterObjects} {
+			testcase := fmt.Sprintf("%v-%v", kind, tt.name)
+			t.Run(testcase, func(t *testing.T) {
+				if !tt.noComputeClass {
+					for _, computeClass := range computeClasses {
+						if err := kclient.Create(ctx, computeClass); err != nil {
+							t.Fatal(err)
```
njhale, line 1165: nit: you can use `require.NoError(t, err)` to end the testcase early.

njhale (approved):
left a nit, but otherwise lgtm

pkg/controller/scheduling/scheduling.go
```diff
@@ -59,7 +59,7 @@ func addScheduling(req router.Request, appInstance *v1.AppInstance, workloads ma
 			tolerations []corev1.Toleration
 		)
 
-		computeClass, err := computeclasses.GetClassForWorkload(req.Ctx, req.Client, appInstance.Spec.ComputeClasses, container, name, appInstance.Namespace)
+		computeClass, err := computeclasses.GetClassForWorkload(req.Ctx, req.Client, appInstance.Spec.ComputeClasses, container, name, appInstance.Namespace, appInstance.Spec.Region)
```
thedadams, line 62:
```suggestion
		computeClass, err := computeclasses.GetClassForWorkload(req.Ctx, req.Client, appInstance.Spec.ComputeClasses, container, name, appInstance.Namespace, appInstance.GetRegion())
```

thedadams (changes requested):

integration/run/run_test.go
keyallis: The test structure was lifted from the function below this one so I'd prefer not to change much about it other than the looping. I thought about combining the two blocks together as well but figured it would be easier to review if it was just it's own function

tylerslaton (approved):

integration/run/run_test.go
```diff
@@ -1192,7 +1497,10 @@ func TestUsingComputeClasses(t *testing.T) {
 				}
 
 				// Assign a name for the test case so no collisions occur
-				app, err := c.AppRun(ctx, image.ID, &client.AppRunOptions{Name: testcase})
+				app, err := c.AppRun(ctx, image.ID, &client.AppRunOptions{
+					Name:   testcase,
+					Region: apiv1.LocalRegion,
```
thedadams, line 1502: Why is this changing? We should be able to continue to use the default region here instead of specifying one, correct?

keyallis: Yes, that was left over from previous versions of this PR, I've reverted it.

PR #2462, opened 2024-01-26
fix: add CPU field to resolvedOfferings and change QuotaRequests to use resolvedOfferings

tylerslaton:
This PR adjusts two main things.
1. Removes the `cpuScaler` field from ResolvedOfferings in favor of a new `cpu` field.
2. Updates QuotaRequests to report against the ResolvedOfferings fields for container compute.

These two things are necessary for ComputeClasses that define a requestScaler on memory as we want to quota against the requested amount, not the incoming scaled down amount.

### Checklist
- [X]  The title of this PR would make a good line in Acorn's Release Note's Changelog

### DIFF
--- a/pkg/controller/appdefinition/testdata/computeclass/all-set/existing.yaml
+++ b/pkg/controller/appdefinition/testdata/computeclass/all-set/existing.yaml
@@ -6,9 +6,9 @@ metadata:
 description: Simple description for a simple ComputeClass
 cpuScaler: 0.25
 memory:
-  min: 1Mi
-  max: 2Mi
-  default: 1Mi
+  min: 50Mi
+  max: 200Mi
+  default: 100Mi
 affinity:
   nodeAffinity:
     requiredDuringSchedulingIgnoredDuringExecution:
--- a/pkg/controller/quota/quota.go
+++ b/pkg/controller/quota/quota.go
@@ -8,7 +8,7 @@ import (
 	adminv1 "github.com/acorn-io/runtime/pkg/apis/internal.admin.acorn.io/v1"
 	"github.com/acorn-io/runtime/pkg/controller/appdefinition"
 	"github.com/acorn-io/runtime/pkg/labels"
-	corev1 "k8s.io/api/core/v1"
+	"github.com/acorn-io/z"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/acorn-io/runtime/pkg/condition"
@@ -124,7 +124,7 @@ func EnsureQuotaRequest(req router.Request, resp router.Response) error {
 // addContainers adds the number of containers and accounts for the scale of each container.
 func addContainers(containers map[string]v1.Container, quotaRequest *adminv1.QuotaRequestInstance) {
 	for _, container := range containers {
-		quotaRequest.Spec.Resources.Containers += replicas(container.Scale)
+		quotaRequest.Spec.Resources.Containers += int(replicas(container.Scale))
 	}
 }
 
@@ -132,18 +132,21 @@ func addContainers(containers map[string]v1.Container, quotaRequest *adminv1.Quo
 func addCompute(containers map[string]v1.Container, appInstance *v1.AppInstance, quotaRequest *adminv1.QuotaRequestInstance) {
 	// For each workload, add their memory/cpu requests to the quota request
 	for name, container := range containers {
-		var requirements corev1.ResourceRequirements
-		if specific, ok := appInstance.Status.Scheduling[name]; ok {
-			requirements = specific.Requirements
-		} else if all, ok := appInstance.Status.Scheduling[""]; ok {
-			requirements = all.Requirements
+		var cpu, memory resource.Quantity
+		if specific, ok := appInstance.Status.ResolvedOfferings.Containers[name]; ok {
+			memory = *resource.NewQuantity(z.Dereference(specific.Memory), resource.BinarySI)
+			cpu = *resource.NewMilliQuantity(z.Dereference(specific.CPU), resource.DecimalSI)
+		} else if all, ok := appInstance.Status.ResolvedOfferings.Containers[""]; ok {
+			cpu = *resource.NewMilliQuantity(z.Dereference(all.CPU), resource.DecimalSI)
+			memory = *resource.NewQuantity(z.Dereference(all.Memory), resource.BinarySI)
 		}
 
-		// Add the memory/cpu requests to the quota request for each container at the scale specified
-		for i := 0; i < replicas(container.Scale); i++ {
-			quotaRequest.Spec.Resources.CPU.Add(requirements.Requests["cpu"])
-			quotaRequest.Spec.Resources.Memory.Add(requirements.Requests["memory"])
-		}
+		// Multiply the memory/cpu requests by the scale of the container
+		cpu.Mul(replicas(container.Scale))
+		memory.Mul(replicas(container.Scale))
+
+		quotaRequest.Spec.Resources.CPU.Add(cpu)
+		quotaRequest.Spec.Resources.Memory.Add(memory)
 
 		// Recurse over any sidecars. Since sidecars can't have sidecars, this is safe.
 		addCompute(container.Sidecars, appInstance, quotaRequest)
@@ -256,9 +259,9 @@ func isEnforced(req router.Request, namespace string) (bool, error) {
 
 // replicas returns the number of replicas based on an int32 pointer. If the
 // pointer is nil, it is assumed to be 1.
-func replicas(s *int32) int {
+func replicas(s *int32) int64 {
 	if s != nil {
-		return int(*s)
+		return int64(*s)
 	}
 	return 1
 }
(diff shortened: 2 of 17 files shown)


tylerslaton:
This will have overlap with https://github.com/acorn-io/runtime/pull/2454. I will merge that one first but want to open this up for review.

pkg/controller/resolvedofferings/computeclass.go
```diff
@@ -94,8 +102,7 @@ func resolveComputeClass(req router.Request, appInstance *v1.AppInstance, config
 
 func resolveComputeClassForContainer(req router.Request, appInstance *v1.AppInstance, configDefault *int64, defaultCC *adminv1.ProjectComputeClassInstance, defaultCCName, containerName string, container v1.Container) (v1.ContainerResolvedOffering, error) {
 	var (
-		cpuScaler *float64
-		ccName    string
+		ccName string
```
g-linville, line 105: nit: this can be reduced to just `var ccName string`.

pkg/controller/resolvedofferings/computeclass.go
```diff
@@ -134,11 +143,13 @@ func resolveComputeClassForContainer(req router.Request, appInstance *v1.AppInst
 		}
 		def := parsedMemory.Def.Value()
 		memory = &def
+		cpuQuantity := computeclasses.CalculateCPU(*cc, *resource.NewQuantity(*memory, resource.BinarySI))
+		cpu = z.Pointer(cpuQuantity.MilliValue())
```
g-linville, line 147: This does not seem correct to me, as it only calculates the cpu if the user did not override the memory in the acorn image or with CLI flags and is just leaving it to the compute class' default memory. I think we want to calculate the cpu in every circumstance, right? And I don't think there is a way for the user to override the CPUScaler since it is hidden from them. So we should just have a separate if statement after this whole `if else if` series that does `if cc != nil` and then calculates the CPU based on the scaler from the CC.

g-linville (changes requested):

pkg/controller/resolvedofferings/computeclass.go
tylerslaton: Yep, you're right. I've gone ahead and fixed this, funny enough this was a last minute change and used to be what you suggested so I should've gone with my original gut feeling here lol.

pkg/controller/appdefinition/testdata/computeclass/all-set/existing.yaml
```diff
@@ -6,15 +6,18 @@ metadata:
...
+  requestScaler: 0.5
+# affinity:
+#   nodeAffinity:
+#     requiredDuringSchedulingIgnoredDuringExecution:
+#       nodeSelectorTerms:
+#       - matchExpressions:
+#         - key: foo
+#           operator: In
+#           values:
+#           - bar
```
g-linville, line 21: This should probably be uncommented

g-linville (approved):

pkg/controller/appdefinition/testdata/computeclass/all-set/existing.yaml
tylerslaton: Just left over testing stuff 👍🏻

pkg/controller/resolvedofferings/testdata/computeclass/compute-class-default/existing.yaml
```diff
@@ -17,9 +17,9 @@ description: Simple description for a simple ComputeClass
 cpuScaler: 0.25
 default: true
 memory:
-  min: 1Mi # 1Mi
-  max: 2Mi # 2Mi
-  default: 1Mi # 1Mi
+  min: 100Mi # 1Mi
+  max: 200Mi # 2Mi
+  default: 100Mi # 1Mi
```
keyallis, line 22: Update the comments here

pkg/controller/quota/quota.go
```diff
@@ -132,17 +132,19 @@ func addContainers(containers map[string]v1.Container, quotaRequest *adminv1.Quo
...
+			cpu = *resource.NewMilliQuantity(z.Dereference(all.CPU), resource.DecimalSI)
+			memory = *resource.NewQuantity(z.Dereference(all.Memory), resource.BinarySI)
 		}
 
 		// Add the memory/cpu requests to the quota request for each container at the scale specified
 		for i := 0; i < replicas(container.Scale); i++ {
-			quotaRequest.Spec.Resources.CPU.Add(requirements.Requests["cpu"])
-			quotaRequest.Spec.Resources.Memory.Add(requirements.Requests["memory"])
+			quotaRequest.Spec.Resources.CPU.Add(cpu)
+			quotaRequest.Spec.Resources.Memory.Add(memory)
```
thedadams, line 147: nit: cal we use the `Mul` method here instead of this loop?

thedadams (commented):
The resolved offerings get recalculated only when the app's generation changes. I believe that means that, after this change goes in, there will be apps that don't have `CPUScaler` (because it is being removed) nor `CPU` (because it's not being set). Are we OK with this? Should we handle this case?

pkg/controller/quota/quota.go
tylerslaton: We can, this logic is being done in a different PR.

tylerslaton:
> The resolved offerings get recalculated only when the app's generation changes. I believe that means that, after this change goes in, there will be apps that don't have CPUScaler (because it is being removed) nor CPU (because it's not being set). Are we OK with this? Should we handle this case?

The resolvedOfferings CPUScaler is currently not utilized anywhere in our codebase, including any dependencies of runtime. Therefore, removing it will have no impact on existing applications. It's worth noting that while there will be instances where resolvedOfferings.CPU is not set for an application, this is acceptable because the scheduling section will ensure the retention of the previously defined request. Once that application updates it will go through the process of getting resolvedOfferings.CPU instead.

thedadams (approved):

pkg/controller/quota/quota.go
```diff
@@ -124,26 +124,29 @@ func EnsureQuotaRequest(req router.Request, resp router.Response) error {
...
+			memory = *resource.NewQuantity(z.Dereference(all.Memory), resource.BinarySI)
 		}
 
-		// Add the memory/cpu requests to the quota request for each container at the scale specified
-		for i := 0; i < replicas(container.Scale); i++ {
-			quotaRequest.Spec.Resources.CPU.Add(requirements.Requests["cpu"])
-			quotaRequest.Spec.Resources.Memory.Add(requirements.Requests["memory"])
-		}
+		// Multiply the memory/cpu requests by the scale of the container
+		cpu.Mul(replicas(container.Scale))
```
tylerslaton, line 145: This was previously in the ComputeClass quota PR but I just brought it here instead.

g-linville:
Forgive me if my memory is failing, but even though I previously approved this PR, I find this choice confusing. I think it makes the most sense to just continue using the `scheduling` field for quota stuff instead of `resolvedOfferings`. I also don't know why we want CPU (or CPUScaler for that matter) on resolvedOfferings, because I think that is irrelevant to the user...

tylerslaton:
> Forgive me if my memory is failing, but even though I previously approved this PR, I find this choice confusing. I think it makes the most sense to just continue using the `scheduling` field for quota stuff instead of `resolvedOfferings`.

This all revolves around a recent field addition that we made to defining memory in ComputeClasses, `requestScaler` for memory. Let's look at an example of this in action to help illustrate the need here. Say that we define a ComputeClass where it has a memory `requestScaler` or `0.5`. If a user requests `1Gi` of memory then what they will actually have scheduled here is `512Mi` of memory. Since quota reads from the scheduling section this means we will quota for `512Mi` of memory. However, we really want to quota for `1Gi` of memory here. Our options were:
1. Read from the `limits` section of scheduling (already vetoed)
2. Read from the `ResolvedOfferings` section of status

> I also don't know why we want CPU (or CPUScaler for that matter) on resolvedOfferings, because I think that is irrelevant to the user...

I added this mainly since it is nicer to have `ResolvedOfferings` field tell me what the CPU should be as opposed to using the `cpuScaler` to determine what the CPU should be. In either situation I need to have an additional field tell me what (or how to calculate what) the CPU should be.

g-linville:
Thanks for the explanation. That makes sense now. Idk why I got so confused on Saturday lol.

g-linville (approved):

pkg/controller/resolvedofferings/computeclass.go
```diff
@@ -45,19 +48,24 @@ func resolveComputeClasses(req router.Request, cfg *apiv1.Config, appInstance *v
...
+		var cpuQuantity resource.Quantity
+		if cc != nil {
+			cpuQuantity = computeclasses.CalculateCPU(*cc, *resource.NewQuantity(*specificMemory, resource.BinarySI))
+		}
 		appInstance.Status.ResolvedOfferings.Containers[""] = v1.ContainerResolvedOffering{
-			Memory:    appInstance.Spec.Memory[""],
-			CPUScaler: appInstance.Status.ResolvedOfferings.Containers[""].CPUScaler,
-			Class:     appInstance.Status.ResolvedOfferings.Containers[""].Class,
+			Memory: appInstance.Spec.Memory[""],
+			CPU:    z.Pointer(cpuQuantity.MilliValue()),
```
njhale, line 67: just curious: is there a behavioral difference between `nil` and a pointer to a zero value?

njhale (approved):
One non-blocking question.

pkg/controller/resolvedofferings/computeclass.go
tylerslaton: Talked offline. This reflects a weirder issue in quota. I'm going to make an issue for that, merge this, and come back to fix what this points to.

PR #2470, opened 2024-02-01
Enable setting a default ComputeClass on projects

njhale:
Enable selecting the default `ComputeClass` for `Apps` deployed to a project via a new project spec field. 

The `DefaultComputeClass` field contains the name of a `[Project|Cluster]ComputeClassInstance` to select as the default. If no compute classes with that name exist, the existing `ComputeClass` defaulting rules are applied.

This PR also extends API validation for project create and update to ensure that a compute class with a matching name exists when the field is set.

Part of https://github.com/acorn-io/manager/issues/1943

### DIFF
--- a/pkg/apis/internal.admin.acorn.io/v1/default.go
+++ b/pkg/apis/internal.admin.acorn.io/v1/default.go
@@ -5,10 +5,12 @@ import (
 	"fmt"
 	"sort"
 
+	internalv1 "github.com/acorn-io/runtime/pkg/apis/internal.acorn.io/v1"
+	"github.com/acorn-io/z"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func getCurrentClusterComputeClassDefault(ctx context.Context, c client.Client) (*ClusterComputeClassInstance, error) {
+func getCurrentClusterComputeClassDefault(ctx context.Context, c client.Client, projectDefaultComputeClass string) (*ClusterComputeClassInstance, error) {
 	clusterComputeClasses := ClusterComputeClassInstanceList{}
 	if err := c.List(ctx, &clusterComputeClasses, &client.ListOptions{}); err != nil {
 		return nil, err
@@ -18,7 +20,7 @@ func getCurrentClusterComputeClassDefault(ctx context.Context, c client.Client) 
 		return clusterComputeClasses.Items[i].Name < clusterComputeClasses.Items[j].Name
 	})
 
-	var defaultCCC *ClusterComputeClassInstance
+	var defaultCCC, projectDefaultCCC *ClusterComputeClassInstance
 	for _, clusterComputeClass := range clusterComputeClasses.Items {
 		if clusterComputeClass.Default {
 			if defaultCCC != nil {
@@ -26,15 +28,26 @@ func getCurrentClusterComputeClassDefault(ctx context.Context, c client.Client) 
 					"cannot establish defaults because two default computeclasses exist: %v and %v",
 					defaultCCC.Name, clusterComputeClass.Name)
 			}
-			t := clusterComputeClass // Create a new variable that isn't being iterated on to get a pointer
-			defaultCCC = &t
+
+			// Create a new variable that isn't being iterated on to get a pointer
+			if projectDefaultComputeClass != "" {
+				defaultCCC = z.Pointer(clusterComputeClass)
+			}
 		}
+
+		if clusterComputeClass.Name == projectDefaultComputeClass {
+			projectDefaultCCC = z.Pointer(clusterComputeClass)
+		}
+	}
+
+	if projectDefaultCCC != nil {
+		return projectDefaultCCC, nil
 	}
 
 	return defaultCCC, nil
 }
 
-func getCurrentProjectComputeClassDefault(ctx context.Context, c client.Client, namespace string) (*ProjectComputeClassInstance, error) {
+func getCurrentProjectComputeClassDefault(ctx context.Context, c client.Client, projectDefaultComputeClass, namespace string) (*ProjectComputeClassInstance, error) {
 	projectComputeClasses := ProjectComputeClassInstanceList{}
 	if err := c.List(ctx, &projectComputeClasses, &client.ListOptions{Namespace: namespace}); err != nil {
 		return nil, err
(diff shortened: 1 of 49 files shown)


g-linville (changes requested):

pkg/server/registry/apigroups/acorn/projects/validator_test.go
```diff
@@ -93,10 +93,95 @@ func TestProjectCreateValidation(t *testing.T) {
...
+		},
+		{
+			name: "Create project with default computeclass that doesn't exist",
+			project: apiv1.Project{
+				Spec: v1.ProjectInstanceSpec{
+					DefaultComputeClass: "compute-class-dne",
+				},
+			},
+			wantError: true,
+		},
```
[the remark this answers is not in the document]
njhale: Thanks for catching that. Dropped the dupe.

pkg/project/handlers.go
```diff
@@ -14,7 +16,33 @@ import (
...
+	if project.Status.DefaultComputeClass != "" {
+		if _, err := computeclasses.GetAsProjectComputeClassInstance(req.Ctx, req.Client, project.Name, project.Status.DefaultComputeClass); err != nil {
+			if !apierrors.IsNotFound(err) {
+				return fmt.Errorf("failed to check existence of default compute class on project [%s] status: %w", project.Name, err)
+			}
+
+			// The compute class does not exist, clear the status field.
+			project.Status.DefaultComputeClass = ""
+		}
+		// TODO(njhale): Unset the status field if the project does not support the same regions as the compute class?
```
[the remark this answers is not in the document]
njhale: Meant to drop this comment. Updated.

g-linville (approved):

pkg/project/handlers.go
```diff
@@ -14,7 +16,32 @@ import (
...
 
-func SetProjectSupportedRegions(req router.Request, resp router.Response) error {
+// SetDefaultComputeClass sets the default compute class status field of a [v1.ProjectInstance] to the value of its spec
+// field if set.
+func SetDefaultComputeClass(req router.Request, resp router.Response) error {
+	project := req.Object.(*v1.ProjectInstance)
+	if cc := project.Spec.DefaultComputeClass; cc != "" && project.Status.DefaultComputeClass != cc {
+		// The spec has been changed, update the status field to match.
+		project.Status.DefaultComputeClass = cc
+	}
```
thedadams, line 26: Similar to default and supported region(s), we should always have a status value for default compute class. Otherwise, the status field is not necessary.

pkg/apis/internal.admin.acorn.io/v1/default.go
```diff
@@ -18,24 +21,33 @@ func getCurrentClusterComputeClassDefault(ctx context.Context, c client.Client)
...
 		if clusterComputeClass.Default {
 			if defaultCCC != nil {
 				return nil, fmt.Errorf(
 					"cannot establish defaults because two default computeclasses exist: %v and %v",
 					defaultCCC.Name, clusterComputeClass.Name)
 			}
-			t := clusterComputeClass // Create a new variable that isn't being iterated on to get a pointer
-			defaultCCC = &t
+
+			// Create a new variable that isn't being iterated on to get a pointer
```
thedadams, line 33: Is this comment still necessary?

Related: should we bump to Go 1.22?

pkg/apis/internal.admin.acorn.io/v1/default.go
```diff
@@ -18,24 +21,33 @@ func getCurrentClusterComputeClassDefault(ctx context.Context, c client.Client)
...
+		}
+
+		if projectSpecified != "" && clusterComputeClass.Name == projectSpecified {
+			projectSpecifiedCCC = z.Pointer(clusterComputeClass)
 		}
 	}
 
+	if projectSpecifiedCCC != nil {
+		return projectSpecifiedCCC, nil
+	}
```
thedadams, line 44: Spoke about this offline: if a user specifies a compute class that doesn't exist, then we should error here.

Additionally, we should check that the default region on the project is supported by the compute class.

Lastly, validation should be added for the user changing the default region on the project to ensure that the default compute class, if one is set, supports the region.

pkg/apis/internal.admin.acorn.io/v1/default.go
```diff
@@ -44,35 +56,70 @@ func getCurrentProjectComputeClassDefault(ctx context.Context, c client.Client,
...
+		}
+
+		if projectSpecified != "" && projectComputeClass.Name == projectSpecified {
+			projectSpecifiedPCC = z.Pointer(projectComputeClass)
 		}
 	}
 
+	if projectSpecifiedPCC != nil {
+		return projectSpecifiedPCC, nil
+	}
```
thedadams, line 79: Same comment here.

pkg/apis/internal.admin.acorn.io/v1/default.go
```diff
@@ -44,35 +56,70 @@ func getCurrentProjectComputeClassDefault(ctx context.Context, c client.Client,
...
 	if err != nil {
 		return "", err
-	} else if ccc != nil {
-		return ccc.Name, nil
 	}
+	if ccc != nil {
+		defaultComputeClasses = append(defaultComputeClasses, ccc.Name)
+	}
+
+	if sets.New(defaultComputeClasses...).Has(projectSpecified) {
```
thedadams, line 116: nit: slightly more performant.
```suggestion
	if slices.Contains(defaultComputeClasses, projectSpecified) {
```

pkg/server/registry/apigroups/acorn/projects/validator.go
```diff
@@ -22,15 +24,27 @@ type Validator struct {
...
-	return nil
+	if defaultComputeClass := project.Spec.DefaultComputeClass; defaultComputeClass != "" {
+		if _, err := computeclasses.GetAsProjectComputeClassInstance(ctx, v.Client, project.Name, defaultComputeClass); apierrors.IsNotFound(err) {
+			// The compute class does not exist, return an invalid error
+			result = append(result, field.Invalid(field.NewPath("spec", "defaultComputeClass"), defaultComputeClass, "default compute class does not exist"))
+		} else if err != nil {
+			// Some other error occurred while trying to get the compute class, return an internal error.
+			result = append(result, field.InternalError(field.NewPath("spec", "defaultComputeClass"), err))
+		}
+	}
```
thedadams, line 45: I left this comment elsewhere, but this is where we should ensure that the selected compute class supports the default region of the project.

thedadams (changes requested):

PR #2475, opened 2024-02-06
chore: add revive lint and fix related issues

thedadams:


thedadams:
I think the CI is failing because of the build cache. Doesn't make much sense to me.

pkg/local/server_generic.go
```diff
@@ -3,11 +3,9 @@
 package local
 
 import (
-	"context"
-	_ "embed"
 	"fmt"
 )
 
-func ServerRun(ctx context.Context) error {
+func ServerRun() error {
```
njhale, line 9: @thedadams the signature for this function in `pkg/local/server_linux.go` needs to be changed as well. I think that explains why this is failing in CI, but not locally.

njhale (changes requested):

pkg/local/server_generic.go
thedadams: Thanks for catching that.

njhale (approved):

tylerslaton (approved):

PR #2479, opened 2024-02-15
Fix: add logic to update volume reclaim policy to delete

StrongMonkey:
### Checklist

Previously, we have set PV/PVC's claimpolicy to `retain` to intentionally keep PV around when user has deleted their app. This is to avoid PV gets deleted when PVC is deleted. However this exposes a problem that when the volume/PV is deleted from acorn, the underlying storage resource is not cleaned up because of `retain` policy. So, when acorn initialize a delete call, we tweak the PV's reclaim policy to be `delete` before we actually delete the resource. This makes sure the driver will clean up the PV and underlying storage resource.

The downside of this fix, is that we will switch every PV's claimpolicy to `delete` regardless of what it has been actually configured in the cluster, if you delete PV through acorn api.

https://github.com/acorn-io/manager/issues/1764

### DIFF
--- a/pkg/server/registry/apigroups/acorn/volumes/policyswitcher.go
+++ b/pkg/server/registry/apigroups/acorn/volumes/policyswitcher.go
@@ -1,0 +1,49 @@
+package volumes
+
+import (
+	"context"
+
+	"github.com/acorn-io/mink/pkg/strategy"
+	"github.com/acorn-io/mink/pkg/strategy/remote"
+	"github.com/acorn-io/mink/pkg/types"
+	v1 "k8s.io/api/core/v1"
+)
+
+type PolicySwitcherStrategy struct {
+	strategy   strategy.CompleteStrategy
+	translator *Translator
+	remote     *remote.Remote
+}
+
+// NewPolicySwitcherStrategy returns a new policy switcher strategy that switch persistvolume policy from retain to delete when the volume is being deleted.
+// This makes sure the actual storage resource is deleted when the volume is deleted.
+func NewPolicySwitcherStrategy(s strategy.CompleteStrategy, t *Translator, r *remote.Remote) strategy.Deleter {
+	return &PolicySwitcherStrategy{
+		translator: t,
+		strategy:   s,
+		remote:     r,
+	}
+}
+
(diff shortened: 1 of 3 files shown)


pkg/server/registry/apigroups/acorn/volumes/translator.go
```diff
@@ -21,6 +21,36 @@ type Translator struct {
 	c kclient.Client
 }
 
+func (t *Translator) FromVolumeToPVName(ctx context.Context, namespace, name string) (string, error) {
+	i := strings.LastIndex(name, ".")
```
StrongMonkey, line 25: This is a copy of `FromPublicName` although it tries to list PV with acorn volume api.

pkg/server/registry/apigroups/acorn/volumes/policyswitcher.go
```diff
@@ -0,0 +1,49 @@
...
+		return nil, err
+	}
+	pv := pvObj.(*v1.PersistentVolume)
+	pv.Spec.PersistentVolumeReclaimPolicy = v1.PersistentVolumeReclaimDelete
+	_, err = p.r.Update(ctx, pv)
+	if err != nil {
+		return nil, err
+	}
+
+	return p.s.Delete(ctx, obj)
```
njhale, line 44: If this delete call fails we'd end up with a PV that doesn't have the retain policy but doesn't have a delete timestamp set.

I think we want to avoid getting into that state.

One suggestion to get around it: set the deletion timestamp before the update call to signal deletion.

njhale (changes requested):
Left a few thoughts and suggestions.

pkg/server/registry/apigroups/acorn/volumes/policyswitcher.go
StrongMonkey: I think the behavior should be ok. If delete fails, we will just leave the PV's reclaimPolicy to delete and wait for user to issue another delete(which is always the case in manager because cluster-agent will try to remove it when upstream volume is gone). No real impact on the actual PV itself expect the reclaimpolicy but we will always modify that next time you call delete.

g-linville (approved):

drpebcak (approved):

njhale (approved):

PR #2480, opened 2024-02-15
fix: resolved offerings: calculate if a container is missing

g-linville:
We saw cases (involving nested acorns) where resolved offerings would have containers missing from the list entirely. I think this was due to weird behavior where the `.Status.AppStatus.Containers` would not be populated on a nested acorn the first time it ran through the resolved offerings handler. To make sure that everything ultimately gets resolved, I added this new `needsCalculation` function that checks to make sure that every container defined in the app is also present in the resolved offerings.

### Checklist

### DIFF
--- a/pkg/controller/appdefinition/deploy.go
+++ b/pkg/controller/appdefinition/deploy.go
@@ -629,7 +629,7 @@ func containerAnnotation(container v1.Container) string {
 }
 
 func resolvedOfferingsAnnotation(appInstance *v1.AppInstance, container v1.Container) (string, error) {
-	if resolved, exists := appInstance.Status.ResolvedOfferings.Containers[container.Name]; exists {
+	if resolved, exists := appInstance.Status.ResolvedOfferings.Containers[strings.ToLower(container.Name)]; exists {
 		data, err := convert.EncodeToMap(resolved)
 		if err != nil {
 			return "", err
--- a/pkg/controller/appdefinition/deploy.go
+++ b/pkg/controller/appdefinition/deploy.go
@@ -31,7 +31,6 @@ import (
 	"github.com/acorn-io/schemer/data/convert"
 	"github.com/acorn-io/z"
 	"github.com/google/go-containerregistry/pkg/name"
-	"github.com/sirupsen/logrus"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierror "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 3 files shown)


pkg/controller/appdefinition/deploy.go
```diff
@@ -629,7 +629,7 @@ func containerAnnotation(container v1.Container) string {
 }
 
 func resolvedOfferingsAnnotation(appInstance *v1.AppInstance, container v1.Container) (string, error) {
-	if resolved, exists := appInstance.Status.ResolvedOfferings.Containers[container.Name]; exists {
+	if resolved, exists := appInstance.Status.ResolvedOfferings.Containers[strings.ToLower(container.Name)]; exists {
```
g-linville, line 632: While stepping through the debugger, I found a case here where `container.Name` was capitalized for some reason, which is why I added this `strings.ToLower`.

thedadams (approved):

pkg/controller/appdefinition/deploy.go
StrongMonkey: Interesting, I thought container name is not allowed to have capital name

StrongMonkey (approved):

g-linville:
Sorry @thedadams and @StrongMonkey. I just requested your re-review. While working on other stuff, I figured out the real problem with `container.Name`. Sometimes it just wasn't set at all for whatever reason. So I changed up the function that creates the resolved offerings annotation to just take the name of the container as a string, and I pass it all the way down to there.

pkg/controller/appdefinition/deploy.go
```diff
@@ -628,8 +629,9 @@ func containerAnnotation(container v1.Container) string {
 	return string(json)
 }
 
-func resolvedOfferingsAnnotation(appInstance *v1.AppInstance, container v1.Container) (string, error) {
-	if resolved, exists := appInstance.Status.ResolvedOfferings.Containers[container.Name]; exists {
+func resolvedOfferingsAnnotation(appInstance *v1.AppInstance, containerName string) (string, error) {
+	logrus.Infof("resolved offerings annotation - container name: %s", containerName)
```
StrongMonkey, line 633: Remove this

StrongMonkey (approved):

pkg/controller/appdefinition/deploy.go
g-linville: oh lol thanks

g-linville: done

thedadams (approved):

PR #2483, opened 2024-02-20
feat: add support for jobs that are both cron and event based

thedadams:
Before this change, a cronjob that also specifies events would produce a cronjob that is only running after one of the events occurs and is deleted after a non-listed event. Additionally, a cronjob is not marked as "ready" until it runs at least once.

After this change, a cronjob is marked as ready unless it is running or failed. Additionally, a job that specifies both events and schedule, a cronjob is created for the schedule and a "regular" job for each event. Once the event-based job succeeds, the job's status is read from the cronjob. That means, if the "regular" job fails, then the job's status will be "failed" until the "regular" job passed, regardless of what is happening with the cronjob.


thedadams:
I had this essentially done on Friday, so I was hoping to get it in if I could.

njhale (approved):

